PR #2, opened 2026-03-12
fix: OFX parsing logic

gmiranda13:
## What
OFX parser was excluding identical transactions but different FITIDs

Brief description of the changes.
## Why: Added FITIDs field to the import flow

Why are these changes needed?
## How: Its common to have idential transactions with same amount, date and description, it doesn´t mean it is duplicated if it has different IDs.

1. Import .ofx with identical transactions, but different ID

Here's what was changed:

  1. backend/app/schemas/transaction.py — Added external_id: Optional[str] = None to TransactionBase so it flows
  through the entire import pipeline (preview → confirm → persist). Removed the redundant field from
  TransactionRead since it's now inherited.
  2. backend/app/services/import_service.py —
    - parse_ofx: extracts txn.id (FITID) into external_id
    - import_transactions: when external_id is present, deduplicates by (account_id, external_id) instead of
  (date, amount, type, description). Also persists external_id to the Transaction model.
  3. backend/tests/test_import_service.py — Added TestParseOfx with two tests:
    - test_parse_ofx_extracts_fitid — verifies FITID populates external_id
    - test_parse_ofx_keeps_duplicate_looking_transactions — verifies two transactions with identical fields but
  different FITIDs are both preserved

## Checklist

- [X] Frontend lints clean (`npm run lint`)

### DIFF
--- a/docker-compose.prod.yml
+++ b/docker-compose.prod.yml
@@ -22,7 +22,7 @@ services:
   redis:
     image: redis:7-alpine
     healthcheck:
-      test: ["CMD", "redis-cli", "ping"]
+      test: [ "CMD", "redis-cli", "ping" ]
       interval: 5s
       timeout: 3s
       retries: 5
@@ -36,7 +36,7 @@ services:
     volumes:
       - pgdata:/var/lib/postgresql/data
     healthcheck:
-      test: ["CMD-SHELL", "pg_isready -U postgres"]
+      test: [ "CMD-SHELL", "pg_isready -U postgres" ]
       interval: 5s
       timeout: 3s
       retries: 5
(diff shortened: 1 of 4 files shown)


.gitignore
```diff
@@ -2,4 +2,7 @@
 .worktrees/
 .DS_Store
 .playwright-mcp/
+.claude
 docs/plans/
+RESUMO_PROJETO.md
```
tassionoronha: One small ask: could you remove RESUMO_PROJETO.md and Extratos/ from .gitignore? Those seem like local files, better for your global gitignore (~/.gitignore_global). 

The .claude one is fine to keep.

Issue #80, opened 2026-04-18
[Bug] Missing transactions on credit card


abdalanervoso-debug:
## Description

Some credit card transactions are not being imported/synced correctly. There are missing entries when compared to the original statement from the bank.

## Steps to Reproduce

Connect a credit card account via Open Finance / Pluggy
Wait for synchronization to complete
Open the credit card transactions list
Compare with the official bank statement

## Expected Behavior
All transactions from the credit card statement should be imported and displayed correctly in the app.

## Actual Behavior

Some transactions are missing. For example,  The April statement total in the bank is R$ 4,195.66, but the app shows R$ 4,086.68.
There are 2 missing transactions:

R$ 99.99
R$ 8.99

## Environment

- Deployment: [Docker / Manual]
- Browser: [e.g. Chrome 120]
- OS:  windows

## Screenshots

<img width="1319" height="190" alt="Image" src="https://github.com/user-attachments/assets/4e909635-b1bb-465d-82fe-50880f815268" />

<img width="717" height="224" alt="Image" src="https://github.com/user-attachments/assets/aad6ffa3-a662-4f17-afe7-c553d1aaf826" />

<img width="560" height="287" alt="Image" src="https://github.com/user-attachments/assets/0bf52696-4807-4f01-b53a-85ee28740608" />

tassionoronha:
Closing sincing was in a way a duplicate of #72 and resolved on #84. 
Statements missmatch to be resolved on #92.

PR #84, opened 2026-04-19
sync Pluggy transactions by createdAt instead of date (#72, #80)

### LINKED ISSUE #72: [Bug] Syncing an account doesn't retrieve every transaction
## Description

The credit card statements don't match the values inside the internet banking statements. Digging deeper to find why, I noticed some transactions missing from the statement. Looking for them inside the transactions page doesn't bring the transactions either.

Syncing the account again doesn't sync the missing transactions that can be found inside Pluggy's API.

## Steps to Reproduce

1. Go to the credit card page
2. Identify which statement does not match the value inside the internet banking
3. Filter the transactions from the statement period inside the transactions page
4. Make a GET request in the Transactions By Account ID endpoint (/transactions) inside Pluggy's API
5. Compare the quantity of transactions returned in step 3 with the quantity of results returned in step 4
6. Sync the account and repeat steps 3 to 5

## Expected Behavior

Every transaction for the synced period is inside Securo.

## Actual Behavior

Some transactions are missing within periods containing other transactions.

## Environment

- Deployment: Docker
- Browser: Safari 26.4
- OS: macOS Tahoe 26.4.1
- Securo version: v0.8.3

## Additional Context

When comparing the statements that weren't synced with the synced ones, I couldn't find any difference to explain why they were ignored.

tassionoronha:
## What

Switch Pluggy `/transactions` filter from `from` (txn date) to `createdAtFrom` (Pluggy ingestion time).

## Why

Closes #72 and #80. Pluggy backdates late-ingested rows (e.g. credit card bill payments dated to the bill close). The old `from` filter excluded them and the 14-day rewind never reached far enough, so they were silently dropped forever.

## How to Test

1. Sync a Pluggy credit card connection.
2. Compare DB transactions against Pluggy's `/transactions` API for that account.
3. No rows missing.

## Checklist

- [x] Backend tests pass (`pytest`) — 1004 passed, 6 skipped
- [ ] Frontend lints clean (`npm run lint`) — n/a, backend only
- [ ] Frontend builds (`npm run build`) — n/a, backend only
- [ ] Translations updated (if user-facing strings changed) — n/a

### DIFF
--- a/backend/app/providers/pluggy.py
+++ b/backend/app/providers/pluggy.py
@@ -210,7 +210,14 @@ class PluggyProvider(BankProvider):
                     "page": page,
                 }
                 if since:
-                    params["from"] = since.isoformat()
+                    # Filter by Pluggy ingestion time, NOT transaction date.
+                    # `from`/`to` filter on `date` (when the txn happened),
+                    # which silently drops transactions Pluggy ingests later
+                    # but backdates — e.g. credit card bill payments dated
+                    # to the bill close, or merchants that settle weeks late.
+                    # `createdAtFrom` filters on Pluggy's row creation time,
+                    # so any newly-ingested row is fetched regardless of date.
+                    params["createdAtFrom"] = since.isoformat()
 
                 resp = await client.get(
                     f"{PLUGGY_API_BASE}/transactions",
--- a/backend/app/services/connection_service.py
+++ b/backend/app/services/connection_service.py
@@ -417,12 +417,11 @@ async def sync_connection(
             if account and account.is_closed:
                 continue
 
-            # Fetch and sync transactions.
-            # Rewind the window by 14 days on incremental syncs so we catch
-            # late-arriving transactions: bank authorizations that post a few
-            # days after the purchase date, or direct debits that Pluggy only
-            # exposes once they've been processed by the issuer. Dedup on
-            # external_id below handles any overlap cheaply.
+            # Fetch and sync transactions. The 14-day rewind is on Pluggy's
+            # `createdAt` (when their row was inserted), so it covers two
+            # cases: (1) PENDING transactions that POSTED since last sync,
+            # (2) any rows Pluggy ingested late but backdated. Dedup on
+            # external_id below handles overlap cheaply.
             since = (
                 connection.last_sync_at.date() - timedelta(days=14)
                 if connection.last_sync_at


tassionoronha:
Hey @ingridmachado and @abdalanervoso-debug, 

Could you check if this fix resolves your cases?

abdalanervoso-debug:
image
Hi, I applied the fix locally and synced again. The totals changed a bit, so it definitely improved, but they still don’t fully match my credit card statement. There are still some transactions missing — I’m sure the Wellhub charge is one of them.

tassionoronha:
Thanks for confirming @abdalanervoso-debug !

Could you confirm that these missing transactions are appearing in the fetch of transactions we request from Pluggy? 
Need to understand where this data is being lost.

abdalanervoso-debug:
Bingo, I just confirmed that the transaction doesn’t show up in MeuPluggy either. It only appears in the Bradesco bank and credit card app. Do you have any idea what might be causing this?
image

tassionoronha:
Very intriguing case @abdalanervoso-debug 😐

But if Pluggy dont manage to catch this transaction via open finance, unfortunately we dont have a lot of control about it on our side :/ I dont have on my two CC missing transactions anymore. I have indeed 2 before the fix of this PR, but now they are present. 

But just to understand: you said the transaction doesnt show up in MeuPluggy, but this screenshot looks to show the transaction which is missing? Not sure I got it. 

To fully explore this case I would recommend start debugging the [Pluggy's transactions endpoint](https://docs.pluggy.ai/docs/transactions) and check if you can somehow catch the missing transaction there.

One last question: are these missing transactions recent(less than 45 days for example)? 
Some bank could maybe delay to expose some data in their open finance interfaces and what we see, but maybe in some days this will available to Pluggy. Just a thought, i dont know if the case for Bradesco since I use Nubank and Inter only.

abdalanervoso-debug:
There should be two Wellhub transactions each month — one for me and one for my wife — as shown on the credit card statement below. My wife’s transaction is missing every month. So these are not recent transactions, but from statements that are already closed. There are also other missing transactions, but this one caught my attention the most because it’s a recurring payment that is missing every month.

image

abdalanervoso-debug:
I would like to update a few points regarding the discrepancy in the invoice values.

Some missing transactions that are causing differences in the invoice total are actually related to the closing date logic in Securo. For example, the transaction of R$ 8.99 from 27/02 appeared in Securo’s March invoice, while on the credit card statement it only appears in April. The same issue also occurred in January’s invoice, with transactions from 27/11 and 28/11 being shifted across billing cycles.

However, this alone does not fully explain the discrepancy, as there are still completely missing transactions that I was not able to find even directly in MeuPluggy.

I also noticed that there may be a limit of 500 transactions per sync. I am wondering if this limitation could be related to the missing data.

Additionally, I observed that my bank provides a separate API for Bradesco Cartões (credit card only). I was previously using the Bradesco account API, which includes both bank account and credit card data together. However, switching to the dedicated credit card API resulted in the same behavior and did not resolve the issue.

ingridmachado:
Hello @tassionoronha!

Testing the sync again in the development environment brought the missing transactions in the production environment.

But I am having the same issue @abdalanervoso-debug is facing, with my Securo statements still not matching the internet banking statements. As far as I know, I think it has something to do with installments.

To keep focus on the problem, I agree you can consider the issue #72 closed. I will go through my last statements to make sure my theory is right. And if it is the case, I will open another issue.

Thank you for the fast fix!

tassionoronha:
Thanks for the deep dig @abdalanervoso-debug 🙏

A few things from my side:

1. **Wellhub (and other Pluggy gaps):** if it's not showing up in MeuPluggy, it's not reaching Pluggy from Bradesco, there's nothing we can pull on our side. Could be specific case for Bradesco, but still what is the pattern for these specific transactions? In any case, this is likely to be in Bradesco or Pluggy side and for now we could only try to highlight to their support maybe? 

2. **I investigate the dedup logic:** Our description match needs ≥ 0.9 (token overlap), and "Wellhub Gustavo Abdala" vs "Wellhub Carla Mello da Lu" only scores ~0.2. The cleanup also requires the sibling to be paired by transfer detection, which doesn't fire for two CC purchases. So even if Pluggy were returning both, we wouldn't drop one.

3. **Closing-date bucketing (R$8.99 of 27/02):** that's a separate issue, could you open it as its own bug with the example? Different code path from this PR. 

4. **500 limit:** that's `pageSize=500` *per page*, the loop paginates to the end, so it's not a hard cap. 

Going to merge this since it fixes the createdAt/from issue @ingridmachado confirmed. The remaining gaps look like upstream Pluggy/Bradesco data issues + the bucketing bug, which deserve their own threads IMO. 

Lot of unknown cases guys, but we're progressing :) 
Thanks for the engagement to investigate your own statements and for aiming the full match, this is the spirit!

tassionoronha:
Thanks for checking @ingridmachado. 
We can maybe open another issue to discuss about this statement discrepeancy once we get more information about what could be happening. The installments or even recurring txs could be the source of this, we need to understand. 

For instance, I have recurring txs working well and settlements are just fine, but don't have installments.

abdalanervoso-debug:
@tassionoronha 
On the Wellhub topic, I found a similar case in another invoice. On the same day, with an identical transaction value, the API ended up ignoring one of the entries and only showing a single occurrence. In the comparison view, the green entry is present both in Securo and in the bank, while the white entry appears only in the bank.

This is extremely concerning, as I have been reviewing all invoices from December until now, and these are the only cases where a non-installment transaction is missing — specifically this one and the Wellhub transaction. Both are purchases with the exact same value and on the same day, which makes the pattern even more suspicious.
image

I also found other missing transactions within the same invoice. However, these are installment payments from very old purchases (around 1 year ago), such as 12/12 installments.

It might be the same issue that @ingridmachado  mentioned, related to how installments are being handled or reconciled in the system.

image

Regarding the transactions that are present but appear in the wrong billing cycle due to an incorrect closing date in another invoice, I will open a separate topic to address this specifically.

Thanks a lot for your support and for looking into this.

ingridmachado:
Hello @abdalanervoso-debug!

I don't know if you have already looked for support from Pluggy, but the MeuPluggy team can help you via Discord. More info [here](https://github.com/pluggyai/meu-pluggy?tab=readme-ov-file).

PR #97, opened 2026-04-20
fix: display account nickname transactions and reports

leorcdias:
## What

Fixes account custom nickname (display_name) not being reflected in the transactions list and net worth report composition view — only the sidebar was updated.

## Why

The transactions list was reading `account.name` directly instead of going through `getAccountName()`. The report composition was also using `account.name` in the backend when building the composition labels.

## How to Test

1. Set a nickname for a bank-connected account (edit → Apelido)
2. Go to **Transactions** — the account column should show the nickname
3. Go to **Reports → Net Worth → Composition → Detailed** — the account label should show the nickname

## Checklist

- [x] Backend tests pass (`pytest`)
- [x] Frontend lints clean (`npm run lint`)
- [x] Frontend builds (`npm run build`)
- [x] Translations updated (if user-facing strings changed)

Refs #85

PR #103, opened 2026-04-20
fix: display name goals link transfer (#97)

leorcdias:
## What

Use `display_name` (with fallback to `name`) when showing account names in the goals list and in the link-transfer dialog anchor card.

## Why

Both places were reading `account.name` directly, ignoring any custom `display_name` set by the user. The backend fix reuses the existing `get_account_name()` helper from `account_service`. The frontend fix reuses `getAccountName()` from `account-utils`, consistent with the rest of the same file.

## How to Test

1. Set a `display_name` on any account
2. Create a goal linked to that account — confirm the goals list shows the `display_name`
3. Create a transaction on that account and open the link-transfer dialog — confirm the anchor card shows the `display_name`

## Checklist

Refs #97

### DIFF
--- a/backend/app/services/account_service.py
+++ b/backend/app/services/account_service.py
@@ -3,6 +3,10 @@ from datetime import date as _Date, datetime, timedelta, timezone
 from decimal import Decimal
 from typing import Optional
 
+
+def get_account_name(account: "Account") -> str:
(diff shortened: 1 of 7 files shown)


tassionoronha:
Good work @leorcdias !
Congrats :)

PR #112, opened 2026-04-21
Fix: translate provider description

leorcdias:
## What

Added i18n support for the Pluggy provider description shown in the connector selection dialog. The description is now stored in the translation files instead of being rendered directly from the API response.

## Why

The provider description `"Open finance provider for Brazilian banks"` was hardcoded in the backend and displayed as-is in the frontend, ignoring the active language. This change ensures the string is translated when the user switches between Portuguese and English.

## How to Test

1. Open the app and go to Accounts
2. Click "Connect Bank" to open the connector selection dialog
3. Verify the Pluggy description appears in the current language
4. Switch the language (Settings → language toggle)
5. Reopen the dialog and verify the description updated accordingly

## Checklist


tassionoronha (approved):
LGTM! Thanks for this @leorcdias !

PR #113, opened 2026-04-21
fix: auto-detect CSV delimiter on bank statement import

leorcdias:
## What

Auto-detect the CSV delimiter (`,`, `;`, `\t`, `|`) when importing bank statements, instead of always assuming comma.

## Why

CSVs using semicolons (or other delimiters) caused a misleading "Could not detect CSV columns" error because the entire header row was read as a single field name.

## How to Test

1. Create a CSV with semicolon delimiter, e.g.:
date;description;amount
15/01/2026;Grocery Store;-120.50
2. Import it via the bank statement import flow.
3. Verify the transactions are parsed correctly without errors.

## Checklist

### DIFF
--- a/backend/app/services/import_service.py
+++ b/backend/app/services/import_service.py
@@ -207,7 +207,12 @@ def parse_csv(
     - inflow_column/outflow_column: use split columns instead of single amount
     """
     text = content.decode('utf-8-sig')  # Handle BOM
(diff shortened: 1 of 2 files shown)


tassionoronha (approved):
Good call, approving here @leorcdias !

Issue #110, opened 2026-04-21
[Bug] Error when deleting account


ivancarlosti:
## Description

Error when I try to delete an specific account with related import history

## Steps to Reproduce

1. Click on delete button of a specific account while the account has an import log loaded, it's not properly deleted
3. See error
<img width="98" height="62" alt="Image" src="https://github.com/user-attachments/assets/1641a268-5df7-46dc-9752-9da689acf964" />

## Expected Behavior

Account deleted successfully or at least a friendly alert reporting that it has an import history that should be deleted first

## Actual Behavior

Error, no deletion, no friendly warning
Other accounts without related import record are deleted successfully
Workaround: To delete manually import entry prior account deletion

```
backend-1        | asyncpg.exceptions.ForeignKeyViolationError: update or delete on table "accounts" violates foreign key constraint "import_logs_account_id_fkey" on table "import_logs"
backend-1        | DETAIL:  Key (id)=(632f0a72-5789-4011-915e-c384dc8ef6a6) is still referenced from table "import_logs".
```

## Environment

- Deployment: Docker

## Screenshots

<img width="1916" height="946" alt="Image" src="https://github.com/user-attachments/assets/b3ffc07f-8407-4cab-a9e7-7d4ecd537953" />

## Additional Context

Docker log:

```
frontend-1       | 172.18.0.1 - - [21/Apr/2026:17:01:16 +0000] "GET /accounts HTTP/1.1" 200 1026 "-" "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/147.0.0.0 Safari/537.36" "2804:14c:bf28:2244:e99d:5253:18e9:5cf2"
frontend-1       | 172.18.0.1 - - [21/Apr/2026:17:01:16 +0000] "GET /assets/index-0-QcKSUq.js HTTP/1.1" 200 664860 "https://money.example.com/accounts" "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/147.0.0.0 Safari/537.36" "2804:14c:bf28:2244:e99d:5253:18e9:5cf2"
[...]

PR #114, opened 2026-04-21
fix: delete account with import history

leorcdias:
## What

(#110) Fixed a 500 Internal Server Error that occurred when deleting accounts with associated import history. Also fixed the import history list not updating immediately after account deletion.

## Why

The `import_logs` table has a foreign key referencing `accounts`, but no cascade delete was configured. When the backend tried to delete the account, PostgreSQL blocked it with an integrity constraint violation. Additionally, the frontend was not invalidating the `import-logs` query cache after account deletion, causing stale data to persist until page reload.

## How to Test

1. Import a file into any account
2. Go to Accounts and delete that account
3. Verify the deletion succeeds without errors
4. Navigate to Import — verify the import history no longer shows entries for the deleted account without needing to refresh

## Checklist


tassionoronha:
nice fix @leorcdias ! just two things before merging:

1. i think it would be cleaner to add a migration with ON DELETE CASCADE on import_logs.account_id, then pg handles it and we don't risk hitting the same 500 from another delete path later.
  
2. could you add a test for "delete account with import logs"? that's exactly the case that broke, would be sad to regress it silently.

PR #117, opened 2026-04-23
feat: read type and category columns from CSV import

leandronascimento:
parse_csv now reads an explicit 'type' column (credit/debit) so that
  all-positive-amount CSVs are imported correctly instead of being treated
  as credits. Falls back to amount-sign derivation when the column is absent
  or has an unrecognized value.

  parse_csv also reads a 'category' / 'categoria' column and stores the
  value as category_name on TransactionBase. import_transactions resolves
  the name to a category_id via a single per-import query against the
  user's categories, leaving the field null when no match is found.

  ## What

  - `parse_csv` now reads an explicit `type` column (`credit`/`debit`) when present
  - `parse_csv` now reads a `category` / `categoria` column and stores it as `category_name`
  - `import_transactions` resolves `category_name` to `category_id` using the user's categories
  - Added `category_name: Optional[str]` to `TransactionBase` schema
  - 13 new unit tests covering all new behavior

  ## Why

  When re-importing a CSV previously exported from Securo, all amounts are
  positive and the `type` column distinguishes credits from debits. Without
  reading it, every transaction was imported as `credit`. The `category`
  column was also silently ignored, requiring manual re-categorization after
  every import.

  ## How to Test

  1. Export a CSV from Securo with existing categorized transactions
  2. Re-import the CSV — transactions should keep their original type and category
  3. Import a CSV with a `category` column containing an unrecognized name — transaction imports successfully with
   no category (no error)
  4. Import a CSV without `type` or `category` columns — existing behavior unchanged (type derived from amount
  sign)

  ## Checklist

  - [ x] Backend tests pass (`pytest`)

### DIFF
--- a/backend/app/schemas/transaction.py
+++ b/backend/app/schemas/transaction.py
@@ -17,6 +17,7 @@ class TransactionBase(BaseModel):
     currency: Optional[str] = None
     fx_rate: Optional[Decimal] = None
     payee_raw: Optional[str] = None  # raw payee string from import (OFX/QIF)
+    category_name: Optional[str] = None  # resolved to category_id during import
 
 
 class TransactionCreate(TransactionBase):
--- a/backend/app/schemas/transaction.py
+++ b/backend/app/schemas/transaction.py
@@ -17,7 +17,6 @@ class TransactionBase(BaseModel):
     currency: Optional[str] = None
     fx_rate: Optional[Decimal] = None
     payee_raw: Optional[str] = None  # raw payee string from import (OFX/QIF)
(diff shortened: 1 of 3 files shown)


tassionoronha:
Hey @leandronascimento , welcome here :D 

The fix looks good overall, solid fix. 

One small thing: category_name on TransactionBase leaks into every TransactionRead response as "category_name": null. I'd move it to a TransactionImport(TransactionBase) subclass used only on the import path. 

If you're fine with my change I'm ready to merge it. 
Cheers!

PR #129, opened 2026-04-25
feat: add backup restore capability

KiwiDev808:
## What

- **Restore Capability**: The main feature added is the ability to securely **restore backups** into the application using an incremental **"Safe Merge / Upsert"** methodology (`session.merge()`). Existing data is preserved entirely during imports.
- **Export Refactoring**: To support the new restore feature smoothly and increase code readability, the export functionality was heavily refactored. Extracted the monolithic API logic into a cleanly isolated Object-Oriented architecture (`ExportService`) and cleanly localized ZIP map processing into an extensible `BackupArchiveHandler`.
- **Typing & i18n Improvements**: Added structured data types and implemented comprehensive frontend Portuguese and English locale translations (`pt-BR.json`, `en.json`) matching the restoration dialog prompts seamlessly.

## Why

Why are these changes needed?
Adding restore capabilities closes the loop on system portability, ensuring that users can safely return their backed-up archives without destroying live configuration records. The internal export code natively needed a cleanup to accommodate this cleanly — doing so vastly scales future extensibility, strips procedural block redundancies, and sets strong PEP 8 boundaries aligned with the core structural architecture. 

## How to Test

1. **Verify Extracted Locales**: View the front-end localized modal strings internally by verifying that the "Restore Backup" prompt functions smoothly inside English or Portuguese.
2. **Execute Restore Integration**: Export a ZIP, insert a manual transaction locally into the UI, then use the brand new "Restore" function to upload your saved ZIP. 
3. Verify that your newest manual transaction is still there intact, proving that the backup successfully _merged_ with live states instead of wiping out the DB entirely.

## Checklist

## Screenshots
https://github.com/user-attachments/assets/c844ace6-22c4-479c-84ed-ce9c3f3fef9f

### DIFF
--- a/backend/app/api/export.py
+++ b/backend/app/api/export.py
@@ -16,9 +16,10 @@ async def backup(
     session: AsyncSession = Depends(get_async_session),
     user: User = Depends(current_active_user),
 ):
-    from app.services.export_service import export_user_data
+    from app.services.export_service import ExportService
 
-    buf = await export_user_data(session, user.id)
+    service = ExportService(session, user.id)
+    buf = await service.export_data()
     today = date.today().isoformat()
     return StreamingResponse(
         iter([buf.getvalue()]),
@@ -33,7 +34,7 @@ async def restore(
     session: AsyncSession = Depends(get_async_session),
     user: User = Depends(current_active_user),
 ):
-    from app.services.export_service import BackupArchiveHandler, restore_user_data
+    from app.services.export_service import BackupArchiveHandler, ExportService
 
     try:
         content = await file.read()
@@ -42,7 +43,8 @@ async def restore(
         raise HTTPException(status_code=400, detail=f"Invalid backup file: {str(e)}")
 
     try:
-        await restore_user_data(session, user.id, data_map)
+        service = ExportService(session, user.id)
+        await service.restore_data(data_map)
         await session.commit()
     except Exception as e:
         await session.rollback()
--- a/backend/app/api/export.py
+++ b/backend/app/api/export.py
@@ -1,37 +1,14 @@
-import io
-import json
-import zipfile
-from datetime import date, datetime
-from decimal import Decimal
-from uuid import UUID
+from datetime import date
 
 from fastapi import APIRouter, Depends, File, HTTPException, UploadFile
 from fastapi.responses import StreamingResponse
-from sqlalchemy import select, delete
 from sqlalchemy.ext.asyncio import AsyncSession
 
 from app.core.auth import current_active_user
 from app.core.database import get_async_session
-from app.models.account import Account
-from app.models.asset import Asset
-from app.models.asset_group import AssetGroup
-from app.models.asset_value import AssetValue
-from app.models.budget import Budget
-from app.models.category import Category
-from app.models.category_group import CategoryGroup
-from app.models.goal import Goal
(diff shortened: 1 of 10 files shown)


tassionoronha:
Hey @KiwiDev808, welcome :) 
Happy to see people from Bahia here!

Could you please add some description in the PR once you're done to also help other contributors to understand what's going on and how to test your work?

This feature is a big and important one, thanks for putting energy on it 🥇 

Cheers!

KiwiDev808:
Hi @tassionoronha thanks for the welcome!

I updated the description with the changes made, if anything more is needed you can tell me.

Cheers

tassionoronha:
Hey @KiwiDev808, thanks for that. 
CI is failling on the lint, can you check that and also if tests pass as well? :) 
Thank you!

tassionoronha:
Really good job here @KiwiDev808. 
I think you picked a really challenge task as your first one tho, so I will point some things we should review:


**1. session.merge() overwrites existing rows, the PR description is then inaccurate**

The PR says "Existing data is preserved entirely during imports," but merge() loads the existing row by primary key and overwrites every column with the backup's values. So if a user restores an older backup, any account/category/transaction whose UUID still exists gets silently reverted to the old state. Only rows with new UUIDs (like a manual transaction added after the backup) survive untouched.

I would suggest to keep it simple to skip on PK collision for now.

**2. Foreign keys aren't validated for ownership**

_sanitize_row overrides user_id to the current user, but doesn't validate that account_id, category_id, etc. actually belong to that user. On a multi-user instance, a crafted backup could insert a transaction with user_id=attacker but account_id=another user's account, corrupting the victim's ledger view.

Each FK referencing a per-user resource should be validated against the current user's owned IDs.

**3. connection_id is silently nulled on restore**

BankConnection is excluded from the export (correctly, credentials is sensitive and, despite the comment, not actually encrypted). But on restore, accounts referencing unknown connections have connection_id blanked silently with no signal to the user.

Better to strip connection_id during export so the backup is self-consistent, rather than mutating it on import.


WDYT about those?

KiwiDev808:
Hello @tassionoronha sorry for the delay in the response, i applied the suggested changes, you can review the PR again?

tassionoronha:
Hey @KiwiDev808, points 1 and 3 are great, thanks for that  🙌

Tested point 2 with a real backup containing an account + transaction (transaction.account_id is NOT NULL), restored into an empty DB. It returns HTTP 500:

NOT NULL constraint failed: transactions.account_id parameters: (..., None, None, ...)

Reason: in restore_data, existing_keys is fetched once before any inserts. So when sanitizing transactions, the just-inserted account isn't in the snapshot yet → account_id gets nulled → INSERT crashes. Same chain for category_groups → categories (silent null instead of crash, since group_id is nullable).

This breaks the primary use case (full restore to a fresh DB). Could you fix it and add a test that restores a parent + child chain?

We're getting there :)

DannyFeliz:
I hope we get this eventually, I'm needing this, I have the backup of my instance but not a way to restore it 😢

tassionoronha:
@KiwiDev808 @DannyFeliz 

Given the level of changes of our app since the PR, I would prefer to restart this work in another PR, with a more flexible and safe approach  regarding how we handle the objects deps.

Issue #126, opened 2026-04-24
[Bug] Investments Not Consistently Imported from Bank Sync (Inconsistent Behavior Across Institutions)


otomay:
## Description

Investments are not consistently imported when connecting accounts via Bank Sync. The behavior appears to vary depending on the order in which institutions are connected.

In one scenario, after connecting Inter and Nubank, investments were only imported for Inter. In another attempt, connecting Nubank first and then Inter resulted in no investments being imported for either institution.

However, when checking the Pluggy dashboard, the investments are correctly present there, indicating the data is available but not being properly reflected in Securo.

## Steps to Reproduce

1. Connect a Bank Sync account (e.g., Nubank)
2. Connect another account (e.g., Inter)
3. Sync transactions and investments
4. Check the Investments section

**Alternative scenario:**

1. Connect Inter first
2. Then connect Nubank
3. Sync data
4. Check the Investments section

## Expected Behavior

* Investments should be consistently imported for all connected institutions
* The order of account connection should not affect whether investments are displayed
* If investments are available in the provider (Pluggy), they should be reflected in the application

## Actual Behavior

* Investments appear inconsistently or not at all
* In some cases, only one institution shows investments
* In other cases, none of the institutions show investments
* This happens despite investments being visible in the Pluggy dashboard

## Environment

* Deployment: Docker (0.9.2 backend/frontend)
* Browser: Brave
* OS: Windows 11

## Screenshots

Inter first:
<img width="274" height="126" alt="Image" src="https://github.com/user-attachments/assets/981889f3-689d-4f16-8e4f-82c2b07715b3" />

Nubank first:
<img width="236" height="175" alt="Image" src="https://github.com/user-attachments/assets/126339ce-2239-451e-a979-09675c547407" />

Pluggy dashboard for inter:
<img width="563" height="153" alt="Image" src="https://github.com/user-attachments/assets/7a3fb2cf-73a5-4396-9c55-5d39f5cca6bc" />

Pluggy dashboard for nubank:
<img width="527" height="141" alt="Image" src="https://github.com/user-attachments/assets/623cf089-a683-486e-bb58-c53eac305a07" />

gustavohrg:
For what I observed, this is less of a Pluggy/Bank Sync problem and more about removing and then including the same account multiple times. While working on #148, I encountered this exact same issue. After including an account for the first time, it works fine, but after removing it and including it again, the behavior produces mixed results. Accidentally, I think I solved this as well in the same deploy!

[...]

PR #148, opened 2026-05-02
Bug/account unlink keeping entries

### LINKED ISSUE #124: [Bug] Account Unlink keeps old entries
## Description

When unlinking an account from Bank Sync, the account disappears immediately from the interface without any confirmation prompt. However, the unlinking does not fully propagate across the system.

The account associations remain visible and active in other sections of the application, such as the Assets (Patrimony) page and the Merchant page, indicating an inconsistency between the UI state and the actual data.

## Steps to Reproduce

1. Go to the Bank Sync page
2. Locate a linked account
3. Click on the option to unlink/remove the account
4. Observe that the account disappears instantly
5. Navigate to the Assets (Patrimony) page
6. Navigate to the Merchant page

## Expected Behavior

* The system should prompt for confirmation before unlinking the account
* Once confirmed, the account should be fully unlinked across all parts of the system
* All related associations should be removed consistently from every page

## Actual Behavior

* The account is removed instantly from the Bank Sync UI without confirmation
* The unlink action does not fully apply system-wide
* The account and its associations remain visible in other sections (Assets and Merchant pages)

## Environment

* Deployment: Docker (0.9.2 backend/frontend)
* Browser: Brave
* OS: Windows 11

## Screenshots

Button to unlink:
Image

Patrimony page (I unlinked 4 bank syncs):
Image

Can't post merchant page, but basically yhe system is with no accounts, and all merchantes are there.

---

Edit: Regarding multiple “Meu Pluggy” entries continuing to appear on the Assets (Patrimony) page, I noticed—while opening another issue—that these connections are still active in the Pluggy dashboard. This may be a possible cause, or another related issue explaining why they continue to appear in the system.

Image

(I only have 2 bank accounts)

gustavohrg:
## What

Closes #124. Implements a mandatory confirmation dialog before unlinking bank connections, updates the Assets page to exclude archived holdings post-unlink, and filters out zero-transaction "ghost" payees from the Merchants page. Adds corresponding i18n keys (`accounts.confirmDisconnectTitle`, `accounts.confirmDisconnectDesc`) and 3 new backend tests.

## Why

Previously, unlinking a connection only updated the Accounts page. Archived investment assets remained visible on the Assets (Patrimony) page because the frontend explicitly requested archived records (`assets.list(true)`), and payees with no remaining transactions persisted on the Merchants page due to a missing transaction-count filter in `payee_service.py`. Additionally, the unlink mutation fired immediately on click without user confirmation, posing a risk of accidental, irreversible data loss. These fixes ensure consistent UI state propagation across all pages and add a necessary safety confirmation step.

## How to Test

1. **Confirmation Dialog:** Navigate to the Accounts page, click the unlink button on a connected institution, and verify a confirmation modal appears (displaying the institution name) before the `disconnectMutation` fires.

Screenshot from 2026-05-02 18-27-06

3. **Assets Page:** After confirming the unlink, navigate to the Assets/Patrimony page and verify that holdings from the removed connection are no longer visible (frontend now queries `assets.list(false)`).

Screenshot from 2026-05-02 18-52-22

5. **Merchants Page:** Navigate to the Merchants page and verify that payees with zero active transactions are automatically filtered out.
6. **Tests:** Run `pytest` to verify all 134 tests pass, including the 3 new tests:
   - `test_delete_connection_archives_linked_assets`
   - `test_get_assets_excludes_archived_by_default`
   - `test_get_payees_excludes_zero_transaction_payees`
7. **i18n & Build:** Verify frontend linting/building passes and confirm the new confirmation dialog strings render correctly in both `en.json` and `pt-BR.json`.

## Checklist

---
### Out-of-Scope Findings (for separate issues)
- **Celery Worker Loop Error:** `RuntimeError: Future attached to a different loop` caused by inherited `asyncpg` pool after fork. *Fix:* Dispose SQLAlchemy engine in `@worker_process_init` signal.
- **Orphaned Pluggy Connections:** External provider dashboard still shows connections after internal unlink. Requires investigation into `connection_service.delete_connection` and the Pluggy provider adapter.
- **`test_create_payee` Fixture Error:** Pre-existing `pytest-asyncio >= 0.24` incompatibility with session-scoped `event_loop`. *Fix:* Pin `< 0.24` or migrate to `asyncio_mode = "auto"`.


gustavohrg:
The failure in [CI Backend](https://github.com/securo-finance/securo/actions/runs/25263276406/job/74073914090?pr=148) is tracked as out-of-scope for this PR and is not caused by the changes in this branch.

gustavohrg:
Maybe this closes #126 too! Need to check it out.

tassionoronha:
Thanks for the work here @gustavohrg, most of the diff is great. Two blockers:

1. CI failure isn't out-of-scope, it's from this PR. CI fails on test_list_payees, ot test_create_payee. The new tx_count > 0 filter in payee_service.py hides every zero-tx payee, including freshly-created ones, breaks the payee picker in
transaction-dialog.tsx, rules.tsx, and transactions.tsx (user creates a payee to assign to a tx, but it doesn't show up until a tx already references it).

Fix at the root: in delete_connection, delete payees with no remaining transactions and no external mappings.

2. asset.is_archived = False in _upsert_asset_from_holding will un-archive assets the user archived on purpose. is_archived is also user-controlled (exposed on AssetUpdate), and there's no flag saying who archived it. So the next sync will quietly un-archive it. Two options: add an archived_by column, or only un-archive when connection_id changed (which is the delete-then-reconnect case you actually want to handle).

Apart of that, I think we're good to go :)

gustavohrg:
Oh, sorry for letting the first one slide! It was a misconception on my part. And thanks for clarifying the second one. I’ll work on both soon.

tassionoronha:
Thanks for the fixes, @gustavohrg . LGTM now! :)

Issue #141, opened 2026-04-30
[Bug] Transfers with category field blocked


ingridmachado:
## Description

When two transactions are linked as a transfer, the form blocks the field Category while showing the message "Bank-synced transaction — only category and notes can be edited".

## Steps to Reproduce

1. Go to the Transactions page
2. Click on a transaction with a transfer badge
3. Click on the Category field
4. The mouse click will be blocked

## Expected Behavior

The category should be allowed to change.

## Actual Behavior

There is a block in the category field and the value can't be changed.

## Environment

- Deployment: Docker
- Browser: Safari 26.4
- OS: macOS Tahoe 26.4.1

## Additional Context

With the addition of the option to treat a category as a transfer, there is no need to block the inclusion of a category to organize the transaction.

ingridmachado:
Implemented in the PR #150

PR #150, opened 2026-05-03
chore(transaction-dialog): remove disabled state for category selecton on transfer

gustavohrg:
## What
This PR closes issue #141 and enables users to change the category of a transaction marked as **Transfer**.

## Why
Currently, users cannot recategorize transfer transactions, which limits flexibility in organizing their financial data. This change allows users to assign or modify categories for transfer transactions, improving usability and alignment with their workflows.

image

## How to Test
1. Open a transaction marked as **Transfer**.
2. Attempt to change its category.
3. Verify that the category is updated successfully.

## Point of Discussion
We could solve this by simply removing the `disabled` attribute. However, I’d like to propose an enhanced solution: Since transfer transactions come in pairs, we could prompt the user to apply the category change to the paired transaction as well. I’d love to hear your thoughts on this approach!

### DIFF
--- a/backend/app/schemas/transaction.py
+++ b/backend/app/schemas/transaction.py
@@ -42,6 +42,7 @@ class TransactionUpdate(BaseModel):
     notes: Optional[str] = None
     amount_primary: Optional[Decimal] = None
     fx_rate_used: Optional[Decimal] = None
+    apply_to_transfer_pair: bool = False
     # CC bucketing override (issue #92). Empty string / explicit null clears
     # it back to auto. Only meaningful for credit-card accounts.
     effective_bill_date: Optional[_Date] = None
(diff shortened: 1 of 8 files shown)


tassionoronha:
Hey @gustavohrg , the solution makes sense for me. 
And I also agree with your suggestion to propose the change also to the paired tx as well since very likely this is something the users will want to do most of the time. Can you add this please?

gustavohrg:
> Hey @gustavohrg , the solution makes sense for me. And I also agree with your suggestion to propose the change also to the paired tx as well since very likely this is something the users will want to do most of the time. Can you add this please?

Absolutely! Here's the update files ;)

image

tassionoronha:
Looks amazing. 
Thanks for this one @gustavohrg!!

gustavohrg:
Thanks! Always glad to help here ;)

PR #159, opened 2026-05-06
fix: update select all functionality to exclude shared transactions

rodrigost23:
## What

This PR fixes the "select all" behavior when there are transactions shared from a group.

## Why

Currently, the select all checkbox selects even transactions that cannot be selected:

before

Also, notice that selecting the only two transactions does not make the top checkbox fully checked.

This is what it looks like after this PR:

after

## How to Test

1. Open a transactions page that contains shared transactions from groups
2. Toggle the "select all" checkbox

## Checklist

- [x] Backend tests pass (`pytest`) (N/A)
- [x] Translations updated (if user-facing strings changed) (N/A)


tassionoronha:
LGTM, well done @rodrigost23 !

PR #160, opened 2026-05-06
feat: unify category selection UI and introduce category grouping

rodrigost23:
## What

This PR improves the category selection UI across the app. The final result replaces native `<select>` dropdowns with a custom component that shows category colors, and organizes the options by category groups.

Screen Shot 2026-05-06 at 08 16 05
Screen Shot 2026-05-06 at 08 15 41

## Why

I initially just wanted to make the category selection in the bulk action bar prettier, with colors, similar to how it looks in the filter bar. 

To get there, the work is split into a few progressive commits. If you don't like the final grouping feature, we can just drop the last commit.

**1. Filter bar refactor (Commit 1)**
Replaces the custom `CheckRow` in the filter bar with the standard `DropdownMenuCheckboxItem`. I did this because I originally thought I'd reuse it for the bulk action bar. I ended up not doing that, but I kept the commit because it removes custom stuff in favor of standard components.

If you feel this is unnecessary, I can remove this commit from the PR and just add grouping (or not) using the `CheckRow` component.

**2. Unified Category Select with colors (Commits 2 & 3)**
Adds the shadcn `Select` component and creates a reusable `CategorySelect`. This replaces the ugly native selects in the bulk action bar, manual transaction dialog, rules, and recurring forms so they all share the same UI and category colors logic.

(I just ran `npx shadcn add select` to add that `Select` component) in commit.

|BEFORE|AFTER|
|----|----|
|Screen Shot 2026-05-06 at 08 21 23|Screen Shot 2026-05-06 at 08 20 17|
|transaction-0|transaction-1|

**3. Category grouping (Commit 4)**
Since category groups already exist in the DB, I updated the new `CategorySelect` and the filter bar to actually group the categories in the dropdowns. 

This commit changes a lot of code, since I needed to move from using categories lists to category groups lists.

|BEFORE|AFTER|
|----|----|
|dropdown-0|dropdown-2|
|Screen Shot 2026-05-06 at 08 20 17|Screen Shot 2026-05-06 at 08 15 41|
|rules-1|rules-2|
|recurring-1|recurring-2|

## How to Test

1. Open the filter bar and check the categories submenu (grouped, standard checkboxes).
2. Select transactions and use the bulk action category dropdown (grouped, with colors).
3. Open the manual transaction, recurring, or rules dialogs and check the category dropdowns (grouped, with colors).

## Checklist

- [x] Backend tests pass (`pytest`) (N/A)

***Disclaimer:** Gemini helped me a bit with this one*

### DIFF
--- a/frontend/src/components/category-select.tsx
+++ b/frontend/src/components/category-select.tsx
@@ -1,10 +1,10 @@
-import { Select, SelectContent, SelectItem, SelectTrigger, SelectValue } from '@/components/ui/select'
-import type { Category } from '@/types'
+import { Select, SelectContent, SelectGroup, SelectItem, SelectLabel, SelectTrigger, SelectValue } from '@/components/ui/select'
+import type { CategoryGroup } from '@/types'
 
 interface CategorySelectProps {
   value: string
   onChange: (value: string) => void
-  categories: Category[]
+  groups: CategoryGroup[]
   placeholder?: string
   disabled?: boolean
   className?: string
@@ -15,7 +15,7 @@ interface CategorySelectProps {
 export function CategorySelect({
   value,
   onChange,
-  categories,
+  groups,
   placeholder = 'Select category',
   disabled = false,
   className = "w-full border border-border rounded-md px-3 py-2 text-sm bg-background disabled:opacity-50 disabled:cursor-not-allowed focus:outline-none focus-visible:ring-ring/30 focus-visible:ring-[2px]",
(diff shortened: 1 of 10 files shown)


tassionoronha:
Hey @rodrigost23, amazing work here!
Such a good UX improvement regarding the categories, thanks for that.

I found in my tests one issues tho:

1. Categories without a group no longer show up in the dropdowns (filter, bulk bar, dialogs). Looks like the switch to GET /api/category-groups drops ungrouped categories. Could we surface them under a synthetic "Sem grupo" group for example?

rodrigost23:
Hi @tassionoronha! Just pushed an update. What I did was bring back the flat categories list query alongside the category groups on all pages, exactly like how the Categories page handles it.

An alternative would be changing the backend's /category-groups endpoint to automatically wrap orphaned categories into a synthetic "No group" bucket. But doing it that way would introduce a localization issue and I felt that the frontend handling this looks cleaner.

tassionoronha:
Again, amazing work here @rodrigost23! 
Thanks for that 🙌🏼

Issue #173, opened 2026-05-08
[Feature] Improve transaction import


paulo-paes:
## Problem

Sometimes when importing credit card ofx files, it will have the payment value from previous bill cycle, but users may already done the in-app transaction of that payment and want to delete that specific row

The current import review flow is too limited and Lacks visibility/control: 
- Only the first 50 imported transactions are displayed, Making it impossible to review large imports completely
- Users cannot exclude individual transactions before confirming the import
- Users cannot see which Category will be assigned to each transaction
- Users cannot adjust categories during the import process 

This becomes especially problematic for credit card OFX imports, where files often include the payment transaction for the previous billing cycle. In some cases, users have already manually recorded that payment in the app. Causing duplicate transactions after import. Since individual rows cannot currently be removed, users are forced to clean this up afterward.

## Proposed Solution

Improve the transaction import review experience by introducing a full import preview table with transaction-level controls.

Allow users to:

- View all imported transactions
- Add pagination or virtual scrolling for performance when handling large files
- Exclude specific transactions before confirming the import
- See the category that will be assigned to each transaction
- Change categories directly during the review step
- Search and filter transactions during review

paulo-paes:
Feature released in v0.11.1

PR #178, opened 2026-05-09
feat: improve transaction import (#173)

paulo-paes:
## What

Solves #173

Improve transaction import

image

- [x] View all imported transactions
- [x] Add pagination or virtual scrolling for performance when handling large files
- [x] Exclude specific transactions before confirming the import
- [x] See the category that will be assigned to each transaction
- [x] Change categories directly during the review step
- [x] Search and filter transactions during review

## Why

Why are these changes needed?
Improve general UX during imports

## How to Test

1. Access import page
2. import a csv or ofx
3. Choose which transactions should be included or excluded
4. Import
5. Verify if transactions were added

## Checklist


paulo-paes:
We might need to update the category component if #160 gets merged

tassionoronha:
Hey @paulo-paes, nice improvements! Tested it and found a couple of issues:

1. Rule actions get dropped when a category is set. In import_service.py, apply_rules_to_transaction is now skipped when category_id is resolved, but that function also handles set_payee and append_notes. The category_already_set flag inside apply_rule_actions already protects the category, so removing the if not category_id: guard should be safe and restores payee/tag rules.
  
2. Can't uncategorize a row that has a rule suggestion. In import.tsx, the import payload uses selected_category_id ?? suggested_category_id ?? undefined. Picking "Uncategorized" sets selected_category_id to null, which falls back to the suggestion. Need to distinguish "untouched" from "explicitly cleared".

3. Heads up on #160 (just merged): the row category picker and the category filter are using native <select>s. We've now got CategorySelect in main which handles grouping + colors — worth using here for consistency.

Can you check those please?

paulo-paes:
1. Removed if not category_id and added 3 tests to verify if payee and rules are applied correctly

2. I used undefined/null to distinguish between untouched and explicity cleared in frontend. Also added a flag force_uncategorized to TransactionImport, so if user set a transaction to uncategorized we skip set_category rules but keep applying others rule types. Added 4 tests to validate it.

3. I added the new category component to import preview list, but had to do a minor refactor on transactions-filter-bar so we are able reuse the same filter component inside import page. Created two components:
    - CategoryFilterDropdown a self-contained dropdown filter component
    - CategoryFilterContent a shared dropdown content component that can be used alongside other filters


@tassionoronha  Let me know if anything else needs to be adjusted

tassionoronha:
Thanks @paulo-paes, I'll come back to you soon :)

tassionoronha:
LGTM @paulo-paes . Good work here and thanks for taking in consideration my comments :)

paulo-paes:
Thanks for the review and feedback @tassionoronha, really appreciated. Glad we could get this merged 🚀

PR #187, opened 2026-05-12
Mitigate Supply chain attacks for frontend dependencies

paulo-paes:
## What

These properties aims to mitigate supply chain attacks, by allowing a cooldown time to devs/maintainers check their own releases and ignore script execution by 3rd party libs

## Why

Today we had a massive [npm supply chain attack](https://safedep.io/mass-npm-supply-chain-attack-tanstack-mistral/) infecting famous packages, and they seem to be happening more frequently

## Checklist


paulo-paes:
To these properties work we need to use npm version >= 11.10.0. So we need to change Docker Image or update npm in the scripts

Also I checked that we have a similar property for python with uv, but that requires a switch from pip to uv

uv property

`exclude-newer = "1 week"`

MarlonPassos-git:
> To these properties work we need to use npm version >= 11.10.0. So we need to change Docker Image or update npm in the scripts
> 
> Also I checked that we have a similar property for python with uv, but that requires a switch from pip to uv
> 
> uv property
> 
> `exclude-newer = "1 week"`

I think this is a useful hardening step, but I would scope the title/description a bit more narrowly.

Right now this PR only adds npm-level protections under frontend/.npmrc, so it mitigates a subset of supply-chain risks for the frontend dependency installation path. It does not cover the Python/backend dependency chain yet. I would suggest renaming the PR to something more specific. And we could handle the Python side in a separate PR.

> Today we had a massive [npm supply chain attack](https://safedep.io/mass-npm-supply-chain-attack-tanstack-mistral/) infecting famous packages, and they seem to be happening more frequently

I would suggest is improving the .npmrc a bit and optionally adding a small CI validation step to make sure these settings are actually active.

Suggested .npmrc:
```txt
min-release-age=7
ignore-scripts=true
save-exact=true
audit=true
```

Why these settings:
- `save-exact=true` avoids introducing new semver ranges when adding dependencies, making dependency updates more explicit.
- `audit=true` keeps npm audit behavior enabled for vulnerability checks.


> **One small caveat:** `ignore-scripts=true` can break packages that legitimately need install scripts. If the frontend build still passes with this enabled, that is great. If not, we may need to document the exception or choose a more targeted approach.

paulo-paes:
> I think this is a useful hardening step, but I would scope the title/description a bit more narrowly.
> 
> Right now this PR only adds npm-level protections under frontend/.npmrc, so it mitigates a subset of supply-chain risks for the frontend dependency installation path. It does not cover the Python/backend dependency chain yet. I would suggest renaming the PR to something more specific. And we could handle the Python side in a separate PR.

Yeah, initially I opened this PR to discuss hardening both frontend and backend. But I noticed in the last version that Tassio already did the changes for the backend. Now this PR will be focused to the frontend

> I would suggest is improving the .npmrc a bit and optionally adding a small CI validation step to make sure these settings are actually active.
> 
> Suggested .npmrc:
> 
> ```
> min-release-age=7
> ignore-scripts=true
> save-exact=true
> audit=true
> ```

Options added to .npmrc

> 
> 
>  **One small caveat:** `ignore-scripts=true` can break packages that legitimately need install scripts. If the frontend build still passes with this enabled, that is great. If not, we may need to document the exception or choose a more targeted approach.

Most dependencies do not rely on install scripts. I also checked the frontend dependencies specifically, and the only one with a post-install script is esbuild, which is **optional**.

Given that, I think adding `ignore-scripts=true` is the safer trade-off. The security benefit of preventing dependency install scripts from executing outweighs the risk of leaving this enabled just for an optional esbuild post-install script

Thank you for your collaboration @MarlonPassos-git

paulo-paes:
I also added the `.npmrc` configuration to the Dockerfile. I think that should be enough for this PR, so I’ll remove the draft status.

paulo-paes:
@tassionoronha Could you please check this PR?

Let me know if anything else needs to be adjusted

tassionoronha:
Sorry for the delay here and thanks for pushing it @paulo-paes ! LGTM now :)

Issue #197, opened 2026-05-16
[Bug] Dashboard Assets Widget Ignores Selected Month


geeosp:
Description
The "Assets Value" metric in the first dashboard widget always shows the current live asset value regardless of which month is selected in the top-right date picker. For historical months, the value also uses today's FX exchange rate instead of the rate at the end of the selected period, and includes assets that hadn't been purchased yet at that point in time.

Steps to Reproduce
Go to the Dashboard
Ensure at least one asset exists with historical value entries (e.g. a stock or real estate)
Select a past month using the top-right date picker
Observe the "Assets Value" metric in the summary widget at the top
Expected Behavior
The "Assets Value" metric should reflect the total value of assets as of the last day of the selected month, converted using the exchange rate valid at that date. Assets purchased after the selected period should not be included.

Actual Behavior
The widget always displays the current live asset price (using last_price × units for market-priced assets), ignoring the selected month entirely. FX conversion uses today's exchange rate. Assets purchased after the selected period are incorrectly included via the purchase_price fallback.

Environment
Deployment: Docker / Manual
Browser: Any
OS: Any
Screenshots
N/A

Additional Context
The same historical asset lookup logic already existed correctly in the Reports → Net Worth service, but the dashboard had its own independent implementation that was missing three constraints: the date boundary (always fetching the latest value instead of the value at the selected period), the FX rate date (always converting at today's rate instead of the period's closing rate), and the asset existence guard (including assets whose purchase date falls after the selected period).

https://github.com/user-attachments/assets/2f9c40a1-0646-4ec6-95ac-73c93c51860e

tassionoronha:
Hey @geeosp , indeed a bug we need to handle. 
I see that you opened a PR and closed it. 
Are you planning to continue this?

geeosp:
I closed my PR because it had a lot of tests failing.  I was not sure if was the one that caused it. Needed to investigate more. 
I am trying again

tassionoronha:
Fixed on #205.

Issue #199, opened 2026-05-16
[Bug] Historical Data for assets are displayed inconsistently across securo


geeosp:
## Description
As a user, I would like to understand how my Wallet grow over time. When looking specific to assets, Securo gives me divergent information in different screens.
I would expect that the value calculated for historical data for  assets from a specific month to be consistent anywhere in securo:
So far, I found those 3 places where they diverge
 
1. The dashboard first widget when a specific month is selected. This is partially  covered by this issue https://github.com/securo-finance/securo/issues/197
2. The Evolution  graphs in the Reports page in the same month
3. The total asset value calculated to the last day of that month in the Portfolio Value graph in the Assets page

I would expect those 3 situations to show the exact same amount for assets total value for a specific month. 

In the following screenshots, I am looking for the calculated total assets value for the month of March of 2026:

<img width="149" height="74" alt="Image" src="https://github.com/user-attachments/assets/9c544d86-2a4c-48dc-8062-dc214ff1396d" />
<img width="149" height="42" alt="Image" src="https://github.com/user-attachments/assets/960fb268-9306-4f50-ad6b-ccf66760a609" />
<img width="254" height="42" alt="Image" src="https://github.com/user-attachments/assets/741c7544-5fe9-49a3-abcc-67980612f7f6" />

<img width="1317" height="277" alt="Image" src="https://github.com/user-attachments/assets/37190940-033d-4115-9a4e-4c30e012e892" />

<img width="1317" height="376" alt="Image" src="https://github.com/user-attachments/assets/719e3411-406f-4b30-8fb7-5e03b6df0f87" />

<img width="1269" height="588" alt="Image" src="https://github.com/user-attachments/assets/6f4f7d8f-9d12-446d-a794-d6d2ab03749c" />

https://github.com/user-attachments/assets/6de44357-f7a7-4698-9b0e-4915e649f559

tassionoronha:
Confirmed. Different separate code paths compute asset totals with different date/FX/sold-asset rules. 
Fix: one shared "value as of date X" helper for all views, which also closes #197.  🙏

tassionoronha:
Fixed on #205.

PR #205, opened 2026-05-17
fix: align asset values and net worth evolution with selected period

geeosp:
- Dashboard assets widget now reflects the selected month's historical value instead of the current live price; uses the same cutoff date for FX conversion as the reports view, and guards the purchase_price fallback with a purchase_date <= cutoff check.

- Consolidate duplicate asset lookup logic into a single get_asset_values_at() in asset_service, replacing the separate get_total_asset_value() (dashboard) and _asset_value_at() (reports). goal_service updated to use the new function.

- Fix net worth evolution chart one-month offset: monthly snapshots were taken on the 1st of each month (= previous month's closing balance). Now snapshots use the last day of each month, capped at today for the current month.

## What

This should fix bugs https://github.com/securo-finance/securo/issues/197 and https://github.com/securo-finance/securo/issues/199

## Checklist

### DIFF
--- a/backend/app/services/asset_service.py
+++ b/backend/app/services/asset_service.py
@@ -175,6 +175,67 @@ async def _get_value_as_of(
     return result.scalar_one_or_none()
 
 
+async def _load_asset_native_values(
+    session: AsyncSession,
+    assets: list[Asset],
+    up_to_date: Optional[date] = None,
+) -> dict[str, list[tuple[date, float]]]:
+    """Bulk-load AssetValue rows for all assets in one query.
+
+    Returns {aid: [(date, amount), ...]} sorted ascending by (date, id).
+    When purchase_price and purchase_date are both set and purchase_date
+    predates the first recorded value, it is prepended as the earliest anchor.
+    """
+    if not assets:
+        return {}
+
+    asset_ids = [a.id for a in assets]
+    q = (
+        select(AssetValue.asset_id, AssetValue.date, AssetValue.amount)
+        .where(AssetValue.asset_id.in_(asset_ids))
+        .order_by(AssetValue.asset_id, AssetValue.date, AssetValue.id)
+    )
+    if up_to_date is not None:
+        q = q.where(AssetValue.date <= up_to_date)
+
+    rows = (await session.execute(q)).all()
+
+    values_map: dict[str, list[tuple[date, float]]] = {str(a.id): [] for a in assets}
+    for aid, d, amt in rows:
+        values_map[str(aid)].append((d, float(amt)))
+
+    for asset in assets:
+        aid = str(asset.id)
+        vals = values_map[aid]
+        if asset.purchase_price is not None and asset.purchase_date is not None:
+            if not vals or asset.purchase_date < vals[0][0]:
+                vals.insert(0, (asset.purchase_date, float(asset.purchase_price)))
+
+    return values_map
+
+
+def _fill_forward_at(
+    asset: Asset,
+    sorted_vals: list[tuple[date, float]],
+    as_of: date,
+) -> Optional[float]:
+    """Return the fill-forwarded native value of asset at as_of, or None.
+
+    Scans sorted_vals for the latest entry on or before as_of. Falls back
+    to purchase_price when purchase_date is None (asset predates any known
+    date) and no value history is available for the requested date.
+    """
+    result = None
+    for d, v in sorted_vals:
+        if d <= as_of:
+            result = v
+        else:
+            break
+    if result is None and asset.purchase_price is not None and asset.purchase_date is None:
+        result = float(asset.purchase_price)
+    return result
+
+
 async def _get_value_count(session: AsyncSession, asset_id: uuid.UUID) -> int:
     """Get the number of AssetValue entries for an asset."""
     result = await session.scalar(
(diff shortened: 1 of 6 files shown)


tassionoronha:
Hey @geeosp, one of the tests are failling here, can you fix that please? :)

geeosp:
Done, now I think it is in a great state

tassionoronha:
Thanks for this @geeosp,  tested it locally and both fixes work great: the assets widget now shows the right historical value per month, and the net worth chart points line up with each month's closing balance. 🙌

One small thing before merging: in `_date_points`, the line `current = date(start.year, start.month + 1, 1)` crashes for December start dates (`month + 1` → 13). Quick fix:

```python
next_month = start.month % 12 + 1
next_year = start.year + (start.month == 12)
current = date(next_year, next_month, 1)
```

tassionoronha:
Also quick note on #199: this fixes the dashboard and the Reports evolution graph (both now share get_asset_values_at 👍), but #199 also calls out a third place, the Portfolio Value graph on the Assets page. That one is powered by get_portfolio_trend, which this PR doesn't touch, so it still diverges.

So this fully resolves #197, but #199 needs get_portfolio_trend to use the same logic before it can be closed. Could be a follow-up if you'd rather keep this PR focused. Let me know :)

geeosp:
Don't worry, let me try to fix them

geeosp:
@tassionoronha Thanks a lot for the catches! The problem before  was only occurring with users  with multiple currencies, which wasn't the case of my personal account. _Yeah, I know,_ we have now a the testing account...  Should be using that. 
I wrote a test to catch that and was able to solve it. I also did a small refactor to both places so they  don't just share the same logic, they actually share the same code path as much as possible. Wrote some helper functions to do that. So, from now on, I expect them to be consistent and this pr solves both bugs now. 
Still, I would like to ask you to keep it open until we get into agreement on my next observation

geeosp:
I made this commit earlier today:
7421592310bd2ff0ce00edd14a203628819a2bec
just to make this specific test pass for the _date_points() 

```
def test_date_points_monthly():
    start = date(2025, 1, 1)
    end = date(2025, 6, 15)
    points = _date_points(start, end, "monthly")
    assert points[0] == start
    # Last point should be end (June 15), not June 1
    assert points[-1] == end
```

I believe this test is enforcing a strange behavior now.  I am not sure that "start" should be part of the returned values. 
In the way it is right now, 
If ask date points from January 1st to January 15th, January 1st is part of the returned points
if I ask date points from January 1st to February 16th , I would receive January 1st and February 16th. But from my point of view I would expect to receive January 31st and  February 16th in my date_points instead.

What I believe should happen we should receive 1 point per month, the most recent date, either the last day of the month, or the "end" date.

The consequence of the current behavior is that to any request of date_points of more than 3 months interval,  the time difference between the 1st and the 2nd dates will be of 2 months, while the consequent ones will be of 1 month until the end date
 

I wrote this test that reinforces this behaviour I believe is strange, but it comply with  the original test, so it is passing right now: 
```
def test_date_points_monthly_end_of_month_snapshots():
    """Monthly points: start date as boundary, then EOM snapshots for each subsequent month."""
    start = date(2025, 1, 1)
    end = date(2025, 4, 15)
    points = _date_points(start, end, "monthly")
    # start is the Jan boundary; loop begins from Feb, so Jan 31 is not emitted
    assert points == [
        date(2025, 1, 1),   # start boundary (Jan reference)
        date(2025, 2, 28),  # end of Feb
        date(2025, 3, 31),  # end of Mar
        date(2025, 4, 15),  # capped at end (Apr 30 → Apr 15)
    ]
```


For instance, in this recording below, we use the _date_points to define which dates to show, may 2025 and June 2025 will have a (unexpected) gap of historical data of 2 months, while all the others will have expected  gap of 1 month. Because the first date will be May 1st, and the second will be June 30th
So the first bar will show the value close to the value of the end of April, while the second bar shows the end of June 

https://github.com/user-attachments/assets/3c8b0da9-86e8-4aee-a4f0-e2b1e929a6e8


So my question is, what we do, what should be the desired behavior?
1. We do nothing, we are happy with the way it is and we can close this PR
2. We enforce exactly one point per month, being the last day of the month or the end date, regardless if it is the first month of the interval.
3. We keep the start date as first point, but enforce that the end of the month appers in the list in case it is earlier than the end date.

I vote for 2, but I will need to update both tests to comply to that.

tassionoronha:
Beautiful work on the follow-ups @geeosp  🚀 

Verified live on my side dashboard and Assets chart now match and December fix confirmed.

**Vote 2**: one rule per bar beats a special case.

Heads up before swapping the tests: `trend[0]` is also the baseline for the "vs início do período" delta in both net worth and income reports. If you drop start from `_date_points`, compute the baseline separately via `_net_worth_at(start)` so that percentage stays accurate. 

We are getting there. Closing 2 issues directly related to Securo's accuracy.

geeosp:
Ok! Updated the monthly points, the tests and the usage in the report
funcions!

Em qua., 20 de mai. de 2026, 12:50, Tássio Noronha ***@***.***>
escreveu:

tassionoronha:
Such an amazing work here @geeosp .
Tested on my side and we are good to go! 
Congrats!! 🙏🏽

PR #208, opened 2026-05-18
fix: reduce celery-worker RAM usage from ~1.2 GB to ~300-500 MB

### LINKED ISSUE #207: [Bug] Celery worker consuming excessive RAM (~1.2 GB) in production
## Description

The `celery-worker` service in `docker-compose.prod.yml` is consuming approximately **1.2 GB of RAM**, while the entire project uses around 1.4 GB. This is caused by a combination of:

1. **Unbounded Celery concurrency** — the worker command does not set `--concurrency`, so Celery defaults to spawning one process per CPU core. On hosts with 8+ cores, this creates 8+ prefork workers, each loading the full Python application stack (FastAPI, SQLAlchemy, yfinance, fastembed/ONNX, etc.).
2. **Missing `engine.dispose()` calls** in 4 out of 5 Celery task files (`asset_tasks.py`, `fx_rate_tasks.py`, `recurring_tasks.py`, `fx_backfill_tasks.py`). These tasks create a fresh SQLAlchemy `AsyncEngine` per run but never close it, leaking asyncpg connections and memory over time.
3. **`sync_tasks.py` using the global FastAPI engine** (`app.core.database.async_session_maker`). Under Celery's prefork model, the parent process loads the module and then forks children, sharing (and corrupting) the same connection file descriptors.

## Steps to Reproduce

1. Run the project with `docker compose -f docker-compose.prod.yml up`.
2. Observe the `celery-worker` container memory usage via `docker stats`.
3. Notice it climbs to ~1.2 GB and stays high even when idle.

## Expected Behavior

The `celery-worker` should use a bounded, reasonable amount of memory (target: ~300-500 MB) and should not leak database engines after each scheduled task.

## Actual Behavior

- `celery-worker` uses ~1.2 GB RAM consistently.
- The base image is the same as the backend, so the worker loads the full FastAPI/uvicorn stack unnecessarily.
- No `--concurrency` flag is set, so the worker spawns as many processes as CPU cores.

## Environment

- Deployment: Docker Compose (`docker-compose.prod.yml`)
- OS: Ubuntu 22.04 / Any Linux host with 8+ CPU cores exposed to Docker

## Additional Context

- `ingest.py` is the only task file that already calls `await engine.dispose()` correctly.
- The `fastembed` embedding model (~120 MB ONNX) can be loaded per worker process when agents are enabled, amplifying the issue.

lgwacker:
## Problem

The `celery-worker` service in production (docker-compose.prod.yml) was consuming approximately **1.2 GB of RAM**, while the entire project stack used only ~1.4 GB. Investigation revealed three root causes:

1. **Unbounded Celery concurrency** — the worker command did not set `--concurrency`, so Celery defaulted to spawning one prefork worker per CPU core. On hosts with 8+ cores, this created 8+ worker processes, each loading the full Python stack (FastAPI, SQLAlchemy, yfinance, fastembed/ONNX, etc.).
2. **Missing `engine.dispose()` calls** — 4 out of 5 Celery task files created a fresh SQLAlchemy `AsyncEngine` per run but never closed it, leaking asyncpg connections and memory over time.
3. **`sync_tasks.py` using the global FastAPI engine** — under Celery's prefork model, the parent process loaded the global engine and forked children, causing shared (and corrupted) connection file descriptors.

### 1. Limit Celery concurrency in Docker Compose
- Added `--concurrency=2` to the `celery-worker` command in both `docker-compose.yml` and `docker-compose.prod.yml`.
- This bounds the number of prefork workers regardless of how many CPU cores the host exposes to Docker.

### 2. Dispose SQLAlchemy engines after every Celery task
- Changed `_make_session_maker()` in `asset_tasks.py`, `fx_rate_tasks.py`, `recurring_tasks.py`, and `fx_backfill_tasks.py` to return the `(engine, session_maker)` tuple.
- Wrapped every async task body in `try ... finally: await engine.dispose()`, ensuring connections are released immediately after each task run.

### 3. Isolate `sync_tasks.py` from the FastAPI engine
- Rewrote `sync_tasks.py` to create its own local `create_async_engine()` instead of importing the global `async_session_maker` from `app.core.database`.
- This eliminates forked-connection corruption and keeps the worker self-contained.

## Expected Impact

| Metric | Before | After |
|--------|--------|-------|
| `celery-worker` RAM | ~1.2 GB | ~300–500 MB |
| DB connection leaks | Yes (engine never disposed) | No |
| Prefork safety with global engine | Risky | Safe |

## Checklist

- [x] Issue #207 opened and linked
- [x] Docker Compose files updated for both dev and prod
- [x] All 5 Celery task files fixed
- [x] `python -m py_compile` passes on all modified Python files

Closes #207

### DIFF
--- a/backend/app/tasks/asset_tasks.py
+++ b/backend/app/tasks/asset_tasks.py
@@ -19,7 +19,7 @@ def _make_session_maker():
     """Create a fresh engine+session for the Celery worker event loop."""
     settings = get_settings()
     engine = create_async_engine(settings.database_url)
-    return async_sessionmaker(engine, class_=AsyncSession, expire_on_commit=False)
+    return engine, async_sessionmaker(engine, class_=AsyncSession, expire_on_commit=False)
 
 
 def _next_due_date(last_date: date, frequency: str) -> date:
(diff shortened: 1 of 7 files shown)


tassionoronha:
Awesome first PR @lgwacker, and a really solid one! The fix is spot on, all three causes handled the right way.

Two small things before merge:

1. The diff has a fair bit of unrelated reformatting (compose quotes, import order, call wrapping). Could you drop those so it shows just the fix? Easier to review and keeps git blame clean.

2. The compose comment is in Portuguese, mind switching it to English to match the rest?
Also, did you get to check the before/after with docker stats? Would love to see your real numbers.

Thanks for jumping on this 🙌

lgwacker:
Removed the cometic changes. Do you mind sharing the code formatting in this repo? I will need to set this or those agents are gonna make my life hard xD


Before:
image


After:
image


Difference is not huge because both were on a fresh start. But after a while, it really starts compouding.

tassionoronha:
Looks great now, thanks for the quick turnaround and for sharing you results :) 

On formatting: the repo only lints, it doesn't enforce a formatter. CI just runs ruff check ., and there's no [tool.ruff] config or ruff format/Black in use. So the fix is to run ruff check . (matches CI) and turn off ruff format / format-on-save in your editor, otherwise it'll keep rewriting whole files. I do need to improve it later.

Merging this. Thanks again for the solid first contribution 🙌

PR #214, opened 2026-05-20
feat: add ignore flag for transactions (closes #200)

### LINKED ISSUE #200: [Feature] Rule Condition or Category Option to delete/ignore transactions
## Problem

I started syncing my credit card via Pluggy. For some reason OpenFinance registers the credit card payment twice so theres 3 transactions:

1. Checking Account:  Sends out X $ to pay credit card bill (This is labeled as Internal Transfer as it should)
2. Credit Card: Receives X $ from Checking (Also labeled internal transfer correctly)

Now heres the issue:
3. Credit Card: Receives another X $, that is not labeled as internal transcation so it just makes it look like i paid my credit card twice.

## Proposed Solution

While i was testing, i realized that all i had to do was just delete the 3. transaction and everything would work fine, the issue is as soon as Pluggy syncs again it re-fetches all transactions i deleted. My first thought was to make a Rule to ignore these transactions but there isn't an option for that.

My proposal is to either add a Rule action to delete/ignore/set inactive these transactions or add an option to Categories to ignore the whole category just like the toggle to "Treat this group as internal transfer".

## Alternatives Considered

I also tried to set the group as internal transfer but it did not work, it would still consider the payment twice not reflecting my actual credit card bills

## Additional Context

I don't have a lot of experience with financing apps and with credit cards at all. If im misjudging this or if its intended  please let me know. Also, not sure if this matters, but my Securo setup has 2 individuals (me and wife) under a single Securo account.

gaamelu:
## Problem

When Pluggy syncs, it re-fetches deleted transactions — making it impossible to permanently remove duplicates like the double credit-card payment described in #200. Users need a way to suppress a transaction without deleting it.

## Solution

Adds a persistent `is_ignored` flag to transactions. Ignored transactions survive re-syncs unchanged and are hidden from views/totals by default.

## Changes

**Backend**
- Migration `051`: adds `is_ignored` boolean column (default `false`) to `transactions`
- `PATCH /transactions/{id}/ignore` — toggles ignored state
- `PATCH /transactions/bulk-ignore` — bulk toggle for selected transactions
- `GET /transactions` — new `ignored_filter` query param: `hide` (default) | `include` | `only`
- Rule engine: new `ignore` action type to auto-ignore matching transactions on sync

**Frontend**
- Filter bar: "Ignored" submenu with hide / show all / only options
- Ignored rows: 40% opacity + strikethrough description + "Ignored" badge
- Hover on table row: EyeOff button to ignore/unignore inline
- "Only ignored" view: inline clickable badge to restore the transaction
- Transaction dialog footer: ignore/unignore action button
- Selection bar: bulk "Ignore" button for selected rows

## Test plan

- [ ] Sync an account — ignored transactions stay ignored after re-sync
- [ ] Toggle ignore via dialog footer, row hover, and bulk selection
- [ ] Filter bar: hide / show all / only ignored each show correct rows
- [ ] Ignored transactions excluded from income/expense totals when hidden
- [ ] Rule with "ignore" action auto-ignores matching new transactions
- [ ] Bulk ignore + bulk restore updates counts correctly

Fixes #200

### DIFF
--- a/backend/alembic/versions/051_transaction_is_ignored.py
+++ b/backend/alembic/versions/051_transaction_is_ignored.py
@@ -1,0 +1,30 @@
+"""add is_ignored to transactions
+
+Revision ID: 051
+Revises: 050
+Create Date: 2026-05-19
+
+"""
+from typing import Sequence, Union
+
+from alembic import op
+import sqlalchemy as sa
+
+
+revision: str = '051'
+down_revision: Union[str, Sequence[str], None] = '050'
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+
+def upgrade() -> None:
+    op.add_column(
+        'transactions',
+        sa.Column('is_ignored', sa.Boolean(), nullable=False, server_default='false')
+    )
+    op.create_index('ix_transactions_is_ignored', 'transactions', ['is_ignored'])
+
+
+def downgrade() -> None:
+    op.drop_index('ix_transactions_is_ignored', table_name='transactions')
+    op.drop_column('transactions', 'is_ignored')
--- a/backend/app/api/transactions.py
+++ b/backend/app/api/transactions.py
@@ -12,7 +12,7 @@ from sqlalchemy.ext.asyncio import AsyncSession
 from app.core.auth import current_active_user
 from app.core.database import get_async_session
 from app.models.user import User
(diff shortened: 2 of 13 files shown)


tassionoronha:
Hey @gaamelu, welcome and thanks for your first PR!

After analyzing your #214 alongside #215 (both addressing #200), I decided to continue the work on #215, picking the implementation pieces that were missing there from your PR: the sync-time guard to skip ignored transactions on Pluggy re-sync, and the rule-engine `ignore` action.

Added you as co-author on those commits in #215. Thanks again for collaborating, and looking forward to your next PRs :)

PR #215, opened 2026-05-20
feat: ignore transactions/categories (issue #200)

### LINKED ISSUE #200: [Feature] Rule Condition or Category Option to delete/ignore transactions

igorfrag:
## What

Adds is_ignored flags for transactions and categories. This flag is used as .where(is_ignored) in SQL queries preventing them from being counted towards reports and balance.

## Why

Resolves #200 

## How to Test

1. Create a transaction and set it to Ignored on dialog
2. It shows up on your transaction list with a grey color and won't be counted on balances and reports.
3. You can also assign a category as "ignored" just like "transfer" categories.

## Checklist

## Examples:
Main page:
image

Transactions:
image

Transaction Dialog:
image

Accounts: 
image

Category:
image

### DIFF
--- a/backend/alembic/versions/050_is_ignored_category_and_transaction.py
+++ b/backend/alembic/versions/050_is_ignored_category_and_transaction.py
@@ -1,0 +1,58 @@
+"""Add is_ignored flag to categories and transactions
+
+Revision ID: 050
+Revises: 049
+Create Date: 2026-05-19
+
+When is_ignored is True:
+- For transactions: excluded from reports, dashboard aggregations, and budget calculations
+- For categories: transactions in these categories are excluded from the above
+"""
+
+from alembic import op
+import sqlalchemy as sa
+
+revision = "050"
+down_revision = "049"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    # Add is_ignored column to categories table
+    op.add_column(
+        "categories",
+        sa.Column(
+            "is_ignored",
+            sa.Boolean(),
+            nullable=False,
+            server_default=sa.text("false"),
+        ),
+    )
+
+    # Add is_ignored column to transactions table
+    op.add_column(
+        "transactions",
+        sa.Column(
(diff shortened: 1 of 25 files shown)


tassionoronha:
Thanks for the PR, @igorfrag! Solid base to build on. I pushed a few follow-ups on top:

- Brought in some pieces from @gaamelu's #214 since both PRs were tackling #200: the sync-time guard so Pluggy can't revive ignored rows, the rule-engine `ignore` action plus its UI, and a one-click toggle button in the dialog footer (replacing the inline checkbox).
- Fixed a small lint issue (unused import).
- Added some tests.

Let me know if this looks good on your end and we're good to merge.

igorfrag:
Looking good!

Issue #219, opened 2026-05-21
[Feature Request?]  Networth Ui Improvements


geeosp:
Hello everyone, 

I feel that Net Worth Report page is a little off.

I found the Evolution graph from the Net Worth page to be very misleading. 
The way that Liabilities is placed in the graph bars is confusing. It is open to interpretation. My understanding is that it is implying either that the Liabilities are a value that contributes positively to the Net Worth, or it is a negative value being subtracted from the assets amount.
The first sounds very wrong, and the second just raised the question whether or not  the assets part was being displayed correctly., because  why would the liabilities only be subtracted from the assets, and not from the accounts too?

<img width="619" height="362" alt="Image" src="https://github.com/user-attachments/assets/22282573-6adf-4d96-a26d-3426a9ea1d60" />

The same concept apply to the Composition pie chart. 
Here the Liabilities clearly sums up to the 100% of the graph. What it communicates is that accounts+assets+liabilities = networth.

<img width="365" height="433" alt="Image" src="https://github.com/user-attachments/assets/2dd44814-6c70-4e5c-821c-01f8609762da" />

In addition to that, there is a missing information that I cannot find anywhere in Securo right now. This very simple question: "How much money did I earned, or lost, this month?" 
The Net Worth page seems to be the place to show this info, and there is the graph that shows the snapshot, but we still have to do some math to get the difference between periods. All the other places are only considering accounts, or assets, or liabilities, but never all of them together, never the full picture.

geeosp:
So, this is what I am thinking about what we can do to this page:


1. Networth - Over time graph:
	Add a waterfall chart overlay to the Networth over time that clearly states your net worth against the last period date. Want to know how much your net worth increased or decresed? Just mouse hover the bar.

<img width="1070" height="405" alt="image" src="https://github.com/user-attachments/assets/ae4d87f5-3622-4db0-b8f0-8ff6a1a5b3c4" />


2. Composition pie charts:
	I Separated the Liabilities from the same pie chart from Accounts and Assets, creating a second one. One represent values that add up to the net worth, other represent the ones that subtracts from it. 

<img width="372" height="395" alt="image" src="https://github.com/user-attachments/assets/f8355708-79d6-4d9b-890f-19d17c5198a4" />

3. Evolution  bar charts:
	Updated  bars to grow to the negative values to the negative  vertical axis of chart. This way it is clear that they contributes by subtracting from the total ammount. 
	Then add the Networth graph overlay to reinforce the idea.

[...]

PR #220, opened 2026-05-21
Networth UI improvements

geeosp:
Hi everyone, here is a group of suggestions on how to tackle  #219 , I would love to hear what you think

1. Networth - Over time graph:
	Add a waterfall chart overlay to the Networth over time that clearly states your net worth against the last period date. Want to know how much your net worth increased or decresed? Just mouse hover the bar.

image

2. Composition pie charts:
	I Separated the Liabilities from the same pie chart from Accounts and Assets, creating a second one. One represent values that add up to the net worth, other represent the ones that subtracts from it. 

image

3. Evolution  bar charts:
	Updated  bars to grow to the negative values to the negative  vertical axis of chart. This way it is clear that they contributes by subtracting from the total ammount. 
	Then add the Networth graph overlay to reinforce the idea.

image

4. The Composition + pie charts.
	The composition + the pie chart were looking like they were  related, but interacting with one did not affected the other. The "detailed / summary" toggle just affect one of them, while one showed the current composition, while the other showed the history.
		So I made the toggle to affect both of them.
		In detail mode, every asset/account/liability is now sorted by value inside its respective group, so its easy to understand which asset/account contributes the most for the final result.
		I updated the colors so its easy to identify what asset/account/liability belong in which group.

image

	Also,  added a new interaction, by clicking in a bar on the evolution graph, the composition bar update to show info about the current bar selected. By clicking again, it deselect the bar and fallback for the default state, which is to show the composition of the today's net worth. By not interacting with the graph, it haves the same behavior as before. 

https://github.com/user-attachments/assets/c30af55f-9073-49a5-bbe8-398fc144e4ea

### DIFF
--- a/backend/app/services/report_service.py
+++ b/backend/app/services/report_service.py
@@ -1,7 +1,9 @@
 import calendar
 import uuid
+from collections import defaultdict
 from datetime import date, timedelta
 from decimal import Decimal
+from typing import Optional
 
 from sqlalchemy import String, select, desc, func, case
 from sqlalchemy.ext.asyncio import AsyncSession
@@ -10,6 +12,7 @@ from app.core.config import get_settings
 from app.models.account import Account
 from app.models.asset import Asset
 from app.models.asset_value import AssetValue
+from app.models.fx_rate import FxRate
 from app.models.transaction import Transaction
 from app.models.category import Category
 from app.models.user import User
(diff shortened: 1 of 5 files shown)


geeosp:
Sorry guys, I will close this PR, This is more of a study than a finished full feature PR, I will move the discussion to the issue page #219

tassionoronha:
Hey @geeosp, took a quick look. The improvements are nice, the breakdown in particular. That said, with the number of assets I have, got hard to read, which made me think we'll eventually need more flexibility in how reports are built: granularity, filters, limits, and ideally letting users save their own views. 

Not proposing we build a BI tool at this point, just leaving the seed for a follow-up discussion.

geeosp:
Yeah, I see your point. I have made some modifications, it should have a
better readability now. To be honest, I wasn’t too happy about the double
donuts. To your point, I went back to grouping it by wallet. It indeed
looks better now. I will post the results later.
I also believe we don't need to go full BI, The current views in escuro are
already very powerful, but filtering and grouping could leverage that by
bringing in a lot of flexibility without obligating the user to design
their own views.

PR #251, opened 2026-06-02
feat: create rule from transaction

geeosp:
Hey Everyone, 
I thought about letting us to create a Category Rule from a Transaction, letting us to categorize our transactions in a quicker way.  So this is what this PR does: 

Add a "Create Rule" shortcut to the transactions UI, accessible from two places: the bulk action bar (when one transaction is selected) and the transaction edit dialog footer.
Opening it pre-fills the rule dialog with the transaction's description as a "contains" condition and its category as the default action, so the user only needs to name the rule and save.

**What**

- Extracted RuleDialog from rules.tsx into a shared component (rule-dialog.tsx) with an optional initialData prop for pre-population
- Added a "Create Rule" button to the bulk action bar on the Transactions page — visible when exactly one non-shared transaction is selected
- Added a "Create Rule" button to the transaction edit dialog footer — visible when editing an existing transaction
- Pre-fills the rule dialog with the transaction's description (contains condition) and payee (equals condition, if set), plus the category as the default set_category action and the related tags
- Extracted RuleDialog from rules.tsx into a shared component (rule-dialog.tsx) with an optional initialData prop for pre-population
- Added transactions.createRule i18n key in en.json and pt-BR.json

**Why**
Creating a categorization rule from an existing transaction required the user to leave the Transactions page, navigate to Rules, and manually type in the conditions from memory. This shortcut eliminates that friction by letting users create rules directly from any transaction they're looking at, with the relevant fields already filled in.

**How to Test**
Go to the Transactions page
Check the checkbox on any single transaction row — the bulk action bar slides up at the bottom
Click "Create Rule" (wand icon) in the action bar
Verify the rule dialog opens pre-filled with the transaction's description as a contains condition and its category pre-selected as the action
Name the rule and save — confirm it appears on the Rules page
Open any transaction by clicking its row to open the edit dialog
Verify the "Create Rule" button appears in the footer alongside Delete/Ignore
Click it — confirm the transaction dialog closes and the rule dialog opens pre-filled with the same transaction's data
Verify the button does not appear on shared (group) transactions

## Checklist


tassionoronha:
Great idea @geeosp. Thanks for this one :)

PR #273, opened 2026-06-05
feat: YTD reporting period implementation

z0diak:
## What

Adds a YTD reporting period option to historical reports. The reports UI now shows `6M, YTD, 1Y, 2Y`, and YTD requests data from January 1 of the current year through today while preserving the selected granularity (`D`, `W`, `M`, `Y`).

## Why

Users who plan/review their finances on a calendar-year basis need a true year-to-date view for reporting. Rolling 12-month data can start in the prior year, which does not match YTD expectations.

## How to Test

1. Open the Reports page.
2. Select `YTD` and switch granularity between `D`, `W`, `M`, and `Y`.
3. Confirm chart data starts at the beginning of the current year.
4. Compare with `1Y` and confirm it still uses the rolling 12-month range.
5. Verify Net Worth and Income vs Expenses both respect YTD.

## Checklist

### DIFF
--- a/backend/app/api/reports.py
+++ b/backend/app/api/reports.py
@@ -13,11 +13,18 @@ router = APIRouter(prefix="/api/reports", tags=["reports"])
 async def get_net_worth(
     months: int = Query(12, ge=1, le=24),
     interval: str = Query("monthly", pattern="^(daily|weekly|monthly|yearly)$"),
+    period: str | None = Query(None, pattern="^ytd$"),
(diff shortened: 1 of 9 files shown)


tassionoronha:
Hey @z0diak thanks so much for this one and welcome here :)

Really clean PR. 
Tests, translations, and the frontend refactor are all spot on, and it works great when I tested it locally. 
Merging now. 🎉

PR #275, opened 2026-06-05
feat: add OIDC login support

### LINKED ISSUE #258: [Feature] OIDC login support
## Problem

As an avid self-hoster with a fairly complex homelab setup, I would like to use OIDC login from my self-hosted provider (Authentik). All of the other apps I currently host are able to utilize OIDC for logging in, and I use a single way of logging in to everything, with Securo being the outlier. 

## Proposed Solution
Implement OIDC login support. It will make the use of the app smoother, and more secure as well, adding (depending on the provider) TOTP and passkey support.

## Alternatives Considered

Alternatively, to improve the security of the app, TOTP and passkeys could be implemented standalone into Securo. It would be beneficial to use them alongside OIDC further down the line as well.

## Additional Context

n/a

RomuloGatto:
## What

Adds generic OIDC login support for Securo, with provider discovery and a frontend login flow compatible with Authentik, Pocket ID, and other standard OIDC providers.

This includes:
- OIDC discovery, authorization redirect, token exchange, ID token validation, and JWKS verification
- Redis-backed state + nonce validation for the auth flow
- Securo JWT issuance after a successful OIDC callback
- Auto-provisioning for new verified-email OIDC users, configurable via env
- Optional OIDC role/group synchronization behind `OIDC_SYNC_ROLES`
- Provider-role to Securo admin + workspace-role mapping
- Login page OIDC button and callback page
- Env examples, Compose wiring, README docs, and translations

## Why

Closes #258.

Self-hosters commonly centralize authentication through Authentik, Pocket ID, or similar OIDC providers. This lets Securo fit into those homelab SSO setups while preserving existing local login and 2FA behavior for users who do not enable OIDC.

## How to Test

1. Configure an OIDC provider with redirect URI `https://your-securo-host/api/auth/oidc/callback`.
2. Set env vars, for example:
   ```env
   OIDC_ENABLED=true
   OIDC_PROVIDER_NAME=Pocket ID
   OIDC_DISCOVERY_URL=https://id.example.com/.well-known/openid-configuration
   OIDC_CLIENT_ID=securo
   OIDC_CLIENT_SECRET=your-client-secret
   ```
3. Optional role sync:
   ```env
   OIDC_SYNC_ROLES=true
   OIDC_ROLES_CLAIM=groups
   OIDC_ADMIN_ROLES=securo-admins
   OIDC_WORKSPACE_ROLE_MAP={"securo-owners":"owner","securo-editors":"editor","securo-viewers":"viewer"}
   ```
4. Restart Securo and use the new OIDC button on the login page.
5. Verify a successful login returns to Securo and creates/updates the local user.

Local verification run:

```bash
cd backend && /opt/homebrew/bin/uv run --extra dev pytest tests/test_oidc_auth.py tests/test_auth_api.py -q
cd backend && /opt/homebrew/bin/uv run --extra dev ruff check app/api/oidc_auth.py tests/test_oidc_auth.py
cd frontend && PATH="/opt/homebrew/bin:$PATH" npm run lint -- --quiet
cd frontend && PATH="/opt/homebrew/bin:$PATH" npm run build
```

## Checklist


tassionoronha:
Hey @RomuloGatto, nice work here! IMO it's clean and well scoped, and the optional feature wiring is exactly right. Two things I'd fix before merge. I confirmed both with quick repro tests:

1. The userinfo response can override the verified id_token. Right now merged = {**claims, **userinfo} lets the unsigned userinfo win, so I was able to make a user admin just by putting admin groups in userinfo. Flip it to {**userinfo, **claims} so the signed token wins, and check that userinfo["sub"] == claims["sub"] before trusting it.
  
2. OIDC_REQUIRE_VERIFIED_EMAIL can be bypassed when the provider doesn't send the claim at all. email_verified is False only catches an explicit false, so a missing value slips through and the user still gets created. Change it to email_verified not in (True, "true") so it fails closed.

Can you please check those?

tassionoronha:
Also pinging @dmaniaa here to have a validation check.

dmaniaa:
> Also pinging @dmaniaa here to have a validation check.

Noted. Will take a swing at it, most likely tomorrow. I’ll try to reproduce the conditions in your notes as well.

RomuloGatto:
Tks for the feedbacks guys! I'll try to complete it later today

RomuloGatto:
Both issues are fixed in the latest push.

Changes made:
- userinfo `sub` is now checked against the signed id_token `sub`; mismatches fail with 400 before any user lookup/provisioning.
- merged claims now use `{**userinfo, **claims}`, so signed id_token claims win over userinfo for overlapping fields like email, email_verified, and groups.
- `OIDC_REQUIRE_VERIFIED_EMAIL` now fails closed with `email_verified not in (True, "true")`, so missing claims no longer pass.
- Added regression coverage for userinfo role escalation, userinfo sub mismatch, and missing email_verified.
- Verification rerun:
- `cd backend && /opt/homebrew/bin/uv run --extra dev pytest tests/test_oidc_auth.py tests/test_auth_api.py -q` → 14 passed
- `cd backend && /opt/homebrew/bin/uv run --extra dev ruff check app/api/oidc_auth.py tests/test_oidc_auth.py` → passed
- `cd frontend && PATH=\"/opt/homebrew/bin:$PATH\" npm run lint -- --quiet` → passed
- `cd frontend && PATH=\"/opt/homebrew/bin:$PATH\" npm run build` → passed

RomuloGatto:
Pushed another security-hardening update for the OIDC flow.

Changes included:

1. Existing local users are no longer silently claimed by OIDC email match by default.
   - Added `OIDC_EXISTING_USER_LINK_MODE=disabled|verified_email`.
   - Default is `disabled`.
   - With the default, an existing Securo account cannot be logged into by a new OIDC identity just because the email matches.
   - `verified_email` can be enabled by self-hosters who intentionally want first-login migration/linking from a trusted IdP.

2. OIDC identities are now durably bound by `(issuer, subject)`.
   - Added `oidc_issuer` and `oidc_subject` to users.
   - Added Alembic migration `055_oidc_identity_linking.py`.
   - Login resolution now first checks the immutable OIDC identity instead of relying on email.
   - If a local account is already linked to a different OIDC subject, login is rejected.

3. Personal workspace ownership is protected.
   - OIDC workspace role sync no longer downgrades a user’s own personal workspace from `owner` to `editor`/`viewer`.

4. Added regression tests for:
   - rejecting unlinked existing users by default.
   - allowing verified-email linking only when explicitly configured.
   - rejecting a different OIDC subject for an already-linked account.
   - preventing personal workspace owner downgrade.

Verification:
- `uv run pytest tests/test_oidc_auth.py -q` → 11 passed
- `uv run pytest tests/test_oidc_auth.py tests/test_workspaces_api_coverage.py -q` → 30 passed
- `uv run ruff check app/api/oidc_auth.py app/core/config.py app/models/user.py tests/test_oidc_auth.py alembic/versions/055_oidc_identity_linking.py` → passed
- `python -m py_compile app/api/oidc_auth.py app/core/config.py app/models/user.py alembic/versions/055_oidc_identity_linking.py` → passed
- `git diff --check` → passed

dmaniaa:
I was not able to go into the security details, as I don't have the time currently, or the tools at hand. I tested with my existing, "production" Authentik instance only.
Login and registration work, role claims work.
I do have a few things to consider:
1. The current OIDC implementation completely ignores the "Allow user registration" Securo setting. Despite registration being disabled, a new user can still register via OIDC. I believe this should not happen, if registration is not allowed, it should not be allowed at all.
2. Linking existing users by only verified e-mail feels a little restrictive. Hardening is good, when it's an option, not forced. I don't use e-mail verification at all in my setup, the security is relaxed as it's not public facing. Because of this, I was not able to link my existing Securo account at all, and I had to remove the account registered to my e-mail and re-create it via OIDC, which would be a problem when this version is released and the users will update, wanting to keep their data. I'd appreciate the option to just link the existing accounts by e-mail without verification as a setting.

RomuloGatto:
Thanks for testing with Authentik — both points make sense.

I pushed a follow-up update addressing them:

1. OIDC registration now respects Securo’s global “Allow user registration” setting.
   - If registration is disabled, new users cannot be created via OIDC either.
   - This is enforced even when `OIDC_AUTO_REGISTER=true`.

2. Existing-account linking now has a less restrictive option.
   - `OIDC_EXISTING_USER_LINK_MODE` now supports:
     - `disabled` — default, safest; do not auto-link existing accounts by email.
     - `verified_email` — link existing accounts by email only when the IdP provides `email_verified=true`.
     - `email` — link existing accounts by matching email even without `email_verified`.
   - This keeps the hardened default, but allows relaxed private/self-hosted deployments like yours to migrate existing Securo users without recreating their accounts/data.

The durable `(issuer, subject)` binding is still used after linking, so email matching is only used for the first link/migration step.

Added regression tests for:
- OIDC respecting disabled registration.
- `email` link mode linking an existing account without `email_verified`.

Verification:
- `uv run pytest tests/test_oidc_auth.py tests/test_workspaces_api_coverage.py -q` → 32 passed
- `uv run ruff check ...` → passed
- `uv run python -m py_compile ...` → passed
- `git diff --check` → passed

dmaniaa:
Okay, linking looks good. 
With OIDC_REQUIRE_VERIFIED_EMAIL=true I wasn't able to register without verification.
With OIDC_REQUIRE_VERIFIED_EMAIL=false & OIDC_EXISTING_USER_LINK_MODE=verified_email I was able to register a new account, but I wasn't able to link an existing account.
With OIDC_REQUIRE_VERIFIED_EMAIL=false & OIDC_EXISTING_USER_LINK_MODE=email I was able to register a new account and I was able to link an existing account.
And after registering or linking an account over OIDC and setting OIDC_REQUIRE_VERIFIED_EMAIL=true & OIDC_EXISTING_USER_LINK_MODE=verified_email again, I'm not able to log in anymore. I understand the intention here is to disallow both registration and login without a verified e-mail. 

Please reconsider if the e-mail verification should be used for both registration and login, or for registration only. Currently, the verification on both login and registration is okay, but registration only sounds more sane, as after an account is linked, the link could be considered "trusted".

The "Allow user registration" option, in the frontend, is still ignored. I don't have the time for a code deep dive right now, but I gave Claude Opus 4.8 the PR and my report for review, and it says this about the specific bug:

> The OIDC registration gate reads settings.registration_enabled (the env default in config.py), but the frontend "Allow user registration" toggle persists to the DB and is read via admin_service.is_registration_enabled(session) — that's what the local /register guard (check_registration_enabled) uses. The OIDC path should call the same admin_service.is_registration_enabled(session) instead of the static settings value. session is already available in _get_or_create_oidc_user.

Worth taking a look.

tassionoronha:
Huge thanks both of you guys for the thorough testing and to be investing on the OIDC 🙏 
Agreed that registration only verification feels saner. Let me know if you need help here too @RomuloGatto

RomuloGatto:
Yeah, that makes sense.

I agree that after an account has already been linked by the OIDC identity, the link itself can be considered trusted. I’ll adjust the flow so `email_verified` is required only when Securo is using email as a trust bridge: new registration, or first-time existing-account linking in `verified_email` mode. Already-linked users should be resolved by `(issuer, sub)` and should not be locked out later just because the IdP does not emit `email_verified`.

The registration toggle bug is real also. The OIDC path currently checks the static `settings.registration_enabled`, but the frontend/admin toggle persists to the DB-backed app setting. I’ll change OIDC new-user creation to use the same `admin_service.is_registration_enabled(session)` path as the regular `/register` guard.

Intended behavior after the fix:

- Existing linked user by `(issuer, sub)`:
  - login allowed without re-checking email verification.
- Existing unlinked user by email:
  - `disabled`: reject.
  - `verified_email`: link only with `email_verified=true`.
  - `email`: link by email without requiring `email_verified`.
- Brand-new OIDC user:
  - blocked if the DB-backed registration setting is disabled.
  - blocked if `OIDC_AUTO_REGISTER=false`.
  - blocked if `OIDC_REQUIRE_VERIFIED_EMAIL=true` and email is not verified.

I’ll add regression tests for these cases as well

dmaniaa:
Brilliant! 
Verified: existing linked user, existing unlinked user, and brand-new OIDC user on each case you mentioned, I now see the expected behaviour. 
No further requests from me.

tassionoronha:
Nice collaborative work guys 🙌🏼

Pushed one commit to add PKCE (S256) to the auth code flow (/login now sends code_challenge, the verifier is stored with the nonce in Redis and replayed on token exchange). This closes the last hardening gap I had, code-interception
protection on top of the state/nonce checks.

If you're happy with that I'm ready to merge.

b52src:
just a note

the OIDC link to existing user info is missing in the docs and .env.example, had to hunt down this issue to find documentation on how to fix linking

OIDC_EXISTING_USER_LINK_MODE now supports:
disabled — default, safest; do not auto-link existing accounts by email.
verified_email — link existing accounts by email only when the IdP provides email_verified=true.
email — link existing accounts by matching email even without email_verified.
This keeps the hardened default, but allows relaxed private/self-hosted deployments like yours to migrate existing Securo users without recreating their accounts/data.

tassionoronha:
> just a note
> 
> the OIDC link to existing user info is missing in the docs and .env.example, had to hunt down this issue to find documentation on how to fix linking
> 
> OIDC_EXISTING_USER_LINK_MODE now supports: disabled — default, safest; do not auto-link existing accounts by email. verified_email — link existing accounts by email only when the IdP provides email_verified=true. email — link existing accounts by matching email even without email_verified. This keeps the hardened default, but allows relaxed private/self-hosted deployments like yours to migrate existing Securo users without recreating their accounts/data.

Good catch, that one never made it into the docs. Adding OIDC_EXISTING_USER_LINK_MODE to .env.example, the README and the docs site, thanks for flagging it 🙌

PR #287, opened 2026-06-07
feat(translation): updated PL strings for #264

dmaniaa:
## What

Added missing strings from #264 to Polish translation

## Why

To update the translation and avoid missing strings

## How to Test

Verify that strings added with #264 are displayed in Polish instead of English.

## Checklist
(only strings changed, n/a)


tassionoronha:
Thanks for keeping that updated @dmaniaa !

PR #288, opened 2026-06-07
fix: Decouple nginx.conf in frontend from container names

Cilenco:
## What

Currently the container name of the backend server is hardcoded into the nginx.conf file of the frontend container.
This PR fixes the problem by using the default envsubst feature of the nginx container and the already used env variables.

## Why

Renaming the backend container name from the `docker-compose.prod.yml` file breaks the application as the frontend container is trying to resolve DNS for `http://backend:8000`. This creates a tight coupling between the containers which should be avoided.

## How to Test

1. On main branch: try to start up `docker-compose.prod.yml` after renaming the backend container and all its dependencies
2. Try the same on this branch: All containers start well and the application is running correctly.

## Checklist

### DIFF
--- a/frontend/Dockerfile
+++ b/frontend/Dockerfile
@@ -15,6 +15,6 @@ RUN npm run build
 FROM nginx:alpine
 
 COPY --from=build /app/dist /usr/share/nginx/html
-COPY nginx.conf /etc/nginx/conf.d/default.conf
+COPY default.conf.template /etc/nginx/templates/default.conf.template
 
 EXPOSE 80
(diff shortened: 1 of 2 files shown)


Cilenco:
The same could be done for the port binding (using the env variable `FRONTEND_PORT`) to allow the container running as non root but this probably should be done in another PR.

tassionoronha:
Hey @Cilenco , welcome here and thanks for this one

Good imprv 🙌 
Decoupling from the container name is the right call, and reusing the existing BACKEND_URL with nginx's envsubst is clean. LGTM!

PR #291, opened 2026-06-07
feat: add vitest configuration and add i18n validation tests

geeosp:
# Add i18n test suite for locale file validation

Hey everyone, I though about adding an automated test to ensure correct translation though all the languages currently supported. 
Used Claude to make a test suite for us. It already got some missing translations.
Added a new dependency  in the project (vitest) creating a test suite for frontend.
It currently ensures that every key has a translation between all languages, that the keys are unique and dynamic texts continue dynamic in every translation.
We could hook this up with the current github pr workflow,  then we can automte the translation checks we are doing by hand.

---

## How to test

From the `securo/frontend/` directory:

```bash
npm test
```

To run in watch mode during development (re-runs on every file save):

```bash
npx vitest
```

---

### New files
- **`src/locales/i18n.test.ts`** — the test suite
- **`vitest.config.ts`** — Vitest configuration (node environment, `@/` alias)

### Modified files
- **`package.json`** — added `"test": "vitest run"` script
- **`tsconfig.node.json`** — extended `include` to cover test files so they get Node.js types
- **`devDependencies`** — added `vitest`

---

## Why it matters

The project has an increasing number of locale files (`en`, `es`, `it`, `pl`, `pt-BR`) that are maintained manually. Until now, there was no automated check to verify they stayed in sync. Any developer adding a new UI string had no safety net — a forgotten translation would silently fall back to English (or render the raw key like `transactions.monthNext`) in production, with no warning anywhere in the build pipeline.

The test suite runs via `npm test` and enforces four properties:

---

**1. No duplicate keys**

Ensures that the same key is not defined twice in any language

**2. All languages contain all keys from English**

English is the source of truth and the i18next fallback language. If a locale file is missing a key, i18next falls back to English silently — there is no runtime error, no console warning, just an untranslated string in production. This test catches that before it ships.

**3. No extra keys not present in English**

The reverse direction: if a key exists in a locale but was removed or renamed in English, it becomes a dead translation.  It is aware of languages withplural forms (`_one`, `_few`, `_many`, `_other`) such as Polish, so it does not false-positive on those.

**4. Placeholder variables match English**

Translations can contain interpolation variables like `{{count}}` or `{{name}}`. If a translator omits a variable — writing `"Hello!"` instead of `"Hello, {{name}}!"` — the app renders the string without the dynamic value, which is a silent data loss bug. This test extracts every `{{variable}}` from each English string and verifies the same variables appear in every translation of that string. It is also plural-form aware: for Polish it checks each plural variant (`_one`, `_few`, etc.) individually against the English base string's placeholders.

---

## What it does not cover

Translation quality and accuracy are intentionally out of scope — those require human review. A translator could technically copy the English text into every field and these tests would pass. The tests focus entirely on structural correctness: completeness, consistency, and variable integrity.

---

## Known failures introduced by this PR

Running `npm test` currently reports 3 failing tests. These are pre-existing gaps in the locale files that the new test suite surfaced — they are not regressions introduced by this PR.

**Spanish (`es.json`) — 2 missing keys**
- `transactions.monthPrevious`
- `transactions.monthNext`

**Italian (`it.json`) — 1 missing key**
- `reports.change`

**Polish (`pl.json`) — 23 missing keys**
Mostly settings-related strings added recently: `transactions.effectiveBillDate`, `transactions.effectiveBillDateHint`, `transactions.effectiveBillDatePlaceholder`, and the full `admin.settings.numberFormat*` and `admin.settings.dateFormat*` groups.

These should be addressed in a follow-up translation pass. The tests are intentionally left failing so they remain visible and don't get silently buried.


MarlonPassos-git:
Perhaps it's important to add another related PR that corrects these problems, so that when everything is merged into the main method, it doesn't start with "problems".

Another thing I noticed: if the idea is to automate things, this PR should already be included with the CI configurations to run the analyses automatically.

I saw that you created another `vitest.config.ts` file; wouldn't it be better to integrate this directly into the `vitest.config.ts` file, already with the `test` property?

I think the initiative is great because, with Vitest configured in the project, we would be "free" to add more unit test coverage to the front end.

tassionoronha:
Thanks @geeosp, great initiative, and it's already proving its point: since you opened this, pl.json got fixed but it.json drifted to 28 missing keys (collections + OIDC landed untranslated). 

Before merging tho:

1. Fix the missing es/it translations first (same PR or stacked), I don't want to merge with red tests on main.
2. Add npm test to the frontend job in ci.yml in this PR, otherwise the suite rots.
3. Minor: add "exclude": ["**/*.test.ts", "**/*.test.tsx"] to tsconfig.app.json so tests aren't type-checked in both TS projects.

@MarlonPassos-git on merging the configs: I'd actually keep vitest.config.ts separate. Our vite.config.ts is async, runs version resolution and loads React/Tailwind plugins, none of which node tests need. The 3-line alias duplication is the cheaper tradeoff for now I guess.

Agreed this opens the door to frontend unit tests in general and I defo like it.

geeosp:
Nice, the tests were added to ci and started passing after add the missing translations.

tassionoronha:
Thanks @geeosp. 
Attacking this one the next days 🙌🏼

PR #301, opened 2026-06-09
Feat: Networth Composition Widget Redesign

geeosp:
## What
Hey Everyone, this is another quality of life addition to the Net Worth page. This is inspired by #219 discussion. I present you a nicer Composition widget for the Networth page. 

Here are the key modifications:
1. On the composition pie chart, I killed the separation between the` Summary` and `Detailed` states. They are now clearly represented in the same pie, The outer ring always shows the detailed view, while the smaller width inner ring  holds the old Summary view.
2. Inner ring and Outer ring are aligned in a way that is easy to identify what is an asset, what is an account, what is a liability.
3. The toggle now  changes between `NetWorth` , `Asset & Accounts` and `Liabilities`. The intent is to make clear how each kind of asset contributes for your networth, showing what is that you own and, in opposite,  what you owe. When The Networth toggle is selected, the graph shows the same information we had originally. 
5. Maintained the colors defined by Assets page the behavior is the same as before. 
6. Maintained the detailed view as individual assets instead of wallets. That was a running discussion from #219. I decided to maintain the breakdown we had before, so the Assets session of the outer pie shows your individual assets.
7. The pie used to show your 6 most significant assets,accounts and liabilities. I changed this behaviour to show everything that hold more than 2% of the total value. I choosed 2% empirically because you can still easily distinct a 2% slice from the rest. Anything bellow it gets grouped in a groups session.
8. By hovering a groups session, the tooltip shows what is being accounted by it. 
9. Assets, Accounts, Liabilities. Each of them have their own groups of "remaining assets"  if needed

Below is the final result 
## How to Test
Go to Reports / Net Worth page. Play with the Composition widget

https://github.com/user-attachments/assets/f5b2ab57-f690-4cc7-aa8c-36a99e166fed

## Checklist


tassionoronha:
Hey @geeosp. 
Nice work you've been doing here, thanks for that.

I found some things to improve tho:

On load (and after switching tabs) the donut center shows the gross total, Accounts + Assets + Liabilities — instead of true net worth, because liabilities get added rather than subtracted. The default compositionView is 'summary', which isn't a valid net_worth toggle, so no button is highlighted and the center falls through to the gross-sum branch. It only corrects after clicking "Net Worth".

Also, on the UI side I have:

image

Liabilities which dont change the colors and make a chart with a poor visual. 

image
 
 This first composition graph still feels like it is breaking the page layout. Imagine a user with even more assets. The legend would keep growing and push the page height further down at the bottom.

 We should think about a better solution here. We could aggregate the items, or set a limit on how many lines we show, and add a "Show more" button so the user can expand the rest only if they want to.

Also, the colors are still not different between the items, so it is hard to tell them apart.

I think this can be a good improv anyway, we just need to refine it better. A fullscreen option opening this in a bigger modal also could be a proper solution for people which would have a considerable number of elements for the chart.

geeosp:
@tassionoronha  , 
Could you please confirm if it loads correctly of the first time? I set the toggle to select Networth by default, so the pie shows the correct value at first try. Also added a small placeholder skeleton for the pie chart while its loading. 
The legend items now overflow to a "Show more" button, which open a temporary panel with the missing items.  So I expect it to not grow indefinitely

https://github.com/user-attachments/assets/b2c8b706-81dd-42ac-a553-bbd46c81b0c1

geeosp:
The only thing I didn't worked is on yet is on how we choose the colors, I am still uncertain about what could be a good solution.

geeosp:
Nevermind, I think I found a good solution:)
Started choosing a random color by the asset/account/liability id hash, this way it stay consistent across runs. Also, tryied to make the chosen colors tied to the specific item type (account/liability/asset)

Assets stay between orange>green, 
Accounts between blue and purple
Liabilities between red and pink

This is a possible result: 

image

geeosp:
Ok, I am happy with this PR

tassionoronha:
Hey @geeosp thanks for that. 
Travelling these days, so I'll check it carefully on the weekend.
Cheers! :)

tassionoronha:
Such a nice work here, @geeosp. 
Just adapted some style points. 
Thanks for the amazing work here!

PR #302, opened 2026-06-09
Feat: Income & Expanse Composition widget review based on #301

geeosp:
## What

Hello, me again, I started looking only to the net worth page. But after working on the networth page, I noticed that the composition widget from `Income & Expanses` page behaves exactly like the Composition widget from `Networth` page. So I adapted it with following the same rules, and this is the result. 
Please note that this PR have all the modifications from #301, its changes will probably get simpler once its gets merged. 

This is the final result: 

https://github.com/user-attachments/assets/dbfdbc9e-4331-4524-891c-3147a2530868

## Checklist


tassionoronha:
Nice cleanup, and it fixes the default value bug from #301. 

Two things here: on the Net Worth tab no toggle shows as active on load, because the default is "net" but the button key is now "netWorth". Also the Income view shows two gray "Uncategorized" items in the legend that look the same.

The points that I made about the desing on #301 are also true here. 
Also, et me know if you want to follow from here or the other one and close one of those so we can focus in one workspace :) 

Again, nice work here @geeosp !

geeosp:
Hi @tassionoronha, lets close this one and continue to work on the other then, I will put both in the branch from #301

PR #305, opened 2026-06-10
feat(accounts): allow overriding the type of bank-connected accounts (#271)

### LINKED ISSUE #271: [Feature] Ability to manually set the type of connected accounts
## Problem

When using Enable Banking's integration with a savings account from mBank, the type of account is set to "Checking". I am not sure if getting account types is implemented or not, but overriding the type of connected accounts would always be a welcome feature in case something didn't work out.

## Proposed Solution

Add the option to change the account type of connected accounts.

## Alternatives Considered

Make sure the account type is always correctly pulled from external providers. Seems to be substantially more difficult than the option to change, but would also be welcome further down the line. Currently, the option to change as a workaround should be enough.

## Additional Context

Possible mockup:

Image

tassionoronha:
Closes #271

## What

Let users override the **account type** of a bank-connected account from the account edit dialog. The Type selector — previously shown only for manual accounts — now appears for connected accounts too, and the API accepts `type` edits on them.

## Why

Providers sometimes misreport the account type. As reported in the issue, Enable Banking labels an mBank **savings** account as **checking**. Type isn't just cosmetic — it drives how accounts are grouped and, for credit cards, whether the balance is treated as a liability in net worth. Users had no way to correct it.

The sync pipeline only writes `type` when it first creates an account and never overwrites it on later syncs, so a user override persists across syncs **without** needing a separate override column or migration.

## Changes

- **`account_service.update_account`** — add `type` to the editable-field whitelist for connected accounts. When the override moves an account off `credit_card`, stale card metadata (limit, cycle days, brand/level) is cleared so it isn't left in a half-credit-card state.
- **`accounts.tsx`** — show the Type selector for connected accounts with a hint that the override survives syncs; always send `type` on save. The type option list is extracted to a shared `ACCOUNT_TYPE_OPTIONS` constant (used by both the manual and connected selectors).
- **i18n** — new `accounts.typeOverrideHint` string (en, pt-BR, es, it, pl).

## How to Test

1. Connect a bank and sync so you have a connected account.
2. Open the account's edit dialog — the **Type** dropdown now appears, with a hint that the override won't be reverted by syncs.
3. Change the type (e.g. Checking → Savings) and save — the account updates immediately.
4. Trigger a re-sync — the override is preserved.
5. Override a connected card to a non-card type — credit-card-only fields are dropped.

Verified in Chrome against a connected NuBank account (Checking → Savings → back).

## Checklist

- [x] Backend tests pass (`pytest` — new `test_account_service` cases for override persistence + card-metadata clearing, plus existing account suites)
- [x] Backend lints clean (`ruff`)
- [x] Translations updated (en, pt-BR, es, it, pl)

### DIFF
--- a/backend/app/services/account_service.py
+++ b/backend/app/services/account_service.py
@@ -267,12 +267,17 @@ async def update_account(
         k in update_data for k in ("statement_close_day", "payment_due_day")
     )
 
-    # Bank-connected accounts are managed by the sync pipeline. Only credit card
-    # metadata (limit + cycle days) can be user-edited, since providers often don't
-    # expose those — users fill them in to unlock cycle-aware filtering.
+    # Bank-connected accounts are managed by the sync pipeline. Beyond display
+    # name and credit card metadata (limit + cycle days, which providers often
+    # don't expose), users may also override the account `type` — providers
+    # sometimes misreport it (e.g. Enable Banking labels an mBank savings
+    # account as "checking"; issue #271). Sync only writes `type` on initial
+    # account creation and never overwrites it afterwards, so the override
+    # survives subsequent syncs without a separate field.
     if account.connection_id is not None:
         editable_fields = {
             "display_name",
+            "type",
             "credit_limit",
             "statement_close_day",
             "payment_due_day",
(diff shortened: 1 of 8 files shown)


dmaniaa:
Hopped onto the branch with a sandbox EB, verified all the types can be set, and synchronization does not restore the type. Looks good to me.
Verified on Firefox 151.0.4 macOS 26.5.1

tassionoronha:
Thanks for checking and confirming @dmaniaa !!

PR #312, opened 2026-06-10
fix(prod): set container TZ so bank sync uses local calendar day

Starmania:
## What

Add TZ env in docker-compose file

## Why

This wasn't explicit that this was needed

## How to Test

1. Set TZ as some timezone
2. `docker compose -f docker-compose.prod.yml exec -T celery-worker date` should show a different value than UTC

## Checklist

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -36,6 +36,9 @@ ENABLE_BANKING_OAUTH_REDIRECT_URI=http://localhost:${FRONTEND_PORT:-3000}/oauth/
 # OIDC_ADMIN_ROLES=securo-admins
 # OIDC_WORKSPACE_ROLE_MAP={"securo-owners":"owner","securo-editors":"editor","securo-viewers":"viewer"}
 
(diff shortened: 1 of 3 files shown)


tassionoronha:
Thanks, good catch and in fact the backend uses date.today() a lot, so TZ does matter.

Two small changes and I'll merge:
- Default to UTC instead of Europe/London (TZ: ${TZ:-UTC}). UTC is what containers use today, so nothing changes for existing users and with people using securo in different places, no single timezone fits as a default.
- Add a TZ= line to .env.example so people can find the option.

With that I think we would be good to go.

Starmania:
@tassionoronha Do you need something more ?

tassionoronha:
I think this PR is complementary with #339. 
So I will need a bit more time in order to conciliate these implementations, but on your side looks good.
Thanks for taking it in consideration!

tassionoronha:
Thanks for this. It's quite far behind now, and #909 is about to make the timezone an app setting rather than a container env, which covers the same need. I'll close this once #909 lands; if you think the env default is still worth having alongside it, say so and we can fold it in there.

PR #313, opened 2026-06-10
fix(prod): add restart policy to db and redis

Starmania:
## What

Added restart policy to db and redis in the prod docker-compose

## Why

Only backend-base carried restart: unless-stopped; db and redis had none. After a host reboot the data services stayed down while backend/celery restart-looped against unresolvable hostnames, taking the whole stack offline. Add restart: unless-stopped to both so they recover automatically.

## How to Test

Reup your stack
Restart the host
You should be able to login

## Checklist

### DIFF
--- a/docker-compose.prod.yml
+++ b/docker-compose.prod.yml
@@ -66,6 +66,7 @@ x-backend-base: &backend-base
 services:
   redis:
     image: docker.io/library/redis:7-alpine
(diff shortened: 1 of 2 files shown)


tassionoronha:
Hey @Starmania, welcome here. 
The changes LGTM, thanks for that! :)

PR #330, opened 2026-06-15
feat: add passkey authentication

Sgubela:
## What

Add passkey/WebAuthn support for Securo authentication.

- Add backend passkey model, migration, schemas, config, and `/api/auth/passkeys/*` endpoints.
- Support passkey registration, management, passkey-only sign-in, and security-key verification during password login.
- Require second-factor verification when a user has passkeys enabled, with TOTP fallback when both methods exist.
- Add frontend WebAuthn helpers, login flow, passkey management dialog, account-menu entry, types, and translations.
- Prefer passkey verification by default when both passkey and authenticator code methods are available.

## Why

Passkeys should provide both convenience and extra account security. A registered security key/passkey can now be used for passwordless sign-in, and it also counts as a second-factor method so password login alone is not enough when a passkey is enrolled.

## How to Test

1. Backend targeted auth tests:
   `cd backend && uv run pytest tests/test_passkeys.py tests/test_two_factor.py tests/test_oidc_auth.py`
2. Backend lint:
   `cd backend && uv run ruff check app tests`
3. Frontend lint:
   `cd frontend && npm run lint`
4. Frontend build:
   `cd frontend && npm run build`
5. Manual browser verification:
   - Start the stack and open `http://localhost:3000/login`.
   - Log in with a test user.
   - Open the account menu and manage passkeys.
   - Add a passkey/security key.
   - Log out.
   - Verify passkey-only sign-in works.
   - Verify password sign-in now requires second-factor verification.
   - If both passkey and TOTP are enabled, verify passkey is the default challenge and authenticator code remains selectable.

Notes:
- `npm run lint` exits successfully but the repo currently emits existing warnings in unrelated files.
- A local full backend coverage run was started with `uv run pytest --cov=app --cov-report=term-missing --cov-fail-under=60`, but it did not complete within the local polling window, so CI should be treated as the source of truth for full-suite coverage.

## Checklist

- [x] Backend targeted tests pass (`pytest`)
- [x] Frontend lints clean (`npm run lint` exits 0; existing warnings remain)


tassionoronha:
Such an amazing PR and work here @Sgubela !
It took me some time to evaluate that and there's some aspects like recovery flow, user management that could be also present, but nothing block us to increment that in future versions. 

For now, the feature is good enough to be merged. 
Thanks for pushing this one!

PR #278, opened 2026-06-06
feature(translation): Added Italian language

Ayoubl96:
## What
Adds Italian (`it`) as a selectable UI language, following the same pattern as the existing es/pl locales.

## Why
Makes Securo usable for Italian speakers. Opened as a first contribution.

## How to Test
1. `docker compose up --build`, open http://localhost:3000
2. Open the language menu → **Italiano** (or visit `/?lng=it`)
3. UI renders in Italian; the choice persists across reloads
4. A new account/workspace seeds the 6 default category groups in Italian

## Checklist
- [x] Backend tests pass (`pytest`) — no behavior change
- [x] Frontend lints clean (`npm run lint`)
- [x] Frontend builds (`npm run build`)
- [x] Translations updated (added full `it.json`, key parity with `en.json`)

Translated with Claude. Manually reviewed code changes and strings. Executed basic manual E2E UI test.

PR #263, opened 2026-06-04
feature(translation): Added Polish language

dmaniaa:
## What

Translated and added localization with Claude. Manually reviewed code changes and strings. As a native Polish speaker, I corrected strings for more natural sound. Executed basic manual E2E UI test as described below.

## Why

To make the app more accessible to people in Poland.

## How to Test

1. Set up a new Securo instance.
2. Verify the setup guide shows PL as an available choice and selecting it switches the language to Polish.
3. Go through initial setup and guide.
4. Verify that Polish language is available and can be set through the user dropdown.
5. Refresh the site to verify the translation sticks.
6. Add 1, 2, 5, 22 transactions manually to verify plural number forms work correctly (expecting 1 transakcja, 2-4 transakcje, 5+ transakcji, x2-x4 transakcje.
7. Add more example data by, for instance, linking accounts and manually filling out each section.
8. Verify all strings are correctly displayed in each section of the UI based on the example data.

## Checklist

- [x] Backend tests pass (`pytest`)
- [x] Frontend lints clean (`npm run lint`)
- [x] Frontend builds (`npm run build`)
- [x] Translations updated (if user-facing strings changed)

Pytest and lints displayed warnings, all verified to be unrelated to my changes.

PR #246, opened 2026-06-01
feat: add spanish language

AlexTorresDev:
## What

Full support for the Spanish language *es.json* has been implemented throughout the Secure platform. Changes include:

- Integral Location: Translation of all the main modules: Dashboard, Transactions, Accounts, Budgets, Objectives, Assets and Reports.
- Language selector: Agged language to all selectors in both configuration and the rest of the UI

## Why

This update is key to the expansion of the project to Spanish-speaking users. By translating the interface, users’ interaction with financial data is ensured to be natural and accessible.

## How to Test

To verify that the translation is correct and well applied:

1. Launch the application and navigate to Workspace Settings > Language and select "English".
2. Browse the Transactions and Dashboard sections to ensure that the tags and tooltips are displayed correctly.

## Checklist

- [x] Backend tests pass (`pytest`)
- [x] Frontend lints clean (`npm run lint`)
- [x] Frontend builds (`npm run build`)
- [x] Translations updated (if user-facing strings changed)

PR #334, opened 2026-06-17
Add Russian localization

sonsay:
## Summary
- add Russian translation resources and register ru in the i18n language resolver
- expose Russian in setup, user, admin, and workspace language selectors
- add missing translation keys used by the AI chat/MCP UI and transaction tag controls
- use a shared date-fns locale resolver so calendars/month labels support all configured languages
- allow the transaction dialog footer actions to wrap instead of overflowing in longer locales

## Validation
- npm run build completed in the production Docker frontend build
- checked en/ru locale key parity: 1448 keys each, no missing or extra keys
- checked literal t('...') usages against ru.json: no missing keys
- git diff --check passes


tassionoronha:
Hey @sonsay, welcome here and thanks for adding the russian translation :)

One gap: the previous language PRs (#246, #263, #278) also added their translations to DEFAULT_GROUPS_I18N in
backend/app/services/category_group_service.py. Without "ru" there, Russian users would get English category group names (Housing, Food & Dining, …) on setup.

Could you add ru to those groups? That's the only thing blocking, rest looks good 👍

sonsay:
Thanks for catching this! I added `ru` entries to `DEFAULT_GROUPS_I18N` and covered the Russian default group names with a focused backend test.

Validated locally with:

```bash
uv run --extra dev pytest tests/test_category_group_service.py
```

tassionoronha:
Thank you @sonsay, LGTM now!

PR #335, opened 2026-06-18
Feat/networth evolution composition link

geeosp:
## What

Links the net worth **evolution chart** to the **composition widget**: clicking any bar on the Evolution chart will update the pie chart from Composition widget to show a snapshot at that date time.
This is the last modification proposed for #219

https://github.com/user-attachments/assets/eab15de9-608e-4314-b7c5-4d3a32a47947

Key Design decisions:

- **Evolution → Composition binding**: selecting a date on the chart drives the composition widget; a date pill and reset button make the active snapshot obvious.
- **Deterministic color mapping**: slice colors are assigned from its rank in todays date, so the current rule still apply. Also, the same chosen colors will be reused to the same specific assets in any historic snapshot in the pie composition chart. This have the effect that the colors can change its ordering between historic datepoints if the relative proportion between assets change over time. But that is ok, the key design decision here is that by making an asset color deterministic, we can track its related asset though history by tracking the color in the chart. 
- **Outer ring 3% threshold**: items whose value is < 3% of total are collapsed into an "Other" slice per group; removed the old fixed rank cap.
- **Other group by asset type** We eliminate the "Others" group, substituting it for one group session for each kind of asset. This enable us to sync and align the inner ring with the outer ring. Within each session, we get assets/accounts/liabilities ordered by total value, but they do not get scattered across the pie anymore. There are specific sessions for accounts, for assets and liabilities. 
- **Pie chart tooltip & "+ more" popover styling**: aligned styles between the two for visual consistency.

Code Changes
- **Seed data**: enhanced `seed_perf.py` with asset groups and additional credit cards for more realistic composition testing.
- **Per-point composition in the API**: each `ReportDataPoint` now carries a `composition` list with per-account and per-asset values as of that date, computed server-side via the refactored `_net_worth_at` + `_asset_value_at` helpers.

## Why

The composition widget previously only showed the current portfolio breakdown, this pr makes possible to investigate past allocations. The evolution chart had no interactive drill-down. This wires the two together so users can click any historical point and immediately see what drove net worth at that moment.

## How to Test

1. Navigate to **Reports → Net Worth**.
2. Click any bar on the evolution bar chart — the composition donut/ring should update to show the breakdown as of that date, with a date pill appearing in the widget header.
3. Verify the "Reset to current" button clears the snapshot and restores the live breakdown.
4. Switch between several historical points and confirm colors remain stable (the same account always has the same color).


tassionoronha:
image

Hey @geeosp , thank you, really nice feature 🙌 
Feels like a strong second step on what you've been doing, and I like it a lot.

Two small UX things before merging:

1. When I click a bar, there's an extra shaded square that lights up on the chart feels noisy, I'd drop it.
2. Clicking outside the box still keeps the snapshot active. I'd rather it stay put and only reset via the × on the composition pill.

Honestly don't have the perfect answer, just challenging you a little, feels like we're almost there but not quite clean/intuitive yet. Mind taking how to make it feels better?

Thanks again 🚀

geeosp:
Ok, I've got some ideas

geeosp:
I Removed the highlight on the selected bar. I wasn't able to reproduce your selection square, but removed it in favor of the next point.
- Dimmed non-selected bars to 40% opacity when a snapshot date is active, giving clear focus to the selected period. I found this is a better aproach, it becames more readable the more dense the dates get.
- Moved the snapshot date pill from the Composition card to inline with the Evolution chart title, where it's closer to the interaction that triggered it.
- Adds a translated "as of {date}" subtitle below the Composition title when a snapshot is active,  so the "which date i am at" info is in both widgets. 

https://github.com/user-attachments/assets/9ecfbaf9-fcb9-4931-9f54-165eb585a96b

geeosp:
@tassionoronha, I dont know if I understood correctly your second feedback, but I would advise against remove the deselect action from clicking the bar. It makes sense to deselect with the same action you select, dont you agree? 
Lets say you select a bar by accident, intuitively you would try to click in the same place I believe. 
Give it a run, let me know if you still find useful to remove it.

tassionoronha:
@geeosp, fair enough. 

Your interpretation is safer than mine for sure. Let's go for it. 
I'm pushing some changes to remove the focus around the selected element, which was not visible on your demos btw(maybe a macos chrome thing), to make it cleaner.

Apart of that, such an amazing work!

PR #336, opened 2026-06-18
Perf/reports page performance enhancement

geeosp:
# perf(reports): eliminate O(N×M) queries from net worth report generation

This PR is built on top of https://github.com/securo-finance/securo/pull/335.
While working on #219 I noticed a lot of opportunities for performance improvement on the networth page.
This PR should not bring any visual updates, its all about performance. 
It changes the optimizes underlying querys for data in the backend for the networth reports page, improvivin the eficiency when accessing the databse.
Its is very noticiable when looking at the Networth page for the test user, when using the 2 year daily view. 
Once #335 is in, this pr should bring only 4 focused commits based on performance.
All previous tests are passing.
 
## Problem

The net worth report called `_net_worth_at` once per trend point (N points), and inside each call it:

- re-fetched the full account list from the DB
- re-fetched the full asset list from the DB
- issued one `SELECT` per account to compute `balance_at(cutoff)` via `_account_balance_at`
- issued one `SELECT` per asset to fetch the latest `AssetValue` row at or before `cutoff`

For a **2-year daily report** with 30 accounts and 20 assets this produced roughly:

| Query type | Count |
|---|---|
| Account list | 730 |
| Asset list | 730 |
| `_account_balance_at` | 30 × 730 = 21,900 |
| `AssetValue` latest row | 20 × 730 = 14,600 |
| **Total** | **~38,000** |

### 1 — Fetch asset list once 
Raise the `Asset` SELECT above the trend loop in `get_net_worth_report` and thread the result down as `prefetched_assets`. `_net_worth_at` falls back to its own query when called standalone.

### 2 — Bulk-load `AssetValue` rows 
Add `_bulk_load_asset_values`: one `SELECT` fetches every `AssetValue` row for all assets up to today, keyed by `asset_id`. Value resolution at any cutoff becomes an in-memory reverse scan (`O(entries)`) instead of a per-asset-per-point DB round-trip.

### 3 — Fetch account list once 
Same pattern as assets — raise `_get_open_accounts` above the loop and pass it as `prefetched_accounts`.

### 4 — Bulk-load account balance data 
Add `_bulk_load_account_balance_data` + `_balance_from_preloaded`: two total queries replace all `_account_balance_at` calls.

- **Bank-connected accounts**: base balance is already on `account.balance` (no query); `balance_at(cutoff) = base − sum(deltas where date > cutoff)`.
- **Manual accounts**: one `GROUP BY` query sums signed transactions up to `start`; `balance_at(cutoff) = base + sum(deltas where date ≤ cutoff)`.

`_compute_signed_amount` replicates the SQL signed-balance expression in Python so both maps stay consistent with what the DB would have returned.

## After

| Query type | Count |
|---|---|
| Account list | 1 |
| Asset list | 1 |
| Account balance data | 2 |
| `AssetValue` bulk load | 1 |
| **Total** | **~5** |

The improvement scales with both axis: longer date ranges and more accounts/assets no longer increase query count.

## Compatibility

All new parameters on `_net_worth_at` default to `None` — existing callers (including standalone tests of `_net_worth_at` and `_asset_value_at`) continue to work without changes.

Net worth trend generation now reuses bulk-loaded account, transaction, asset, and asset value data. This removes repeated database queries while preserving existing report results.

- Two-year daily reports load much faster.
- Net worth values and trends remain unchanged.
- Balance calculations avoid float precision drift.

Risk: money math.


geeosp:
I will let this pr as a draft until we merge https://github.com/securo-finance/securo/pull/335

tassionoronha:
Hey @geeosp,  

Thank for this one :)

Tested locally and the net worth report works great, even the 2Y daily view loads fast with no errors. Logic looks correct too. One small thing: summing floats per transaction can drift a cent vs the old Decimal SUM, so I'd accumulate Decimal and cast once.

geeosp:
Sure, I will look into it!

geeosp:
Done!

tassionoronha:
Nice work @geeosp! I ran the branch and compared the new bulk path against the old one across many dates: the results match exactly, so the logic looks solid.

Could you add tests please? :)

tassionoronha:
Merged main into the branch so CI runs on current code. Backend Tests still fails, but not on the tests: the type checker flags two things in report_service.py (an argument to _net_worth_at and a .currency access on a plain object). Those need a look on your side before this can move.

PR #338, opened 2026-06-18
Add categorization rules import/export

### LINKED ISSUE #307: [Feature] Import/export auto-categorization rules for sharing between instances
## Problem

I've been self-hosting Securo for a while and have built up a good set of auto-categorization rules. A friend now wants to host their own instance and is interested in the rules I already made. Right now the only way for them to get those rules is to recreate each one by hand, which is slow and easy to get wrong.

## Proposed Solution

Add the ability to export auto-categorization rules to a file (like JSON) and import them into another instance.

Ideally:
- An Export button to download all (or selected) rules as a file.
- An Import button that reads that file and creates the rules in the target instance.
- On import, match categories by name and let the user resolve any that don't exist, with an option to skip, overwrite, or create duplicates.

## Alternatives Considered

- Recreating the rules by hand. Works, but slow and error-prone.
- Copying database rows directly. Too risky, and not practical for a fresh instance.
- Sharing rules in a doc to retype. Still manual re-entry.

## Additional Context

The use case is two separate self-hosted instances where one person tuned the rules and wants to share that work. A portable format would also help with backups and migrating rules to a new instance.

RomuloGatto:
## Summary
- Add portable JSON export for categorization rules, serializing category actions by category name instead of workspace-specific UUIDs.
- Add import endpoint that maps categories by name, skips rules whose categories are missing, and requires explicit overwrite confirmation before replacing current rules.
- Add Rules page Export/Import controls plus an overwrite confirmation dialog before importing.

Closes #307

### Export / Import controls
Rules export/import toolbar

### Import overwrite confirmation
Rules import overwrite confirmation dialog

## Verification
- `cd backend && . .venv/bin/activate && pytest tests/test_new_rules_api.py -q`
- `cd backend && . .venv/bin/activate && ruff check app/api/rules.py app/services/rule_service.py app/schemas/rule.py tests/test_new_rules_api.py`
- `cd frontend && npm run lint -- --quiet`


tassionoronha:
Thanks @RomuloGatto. Nice work! One thing I hit, if I import a file whose category names don't match my workspace every rule gets skipped but with overwrite on it still deletes all my existing rules first and imports nothing. So I end up empty. 

Could you skip the delete when nothing will replace them?

RomuloGatto:
Thanks for catching that. I updated the import behavior so overwrite only deletes existing rules after the import payload has at least one rule that can actually be created. If every incoming rule is skipped (for example because none of the category names match the workspace), the existing rules are preserved and the response reports `imported: 0`, `skipped: n`, `overwritten: 0`.

Added a regression test for the all-skipped overwrite case.

Verification:
- `pytest tests/test_new_rules_api.py -q`
- `ruff check app/services/rule_service.py tests/test_new_rules_api.py`

tassionoronha:
Thanks for fixing that @RomuloGatto . 
LGTM!

PR #347, opened 2026-06-23
Assign transactions to existing categorization rules

DannyFeliz:
## What

- Adds an `Add to existing rule` action to the transaction detail `Create Rule` split button.
- Adds a modal to select an active category rule and append the current transaction description as a `contains` or `starts_with` match.
- Updates `PATCH /api/rules/{id}` to apply the updated rule to existing transactions and return `applied_count`.
- Refreshes transaction/financial views when updated rules affect historical transactions.
- Adds localized strings and backend regression coverage for rule update backfills.

## Why

Users can teach an existing categorization rule directly from the transaction they are reviewing, without navigating to the rules page and manually editing the rule. Matching uncategorized history is also updated immediately.

## How to Test

Open a transaction detail modal, use the `Create Rule` dropdown, choose `Add to existing rule`, submit a description match, and verify matching transactions refresh with the rule category.

## Checklist

## Demo

https://github.com/user-attachments/assets/af2c7797-e466-42b3-9450-58d993f0d5ca


tassionoronha:
Hey @DannyFeliz :)

Love PR #347.  Adding a transaction to an existing rule from the detail view is a great shortcut 🙂 
Before merging, a few things to fix: 

1) PATCH now re-applies the rule on every edit (even a rename), which can overwrite manual payee/ignore changes and scans all transactions each time — worth gating to real condition changes. 

2) The duplicate check misses accented text. 

3) The match seeds from the unsaved description, so it can silently match nothing. 

Mind a look?

DannyFeliz:
@tassionoronha Good catch.

I fixed all the issues you mentioned. Please take another look and let me know if there's something else I need to address.

tassionoronha:
Thanks for the fixes @DannyFeliz. 
Just fixed a minor point and now LGTM! :)

PR #350, opened 2026-06-24
fix(oidc): verify id_token at_hash so Google sign-in works

ADolkun:
## Problem

Signing in with Google via OIDC fails with `400 {"detail":"Invalid OIDC id_token"}`. Any provider whose ID token carries an `at_hash` claim is affected — and Google always includes it — so "Sign in with Google" can never succeed.

## Root cause

`_decode_id_token()` in `backend/app/api/oidc_auth.py` calls `jwt.decode(...)` (python-jose) **without** an `access_token`. python-jose verifies `at_hash` by default (`verify_at_hash=True`), and `_validate_at_hash` raises `JWTClaimsError("No access_token provided to compare against at_hash claim.")` when the claim is present but no `access_token` was supplied. That `JWTError` is caught and re-raised as `400 Invalid OIDC id_token`, so login fails for every Google account.

The existing OIDC tests monkeypatch `_decode_id_token` wholesale, so they never exercised real JWT decoding and didn't catch this.

## Fix

Forward the `access_token` — already returned by the token endpoint and in scope at the call site (the very next line passes it to `_fetch_userinfo`) — into `_decode_id_token` → `jwt.decode(..., access_token=access_token)`. This **validates** `at_hash` (binding the ID token to the access token, as the OIDC spec intends) instead of failing. The new parameter defaults to `""`, so providers that omit `at_hash` (e.g. the one in the existing tests) are unaffected.

An alternative would be `options={"verify_at_hash": False}`, but that *disables* a check this code can satisfy for free; verifying is strictly more correct and costs the same.

## Test

Adds `test_decode_id_token_accepts_google_style_at_hash`, which mints a real RS256 ID token carrying an `at_hash` claim (via python-jose, same lib as prod), stubs the JWKS fetch, and asserts the **real** `_decode_id_token` validates it. Fails before the fix (`JWTClaimsError` → `400`), passes after. Existing stubs that monkeypatch `_decode_id_token` were updated to accept the new optional arg.

- `pytest tests/test_oidc_auth.py` → **17 passed**
- `ruff check` → clean

### DIFF
--- a/backend/app/api/oidc_auth.py
+++ b/backend/app/api/oidc_auth.py
@@ -127,7 +127,9 @@ async def _fetch_userinfo(discovery: dict[str, Any], access_token: str) -> dict[
         return response.json()
 
 
-async def _decode_id_token(discovery: dict[str, Any], id_token: str, nonce: str) -> dict[str, Any]:
+async def _decode_id_token(
+    discovery: dict[str, Any], id_token: str, nonce: str, access_token: str = ""
+) -> dict[str, Any]:
     settings = get_settings()
     jwks_uri = discovery.get("jwks_uri")
     issuer = discovery.get("issuer")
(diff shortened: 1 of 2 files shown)


tassionoronha:
Nice implementation @ADolkun. 
Thanks for providing that. LGTM!
And welcome here :)

PR #351, opened 2026-06-24
fix(simplefin): normalize credit-card balance sign so connected cards don't double-count

ADolkun:
## Problem

A SimpleFIN-connected account that the user re-types to `credit_card` shows the **wrong balance sign** and **double-counts** in the dashboard, sidebar, account list, and account summary — net worth inflates by roughly 2× the card's debt. (The Reports net-worth page uses `abs()` and is unaffected, which is why it's easy to miss.)

## Root cause

Providers disagree on the credit-card balance sign convention:

- **SimpleFIN** reports card debt as a **negative** balance and exposes no account type, so `providers/simplefin.py::_parse_accounts` stores the raw negative value and labels every account `"checking"`.
- **Pluggy / Enable Banking** report card debt as a **positive** number.

The connected-credit-card display logic assumes the **positive-for-debt** convention and negates it (`account_service.serialize_account`, `account_service.get_account_summary`, `account_service.get_account_balance_history`, `dashboard_service._account_balance_at`, and the `sync_opening_balance_for_connected_account` reconciler). So for a SimpleFIN card (stored negative), the negation flips it positive → it's counted as an asset.

This is latent until a user sets the account's type to `credit_card` (a supported edit — `update_account` allows it for connected accounts, #305), because SimpleFIN accounts import as `"checking"` and the negation only fires for `credit_card`.

## Fix

Normalize SimpleFIN credit-card balances to the **positive-for-debt** convention at ingestion, so every downstream site stays provider-agnostic and unchanged:

1. A small provider-guarded helper `_simplefin_to_internal_balance(provider, type, balance)` — flips the sign only for `simplefin` + `credit_card`.
2. The SimpleFIN **sync upsert** applies it using the account's *current* type (which carries the user's override — sync never rewrites `type`). Idempotent: it re-derives from the fresh provider value each sync, so affected accounts self-heal on the next sync.
3. **`update_account`** flips the stored balance when the type crosses the `credit_card` boundary on a SimpleFIN connection, so the correction is immediate instead of waiting for the next sync. The connection is loaded via `session.get` (identity-map hit, no lazy-load), guarded for `connection_id is None`.

Pluggy / Enable Banking and any other provider are untouched (the guard is `provider == "simplefin"`), and their existing credit-card sign tests stay green.

## Note for reviewers

I used a localized `provider == "simplefin"` guard in one helper, matching the "small, focused fix" goal. A more general alternative would be to express the sign convention on the provider contract (e.g. a `BankProvider` flag) so the service layer never name-checks a provider. Happy to switch to that if you'd prefer — it's a slightly larger change and felt out of scope for a bug-fix. (Separately, SimpleFIN importing every account as `"checking"` — the reason this only bites after a manual re-type — is a type-inference gap I'm filing as its own issue.)

## Tests

9 new tests (helper unit cases for simplefin/pluggy/enable; the decisive type-edit flip both directions; sync-branch normalization; Pluggy-unaffected). Decisive test fails before piece 3 (`balance` stays `-500` → resolves to `+500`) and passes after.

- `pytest tests/test_account_service.py tests/test_connection_service.py tests/test_dashboard_service.py` → **147 passed**
- `ruff check` → clean

### DIFF
--- a/backend/app/services/account_service.py
+++ b/backend/app/services/account_service.py
@@ -21,6 +21,21 @@ def get_account_name(account: Account) -> str:
     return account.display_name or account.name
 
 
+def _simplefin_to_internal_balance(provider: str, account_type: str, balance: Decimal) -> Decimal:
+    """Normalize a SimpleFIN balance to Securo's positive-for-debt convention.
+
+    SimpleFIN reports a credit card's balance as negative debt and exposes no
+    account type, so the provider stores it raw and labels every account
+    "checking". Pluggy/Enable report card debt as a positive number, which is
+    the convention every downstream site (serialize_account, _account_balance_at,
+    sync_opening_balance_for_connected_account, ...) assumes. Flip SimpleFIN card
+    balances to match so those sites stay provider-agnostic.
+    """
+    if provider == "simplefin" and account_type == "credit_card":
+        return -balance
+    return balance
+
+
 async def get_accounts(session: AsyncSession, workspace_id: uuid.UUID, include_closed: bool = False) -> list[dict]:
     # Subquery: compute current_balance per account from transactions in one pass
     # Use amount_primary only when tx currency differs from account currency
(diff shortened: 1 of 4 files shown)


tassionoronha:
Thanks for this one too @ADolkun . 
I only use the EnableBanking and Pluggy integrations myself, so I rely on other contributors like you to improve the experience and correctness regarding SimpleFin.

Merging this one as well.

Issue #105, opened 2026-04-21
[Feature] Discussion: User-Defined Account Groups and Filtering


geeosp:
# Feature Request: User-Defined Account Groups and Filtering

Hi everyone, this is a feature suggestion that I belive other users may relate to.

Securo has the potential to hold all the data I need to control my financial life. Right now it is already very useful as a data aggregator. With providers like Pluggy and OFX/CSV imports, it simplifies and gives me more awareness than any other system I've used before. (mostly hand made spreadsheets)

My pain point is that Securo currently groups accounts and reports only by account types, treating all accounts of the same type the same way. However, not every account serves the same objective to me. I'd like to be able to give them meaning.

## Problem

- A user with multiple banks, savings accounts, investment accounts, and credit cards across different institutions can't easily view reports filtered to specific account groups

## Proposed Solution

Create **persistent account groups** that allow users to organize and filter their accounts across Securo's existing dashboards and reports:

### Account Groups / Saved Filters
- Allow users to group accounts with meaningful names (e.g., "Daily Spending Accounts", "Long-term Investments", "Emergency Fund", "Business Accounts", "Credit Cards")
- Groups are persistent and reusable across dashboards and reports
- Groups can include accounts from any institution or type
- An account can belong to different groups simultaneously

### Integration with Existing Pages
- Add a **global "Active Group" selector** (sidebar or toolbar) that filters data across existing pages
- When a group is selected, all dashboards, reports, and transaction views updates to considere only those accounts
- Users can quickly switch between groups or view "All Accounts"
- Persist the last-used group selection per user

### Alternative 1: Simple Filtering UI (No Persistence)
- Just add a checkbox multi-select for accounts on each page
- **Pros**: Simpler to build, less backend work, minimal UI changes
- **Cons**: User has to reselect every time, no saved views, cluttered UI if user has many accounts

### Alternative 2: Flexible User-Defined Dashboards and Custom Reports
- This was actually one of my first thoughts, but the scope felt too big for now
- Create a full dashboard builder with custom widgets (net worth summary, cash flow, category breakdown, investment performance)
- Each dashboard can link to specific account groups with user defined reports and widgets
- Add ability to create custom reports with filters, date ranges...
- All securo views could be built using this system, and all views used by securo would be made available to the user to build its own. 
- **Pros**: Ultimate flexibility, covers all user analysis needs, future-proof extensibility
[...]

Issue #59, opened 2026-04-05
[Feature] Sync investment accounts from Pluggy /investments endpoint


tassionoronha:
## Problem

Pluggy-synced bank connections only import checking and credit card accounts. Investment accounts (brokerage, fixed income, stocks, funds) are not synced, even when they exist in the MeuPluggy dashboard. Additionally, savings accounts (Poupança) are mislabeled as "Checking" because the current mapping uses Pluggy's `type` field instead of `subtype`.

## Proposed Solution

- Add a call to Pluggy's `/investments` endpoint in the `PluggyProvider` to fetch and sync investment holdings as `investment`-type accounts.
- Fix `_map_account_type` in `backend/app/providers/pluggy.py` to use Pluggy's `subtype` field (`CHECKING_ACCOUNT`, `SAVINGS_ACCOUNT`, `CREDIT_CARD`) instead of only `type` (`BANK`, `CREDIT`), so Poupança accounts are correctly labeled as "Savings".
- Surface investment balances in the Accounts page, sidebar, and net worth calculations.

## Alternatives Considered

- Manually adding investment accounts as "Manual Accounts" — works today but requires users to track balances by hand and doesn't scale for brokerage portfolios.
- Treating investments as Assets (existing assets module) — possible, but investments from open finance are naturally account-like and fit better alongside bank accounts.

## Additional Context

- Pluggy API docs confirm `/accounts` only returns `BANK` and `CREDIT` types. Investments are served from a separate `/investments` endpoint: https://docs.pluggy.ai/reference/investments-list
- Related issue: #57
- No new model needed: the investment account type already exists in the frontend config (ACCOUNT_TYPE_CONFIG in frontend/src/pages/accounts.tsx) with its icon (TrendingUp), color, and label. The backend's Account.type is a free-form string, so syncing Pluggy investments just means creating Account rows with type: "investment"

abdalanervoso-debug:
Hi, I’m currently waiting for this solution to be implemented.

In the MeuPluggy dashboard, I can see my investment data (for example, the PicPay “Cofrinho”), but this information is not being fetched or displayed anywhere in the app.

geeosp:
This is THE feature missing in securo right now

tassionoronha:
Thank you guys for pointing that. 
There's also some people from reddit asking for this one, so I'll take action on it this week with some other improvements on the assets page as well :)

tassionoronha:
[...]

PR #356, opened 2026-06-25
Add wallet tracking for goals

### LINKED ISSUE #342: [Feature] Using assets from the same investment for tracking goals
## Problem

When syncing my assets, I noticed that every deposit in my cofrinho at Itaú is shown as a different item. It means that the cofrinho "Reserva de emergência" is not listed as one asset, but as various individual assets.

Image

With this structure is impossible to use my assets linked with my goals, and using a group would let me link them one-on-one.

Image

## Proposed Solution

The simplest solution would be add the wallet as a tracking type inside the goal. It would allow the transfer of the asset to a wallet representing the cofrinho and would serve as a group for the goal.

## Alternatives Considered

If the first suggestion goes against the definition of a wallet, an asset label could be an alternative solution.

## Additional Context

The features to create a wallet and transfer assets to a wallet already exist. The proposed solution would only require the link inside the goals tracking type.

A cofrinho is an investment that emulates a piggy bank inside Itaú.

RomuloGatto:
## Summary
- add `asset_group`/wallet as a goal tracking type so a goal can track the combined value of all assets in a wallet
- persist the linked wallet on goals and return `asset_group_name` in goal API responses
- add the wallet option to the Goals dialog with localized labels/help text across supported locales
- validate linked accounts/assets/wallets stay scoped to the active workspace

## Related issues checked
- Closes #342
- Builds on the wallet/asset-group model from #105 instead of introducing a new label system
- Supports the synced investment/cofrinho asset shape from #59 by allowing several synced asset rows to be grouped and tracked as one goal

## Tests
- `pytest tests/test_goals_api.py tests/test_goal_service_coverage.py -q`
- `ruff check app/api/goals.py app/models/goal.py app/schemas/goal.py app/services/goal_service.py tests/test_goals_api.py tests/test_goal_service_coverage.py alembic/versions/063_goal_asset_group_tracking.py`
- `npm run build`
- `npm run lint -- --quiet`

## Screenshots

Goal list showing a goal tracked from the `Reserva de emergência` wallet:

Goal list with wallet-tracked goal

Add Goal dialog with `Tracking Type` set to `Wallet` and the wallet selector visible:

Add Goal dialog with wallet tracking


tassionoronha:
Amazing work here @RomuloGatto !
Something that was missing indeed in our goals tracking.
LGTM!

PR #369, opened 2026-06-30
feat(period): add account period selector

glaucodasilva:
## What

Adds a reusable period selector for non-credit-card account detail pages, with period navigation and filtering across the account summary, balance history, and transaction list.

This includes:

- Backend period resolution for daily, weekly, monthly, quarterly, half-yearly, yearly, and custom ranges.
- Additive `mode`, `anchor_date`, and `week_start` API support for account summary, balance history, and transactions.
- Frontend period utilities mirroring backend date math.
- Calendar support for opening/selecting at day, month, or year granularity.
- A responsive `PeriodSelector` component aligned with the Reports segmented-control pattern.
- Account detail integration for checking/savings-style accounts while preserving the existing credit-card cycle flow.
- Tests for backend/frontend period resolution and relevant API paths.
- Locale keys for period selector labels and short mode labels.

## Review updates

Addressed the review feedback around polish and PR reviewability:

- Replaced the dropdown/kebab period-type picker with a segmented selector matching the Reports UI pattern.
- Kept period mode selection separate from previous/next period navigation.
- On mobile, the mode selector stacks above the navigator and uses the available width without truncating the selected period label.
- Removed forced period persistence. Reloading or returning to an account starts from the default period, matching the previous account-detail behavior.
- Fixed regular account summary cards so Balance/Income/Expenses follow the selected period instead of staying tied to the initial month.
- Added explicit short labels for all supported locales used by the mode selector.
- Reused the shared date-fns locale helper instead of duplicating locale resolution in the selector.
- Removed English-only quarter/semester labels from the selected period display; quarterly and half-yearly labels now show only localized month range + year.
- Invalid backend period modes now return `422` instead of bubbling as `500`.

## Why

The previous account detail date filter required manually selecting `from` and `to` dates. This made common navigation patterns, such as moving by month, week, quarter, semester, or year, slower than necessary.

The new selector gives users a faster period-based workflow while keeping custom ranges available and preserving the existing credit-card billing-cycle selector.

## Media
Recommended visual attachments for review:

desktop mode switching and period navigation.
mode-change

language switching / localized labels.
language-change

Mobile screenshot: mode selector stacked above the period navigator without truncating the selected label.
image
 

## How to Test

1. Open a non-credit-card account detail page.
2. Switch between Daily, Weekly, Monthly, Quarterly, Half-yearly, Yearly, and Custom.
3. Use previous/next navigation and confirm the period changes by the selected mode.
4. Confirm the Today button jumps to the current period.
5. Confirm Balance/Income/Expenses, balance history, and transactions follow the selected period.
6. On mobile widths, confirm the selector adapts without truncating or shifting awkwardly.
7. Confirm credit-card accounts still use the existing billing-cycle selector.

Validation run locally:

- `cd backend && python -m pytest tests/test_period_service.py tests/test_accounts_api.py tests/test_transactions_api.py`
- `cd backend && python -m ruff check app tests/test_period_service.py tests/test_accounts_api.py tests/test_transactions_api.py`
- `git diff --check`

## Checklist

- [x] Backend lint passes (`ruff`)
- [x] Translations updated for user-facing strings
- [x] Mobile layout reviewed


frontend/src/locales/es.json
```diff
@@ -1660,5 +1660,20 @@
...
+    "quarterly": "Quarterly",
+    "half_yearly": "Half-yearly",
+    "yearly": "Yearly",
+    "custom": "Custom",
+    "today": "Today",
+    "previous": "Previous period",
+    "next": "Next period",
+    "swap": "Swap",
+    "confirm": "Confirm",
+    "cancel": "Cancel"
```
DannyFeliz: None of this is translated

glaucodasilva: Thanks! I translated the period selector labels in Spanish and also updated the other locale files that had the same English fallback.

tassionoronha:
Hey @glaucodasilva, welcome here and thanks for the PR. :)

I found some inconsistencies here that it worth to be flagged:

1. The Income/Expenses cards are broken on regular (non-credit-card) accounts. When you change the period, the chart and the transaction list update, but those two cards stay stuck on the starting month. So the page shows itself contradictory numbers.

2. The design doesn't match the rest of the app. Reports already lets you pick weekly/monthly/etc., but with a clean row of buttons. This new one uses a totally different style (arrows + a dropdown menu). Same idea, two different looks. It also reuses the word "Monthly" to mean something different than Reports does, which is confusing.

3. Switching between two accounts leaks state. Because of how it's built, opening account B can show account A's period, and can even overwrite B's saved setting.

Also some small points like:

- Confirm button is hardcoded Portuguese. It falls back to "Confirmar" even in English, and the translation keys they added for it (periodSelector.confirm/cancel) are sitting unused. Easy fix, visible to users.

- Invalid period value returns the wrong error. A bad mode from the API gives a 500 (server error) instead of a clean 422 (bad request). Low risk since the app only sends valid values, but it's a public endpoint.

I would suggest it to be more polished and also, given the size of the PR, includes more media to make our PR process easier. 

Let me know your thoughts. 
Cheers! :)

glaucodasilva:
> Hey @glaucodasilva, welcome here and thanks for the PR. :)
> 
> I found some inconsistencies here that it worth to be flagged:
> 
> 1. The Income/Expenses cards are broken on regular (non-credit-card) accounts. When you change the period, the chart and the transaction list update, but those two cards stay stuck on the starting month. So the page shows itself contradictory numbers.
> 2. The design doesn't match the rest of the app. Reports already lets you pick weekly/monthly/etc., but with a clean row of buttons. This new one uses a totally different style (arrows + a dropdown menu). Same idea, two different looks. It also reuses the word "Monthly" to mean something different than Reports does, which is confusing.
> 3. Switching between two accounts leaks state. Because of how it's built, opening account B can show account A's period, and can even overwrite B's saved setting.
> 
> Also some small points like:
> 
> * Confirm button is hardcoded Portuguese. It falls back to "Confirmar" even in English, and the translation keys they added for it (periodSelector.confirm/cancel) are sitting unused. Easy fix, visible to users.
> * Invalid period value returns the wrong error. A bad mode from the API gives a 500 (server error) instead of a clean 422 (bad request). Low risk since the app only sends valid values, but it's a public endpoint.
> 
> I would suggest it to be more polished and also, given the size of the PR, includes more media to make our PR process easier.
> 
> Let me know your thoughts. Cheers! :)

Thanks for the detailed review.

I pushed a follow-up commit addressing the points raised:

- Reworked the period type selector to use the same segmented-control pattern used in Reports, instead of the dropdown/kebab menu.
- Kept period navigation separate from mode selection.
- Updated the mobile layout so the mode selector stacks above the period navigator and the selected period label remains fully visible.
- Removed period persistence from account detail, so it resets to the default period when re-entering/reloading the page, matching the previous behavior.
- Fixed the regular account summary cards so Balance/Income/Expenses follow the selected period.
- Added/updated localized short labels for the mode selector.
- Removed English-specific quarter/semester labels from the selected period display; quarterly and half-yearly now show localized month ranges.
- Added backend validation so invalid period modes return `422`, with tests.

I also updated the PR description and added media for desktop mode switching, language switching, and mobile layout.

Validation run locally:

- `cd backend && python -m pytest tests/test_period_service.py tests/test_accounts_api.py tests/test_transactions_api.py`
- `cd backend && python -m ruff check app tests/test_period_service.py tests/test_accounts_api.py tests/test_transactions_api.py`
- `git diff --check`

tassionoronha:
Thanks for the fixes! 

Honest feedback though: this needs real hands-on testing and a feel for the design before pushing. AI-assisted PRs are totally fine, but the output still has to be tried in the browser and judged by eye. This period control isn't clear enough as UX, it's poorly placed, and it doesn't follow the Reports pattern we already use. That consistency is the product we're building. Can you rework and test it with the goal to produce a nice UX? 

Also not sure we need that much granularity on the periods. And please avoid to do changes which are unrelated with the PR goal. Here I see that you chagned the navbar(language switching screenshot), but I dont see the link with this PR. This should be better an individual PR. 

Side note: I really appreciate your intention and and wish to collaborate, but we need to follow and kep some guardrails regarding the code quality, UX and cohesion of what we expect from the product and its features. 

I hope you understand. :)

glaucodasilva:
> Thanks for the fixes!
> 
> Honest feedback though: this needs real hands-on testing and a feel for the design before pushing. AI-assisted PRs are totally fine, but the output still has to be tried in the browser and judged by eye. This period control isn't clear enough as UX, it's poorly placed, and it doesn't follow the Reports pattern we already use. That consistency is the product we're building. Can you rework and test it with the goal to produce a nice UX?
> 
> Also not sure we need that much granularity on the periods. And please avoid to do changes which are unrelated with the PR goal. Here I see that you chagned the navbar(language switching screenshot), but I dont see the link with this PR. This should be better an individual PR.
> 
> Side note: I really appreciate your intention and and wish to collaborate, but we need to follow and kep some guardrails regarding the code quality, UX and cohesion of what we expect from the product and its features.
> 
> I hope you understand. :)

Hello @tassionoronha,

Thanks for the feedback. I understand, and I want to make sure this lands in a way that feels cohesive with the product.

Let me explain the intent behind the PR more clearly.

The problem I was trying to solve is on the account detail page, especially when reconciling transactions against a bank statement. With the current manual date range, moving from one period to another requires changing both start and end dates every time. For that workflow, I felt the missing piece was not only “select a range”, but being able to navigate fixed periods easily, like previous/next week or previous/next month.

That is why I kept a period navigator. A purely static selector like Reports solves “show me the last X period”, but it does not solve as well the account-review workflow of comparing one closed period with another.

That said, I understand your point that the current UX may still not be clear enough or cohesive enough with the rest of Securo. I’m not attached to this exact implementation. What I’d like to understand is the direction you think fits the product better:

- Do you think period navigation itself makes sense for account detail, but the placement/visual design needs to be reworked?
- Or do you think account detail should avoid this kind of period navigation entirely and stay closer to the existing manual date range?
- Would reducing the modes to day/week/month/year/custom make the feature feel more focused?

Also, about the language-switching media: I included it because previous feedback mentioned fixed Portuguese fallback and missing translations, but I see now that it distracts from the PR scope. I’ll remove it from the PR media.

I appreciate the guardrails. My intention is to improve the account reconciliation flow, not to force a feature that does not fit the product.

I’ll wait for your thoughts before making more changes, so I can rework this in the direction that makes the most sense for Securo.

tassionoronha:
Closing this feature, by lack of previous alignement. 
If you intend to continue working on it @glaucodasilva , I would suggest to come to us on Discord, present the idea, then we could iterate from there before jumping in the implementation itself.

Thank you for the effort here :)

PR #374, opened 2026-07-01
fix: harden connection sync correctness

ADolkun:
## What
- prevent SimpleFIN rekeyed accounts from duplicating during partial syncs
- recover transactions that the Bridge exposes weeks after their posted date by rereading a 45-day routine overlap
- honor SimpleFIN exclusive end-date semantics while retaining the 90-day per-request chunk cap
- reconcile provider and imported history in both directions with conservative one-to-one matching
- serialize connected-account imports against sync using a shared connection-row lock
- refresh connections during scheduled syncs and surface reauth-required failures

## Duplicate-safety rules
- require the same account, amount, type, and a date within three days
- require exact normalized merchant or payee text; shared fragments are not enough
- consume each candidate at most once, so genuine repeated purchases remain separate
- never collapse a provider row whose external ID is still present in the incoming batch
- preserve user-owned fields when sync claims an imported row

The 45-day value is the routine reread overlap, not the protocol limit. Requests remain capped at 90 days.

## Stack

Review and merge in this order — each PR builds on the one above it:

1. #374 — connection sync correctness (base of the stack)
2. #443 — workspace payee uniqueness
3. #444 — sync retry reliability (merges last)

GitHub requires a PR's base branch to live in this repository, and these
branches live in a fork, so all three necessarily target `main` rather than
each other. This PR is the base: its diff is its own work only.

## Verification
- Ruff passed on all changed backend files
- 188 tests passed across the modified SimpleFIN provider, import service, and connection service suites
- three independent post-implementation reviews found no blocking issues

## Checklist
- [x] Backend focused checks pass
- [x] Genuine same-day repeats remain distinct
- [x] Import and sync matching is one-to-one
- [x] No Lovenest-only repair data or runtime changes are included

## Rebase note

Merged current `main` (2026-08-09) and resolved the conflicts:
`_surface_errors` keeps its partial-success behaviour alongside the new
`_iso_currency`/`_ticker` helpers, the import dedup keeps one-to-one
matching, and both sides' new tests are retained.

Re-verified after the merge: Ruff clean, `ty check` clean, full backend
suite 2421 passed / 7 skipped.

Improves connection synchronization and import matching to prevent duplicate accounts and transactions. It also recovers delayed transactions and reports provider reauthentication failures.

- Keeps delayed transactions during routine syncs.
- Preserves separate identical transactions.
- Prevents duplicate accounts during partial syncs.
- Refreshes providers during scheduled syncs.
- Reports when a connection needs reauthentication.
- Preserves user-edited transaction and account fields.

Risk: sync provider and auth.


tassionoronha:
Thanks for this, the partial-sync and reauth handling is a really nice improvement. One thing I want to sort out before merging.

In the posted-fingerprint dedupe path (`_find_synced_duplicate`, Path 3), the match requires the same status and often `transacted_at` is missing for SimpleFIN, so two genuinely separate but identical charges on the same day (think two coffees for the same amount at the same place) could get collapsed into one.

Could we only treat a row as a rekey when its old `external_id` is actually gone from the incoming batch? A test with two legit identical same-day transactions staying separate would give us confidence too.

Small extra note: the oauth callback rekey branch overwrites `raw_data` instead of merging like the sync path does, might be worth unifying.

Everything else looks good, 130 tests passing on my end.

ADolkun:
Addressed in 005b4b4.

- The posted-fingerprint path now only rekeys when the candidate old external ID is absent from the full incoming batch, in both manual callback and regular sync flows.
- Added a regression proving two legitimate identical same-day transactions remain separate.
- OAuth callback rekeys now use the shared raw-data merge helper, preserving old-only fields while fresh provider values win collisions.
- Rebased onto current upstream main and resolved the stale conflicts.

Validation: Ruff passed; 141 focused backend tests passed, followed by an independent review and focused API recheck.

ADolkun:
Updated with the delayed SimpleFIN history fix proven in downstream use.

The new commit adds the 45-day routine rewind with correct exclusive end dates, conservative one-to-one import/provider reconciliation, and complete import-vs-sync serialization. It does not include any Lovenest-specific data repair.

Validation: Ruff passed; 188 affected-suite tests passed; three independent reviews approved the final diff. PR #444 should be rebased after this PR and drop its duplicate connection-lock hunk.

ADolkun:
Resolved against current `securo-finance/securo:main` at `5daf3b2` (2026-08-23). Preserved the PR’s one-to-one import/sync matching and SimpleFIN re-key handling alongside upstream’s `original_description`, recurring-match, and provider-configuration changes. Local verification: Ruff and 209 focused backend tests passed; the full stacked backend suite later passed 3,105 tests / 45 skipped.

ADolkun:
Updated this branch against current upstream main and addressed the integration review findings.

Changes in the refreshed head:
- scheduled provider refreshes now use a persisted clock independent of last_sync_at, so the four-hour cached sync cadence cannot postpone the daily provider refresh forever
- SimpleFIN account rekeys reserve every incoming account ID, stay within the resolved institution, preserve closed accounts, and refuse ambiguous same-name/currency matches
- reverse import reconciliation now also applies to identifier-bearing imports and matches rule-preserved original_description text
- added regressions for cross-institution selection, later-batch IDs, closed accounts, same-institution ambiguity, scheduled refresh persistence, and OFX/provider ID reconciliation

Validation:
- Ruff: passed
- ty: passed
- full backend before the final two edge-case follow-ups: 3,173 passed / 45 skipped, 92.48% coverage
- final focused regressions: 4 passed
- Alembic: one head at 075
- independent final re-review: no blockers

Head: e0779c3

ADolkun:
Refreshed onto current `main` (`2869ff40`) — this was conflicted. Merged rather than rebased, matching how the branch already integrates upstream. Full backend suite green afterwards: **3837 passed, 7 skipped** (the skips are the pre-existing `to_char()`-is-PostgreSQL-specific guards), plus `ruff check` and `ty check` clean. No migrations on this branch, so no numbering collision with head `089`.

Three files conflicted. One resolution is worth calling out because it wasn't a straight pick-a-side: `_description_similarity` looked like "mine adds a block, upstream adds nothing", but upstream had actually **moved** it into `app/services/text_similarity.py` as `token_overlap`. I adopted the move and pointed this branch's call site at it rather than resurrecting a deliberately-deleted symbol — the formulas are equivalent, so no behavior change. The other two were the initial-import loop (kept both `incoming_external_ids` and `seen_external_ids`; they answer different questions) and the account-sync loop, where I re-applied #750's re-ownership block after this branch's `_find_existing_connected_account` fallback, so rekeyed accounts now get re-owned too.

---

Separately, I want to be straight about the scope here, because this PR is no longer the thing you reviewed.

Your review on 2026-07-02 covered the SimpleFIN rekey/dedupe fix, and all three asks were addressed in `005b4b45` on 2026-07-18. **Everything after that commit, you have never seen.** Two later pushes added roughly +740 lines:

- **`71e315d3`** — SimpleFIN provider windowing: default history 90 → 45 days with a full-window rewind on every sync instead of trusting the cursor, exclusive end-date handling, and a new pure `transaction_match_service.py` used by both connection and import paths.
- **`951891f6` + `e0779c3b`** — a persisted scheduled-refresh clock (`last_provider_refresh_at` in `connection.settings` rather than inferring from `last_sync`), and ambiguity rejection: the old "sort candidates by heuristic, take the best" is replaced with a flat "if there isn't exactly one candidate, don't match."

That last one is a deliberate policy change and the direct descendant of your Path-3 concern — rather than trying to be cleverer about ranking near-duplicates, it declines to guess.

These are two coherent, separable stories, and the reviewed fix is everything up to `005b4b45`. **Happy to split this into three PRs if that's easier to review** — say the word and I'll cut them. I'd rather do that than have you approve a diff three times the size of the one you looked at.

tassionoronha:
LGTM now @ADolkun . 
Thanks for the effort to keep it updated, I really appreciated that.

PR #375, opened 2026-07-01
fix: harden transaction and rule safety

ADolkun:
## What
- enforce workspace boundaries for transaction/rule writes
- validate rule actions and transaction references more defensively
- compare rule dates as dates
- add explicit rule backfill and category-overwrite controls

## Why
These are related data-safety fixes around rule application and transaction mutation boundaries.

## How to Test
- `cd backend && .venv/bin/ruff check ...` for touched rule/transaction files and tests
- `cd backend && .venv/bin/pytest -q tests/test_import_service.py tests/test_new_rules_api.py tests/test_providers_simplefin.py tests/test_rule_engine.py tests/test_rule_service.py tests/test_transaction_service.py tests/test_transaction_service_coverage.py tests/test_transactions_api.py tests/test_transactions_api_coverage.py` -> 388 passed

## Checklist


tassionoronha:
Really like this one, the rule dialog controls and the date comparison fix are great, and I tested the new checkboxes in the UI and they behave exactly as expected. There is one thing I need us to fix before merge though.

I reproduced a data issue: `get_transaction` sets `is_ignored = True` for display when the category is ignored, but `update_transaction` calls it and then commits, so editing any field (even just the notes) on a transaction in an ignored category permanently saves `is_ignored = True`. It then stays ignored even after you move it back to a normal category, so it quietly disappears from income and expense totals.

Could we move that flag to the read/serialization layer instead, since the summaries already exclude by category anyway? A quick regression test for that flow would be perfect.

Two smaller things while we are here: in `import_rules` a single bad rule now fails the whole import instead of skipping, and the dialog defaults apply-to-existing to false on edit but true on create, just want to confirm that is intentional.

ADolkun:
Addressed in 22a6ab0.

- Category-derived ignored state is now applied only by TransactionRead serialization, so editing notes or other fields cannot persist it to the transaction row.
- Added a regression covering edit-in-ignored-category, persisted database state, and moving back to a normal category.
- Rule import again skips invalid rules individually while importing valid siblings.
- Create=true / edit=false for apply-to-existing is intentional: creation keeps the existing immediate-backfill default, while edits require explicit opt-in before rewriting history.

Validation: Ruff passed; 135 relevant backend tests passed; frontend lint and build passed; independent review approved.

ADolkun:
Follow-up: rebased onto current upstream main and added the same four rule strings to the newly supported German locale in bf722e6. This fixes the locale-parity failure seen on the merge commit. Locale tests are now 29/29; focused backend tests, Ruff, frontend lint, and build all pass locally.

tassionoronha:
Good work here @ADolkun !
And thanks for keeping the branch updated.
I resolved a conflict and fixed some translations. 

LGTM!

PR #376, opened 2026-07-01
fix: tighten transaction and rule frontend flows

ADolkun:
## What
- fix transaction import/review date handling and export filter payloads
- reset rule dialog fields correctly
- confirm destructive rule reapply and improve transaction bulk rule UI handling

## Why
These are frontend-only corrections for transaction/rule workflows.

## How to Test

## Checklist


tassionoronha:
Nice work, I checked the field-change reset in the rule dialog and the bulk add and remove tags in the browser and they all work great.

Just one small thing missing before merge: two of the new strings are hardcoded English fallbacks and are not in the locale files yet, `rules.confirmReapplyAll` and `transactions.bulkRemoveTags`. Its sibling `bulkAddTags` already exists across the locales, so it would be great to add matching keys (at least EN and PT-BR).

Heads up too that this PR keeps the "Reapply all" button while #375 renames it to "Reset and reapply", so depending on merge order those two will need reconciling.

ADolkun:
Addressed in 75516d8.

- Added rules.confirmReapplyAll and transactions.bulkRemoveTags to all 8 supported locales, including EN and PT-BR.
- Rebased onto current upstream main and resolved the stale frontend conflict.
- Kept this PR aligned with current main wording; PR #375 still owns the Reset and reapply rename, so merging #376 before #375 avoids reverting that later wording.

Validation: locale parity tests, frontend lint, and production build passed; independent review approved.

PR #379, opened 2026-07-01
fix: align dashboard drilldown totals

ADolkun:
## What
- carry dashboard drilldown filters into transaction queries
- expose the necessary dashboard filter fields in API schemas/client types
- add backend coverage for totals/list parity

## Why
Dashboard cards and their drilldown transaction lists should agree on the same filtered totals.

## How to Test
- `cd backend && .venv/bin/ruff check app/api/transactions.py app/schemas/dashboard.py app/services/dashboard_service.py app/services/transaction_service.py tests/test_dashboard_service.py tests/test_transaction_service_coverage.py tests/test_transactions_api.py`
- `cd backend && .venv/bin/pytest -q tests/test_dashboard_service.py tests/test_transaction_service_coverage.py tests/test_transactions_api.py` -> 132 passed

## Checklist


tassionoronha:
This is a solid fix, I opened the income and expense drilldowns and the footer totals now match the cards exactly, so the main problem is definitely solved. Two small things I would love to close out since they are in the same area.

The card total filters out closed accounts but the drilldown query does not, so an in-month transaction on a closed account would make the footer come out higher than the card.

And it would be great to add a test that asserts the summary total equals the sum of the drilldown list for a month, just so a future filter change cannot quietly break this again.

Tiny thing: the new `ProjectedTransaction.is_ignored` field is not read anywhere, so we can probably drop it.

Thanks for tackling this one.

ADolkun:
Addressed in c3a3365.

- user_pnl_only drilldowns now exclude closed accounts exactly where the dashboard summary does; normal account ledgers are unchanged.
- Added a regression asserting monthly summary income/expense equals the corresponding drilldown row sums while closed-account activity is excluded from both.
- Removed ProjectedTransaction.is_ignored end-to-end from the backend schema and frontend type/usage.

Validation: Ruff passed; 193 relevant backend tests passed; frontend lint and build passed; independent review approved.

PR #381, opened 2026-07-02
feat: link recurring bills to actual transactions (#116)

### LINKED ISSUE #116: [Feature] Request: Automatic linking of recurring bills with transactions
Problem description
Currently, it is already possible to create recurring fixed bills (e.g., subscriptions, rent, etc.).
However, when these bills are actually charged to a bank account or credit card, the system records a separate transaction, resulting in duplicated entries.

This forces users to manually delete one of them, which negatively impacts the user experience and increases the risk of inconsistencies in financial tracking.

Proposed solution
Implement a mechanism for automatic linking between recurring bills and actual transactions.

Suggested approaches:

Use rule-based matching considering:
Payee/beneficiary name
Amount (or amount range tolerance)
Frequency (monthly, weekly, etc.)
Allow association through:
Beneficiary configuration
Custom rule engine
When a match is identified:
Link the transaction to the existing recurring bill
Prevent duplication

Expected benefits

Elimination of duplicate entries
Reduced manual work
Improved accuracy in financial tracking
Better overall user experience

tassionoronha:
Closes #116.

Recurring bills now reconcile with the real synced/imported charges that pay them, instead of appearing as duplicate transactions.

- Auto-links an incoming charge to a matching bill (same account, currency, direction, exact amount, a date window around the expected occurrence, and payee similarity); high-confidence matches link silently.
- `generate_pending` links an existing charge instead of writing a duplicate placeholder, and stamps the link on placeholders it does create.
- Per-bill "generate automatically" toggle: on by default (current behavior), off keeps the bill as a projection until the real charge is matched.
- Unlink action on a linked transaction as an escape hatch.

Additive, nullable migration — existing bills keep the current behavior.

Verified with new backend tests (matcher units + sync/import/generate/unlink integration) and an end-to-end browser check (toggle, badge, link/unlink, and the unlinked-transaction negative case).


RomuloGatto:
I found one blocker before merge: early-posted charges inside the before-window link successfully, but don’t advance next_occurrence, because callers pass txn_data.date into advance_past(). A Jan 8 charge for a Jan 10 occurrence remains at Jan 10, so generate_pending can later create a duplicate. I’d fix by advancing past the matched occurrence date, not the posting date, and adding early-match regression tests.

tassionoronha:
> I found one blocker before merge: early-posted charges inside the before-window link successfully, but don’t advance next_occurrence, because callers pass txn_data.date into advance_past(). A Jan 8 charge for a Jan 10 occurrence remains at Jan 10, so generate_pending can later create a duplicate. I’d fix by advancing past the matched occurrence date, not the posting date, and adding early-match regression tests.

Thanks for pointing that @RomuloGatto. 
Just fixed that :)

PR #386, opened 2026-07-03
Allow disabling asset sync per connection

### LINKED ISSUE #349: [Feature] Allow disabling asset sync from a given source
## Problem

When adding linked account, assets are automatically imported from it without asking the user beforehand. The app also doesn't offer an way to opt out from this behavior.

## Proposed Solution

Add a new option in "Connection settings" to define wheter assets should be imported from it. Bonus point if it also defines the behavior for other items (accounts, transactions, etc)

## Alternatives Considered

-

## Additional Context

This was observed when using Pluggy connector with an Itau account

RomuloGatto:
## Summary
- add a per-connection `sync_assets` setting, defaulting to enabled for existing behavior
- skip provider investment holding imports on initial connect and later syncs when asset sync is disabled
- expose the option before new widget/OAuth/token connections and in Connection settings
- add backend coverage for settings persistence, initial opt-out, redirect-state opt-out, disabled sync skipping holdings, and default enabled behavior

Closes #349

## Screenshots

Connection settings can disable investment asset imports for an existing bank connection while keeping account/transaction sync enabled:

Connection settings showing Import investment assets unchecked

New connections show the same asset-import opt-out before the provider connector opens:

Initial sync options showing Import investment assets unchecked

## Tests
- `unset PYTHONPATH PYTHONHOME && . /Users/romulogatto/Fontes/securo/backend/.venv/bin/activate && python -m compileall -q app tests && pytest tests/test_connection_settings_api.py tests/test_connection_service.py tests/test_sync_holdings.py -q` — 91 passed
- `npm ci`
- `npm run build` — passed
- `npm run lint` — passed with existing warnings only (40 warnings, 0 errors)
- `npm test` — 36 passed
- `alembic heads` — `063 (head)`
- fresh `pgvector/pgvector:pg16` Postgres + `alembic upgrade head` — passed, current `063 (head)`

## Notes
- No schema migration is needed because the new flag is stored in the existing `bank_connections.settings` JSON field.


RomuloGatto:
@tassionoronha this may get into the same point of the latest PR, as it touches core business as well, feel free to proceed as you wish

tassionoronha:
Hey @RomuloGatto. 

Thanks for this one. Actually this one doesn't touch necessarily core rules of how we manage the finance itself, it's well contained in the bank sync logic, which is fine. 

I'll just add a commit to make it only visible for integrations which supports assets, for instance, only Pluggy and SimpleFin.

PR #387, opened 2026-07-04
fix: stop grouping assets in the pie chart  into "others"group when t…

geeosp:
## What
Small tweak on the grouping and color choosing of the pie chart of networth reports page. 
The idea is to stop grouping assets if there is only one of that kind, avoiding neighbor sessions of gray color if possible.
In my use case, most of my chart is covered by assets, and the sum of both accounts and liablities falls below the grouping threshold, but there is only one of each. this ends up creating 3 neighnbor sessionss in gray, "other assets', "other accounts" and "other liablities".  

Below there are screenshots of before and after

https://github.com/user-attachments/assets/84c9b9da-90c7-4092-878d-aff4f843520f

Screenshot 2026-07-03 at 21 31 04

## Checklist


tassionoronha:
LGTM, thanks @geeosp !

PR #396, opened 2026-07-06
Fix SimpleFIN stale access URL reconnect flow

### LINKED ISSUE #332: [Bug] SimpleFIN: stale access URL -> sync 500 with no non-destructive recovery (Reconnect is Pluggy-only)
When a SimpleFIN access URL becomes invalid (the bridge returns 403 — happens when the bank-side auth lapses or the bridge rotates/revokes the URL), connection sync fails with 500 Internal Server Error and there is no working non-destructive recovery path. "Reconnect" doesn't work for SimpleFIN, so a user's only option is to delete + re-add the connection, which cascades to accounts/transactions and forces full re-categorization and rule re-creation.

## Steps to Reproduce

1. Connect a bank via SimpleFIN.
2. Let the access URL go invalid (bank re-auth lapses, or revoke/regenerate at the bridge). GET {access_url}/accounts now returns HTTP 403.
3. Trigger a sync in Securo.

## Expected Behavior

Sync fails gracefully, the connection is flagged as needing re-auth, and the user can paste a fresh SimpleFIN setup token to refresh the stored access URL in place — without losing accounts/transactions/categorizations.

## Actual Behavior

- POST /api/connections/{id}/sync → 500 Internal Server Error.
- Connection status stays active, so the UI never surfaces an error/reconnect affordance.
- Clicking Reconnect shows "failed to connect bank" with no token-entry field. (On an earlier v0.12.x, POST /api/connections/{id}/reconnect-token returned {"detail":"Connection has no item_id for reconnection"} — item_id is a Pluggy concept, so the reconnect path appears hard-coded for Pluggy and never handles SimpleFIN.)
- Remaining options are destructive and wipe transaction history + categorizations.

## Environment

- Securo v0.13.1, self-hosted (docker-compose.prod.yml), Postgres (pgvector/pgvector:pg16)
- Provider: SimpleFIN (beta-bridge.simplefin.org)

## Additional Context

Root cause (from reading the code): 

- app/providers/simplefin.py stores the credential as {"access_url_enc": encrypt(access_url)} (encrypt/decrypt from app/agents/services/crypto.py).
- When the bridge returns 403, sync surfaces it as an unhandled 500 rather than a typed "needs reauthorization" result, and the connection is never marked error.
- The reconnect flow requires item_id (Pluggy), so SimpleFIN connections cannot be reconnected via the UI.
- Note: SimpleFinProvider.handle_oauth_callback(setup_token) already implements exactly the needed flow (decode setup token → claim → encrypt(access_url) → return credentials). A SimpleFIN reconnect just needs to reuse it to overwrite access_url_enc on the existing connection.

## Suggested fix

1. Make reconnect provider-aware: for SimpleFIN, accept a fresh setup token and reuse the existing handle_oauth_callback claim+encrypt logic to overwrite credentials.access_url_enc on the existing connection (preserving accounts/transactions). Don't require item_id for non-Pluggy providers.
2. On sync, catch the SimpleFIN 403 (and con.auth / gen.auth errlist codes) and set the connection to error / needs_reauth instead of returning a bare 500, so the UI surfaces the reconnect affordance.

## Workaround (for others who hit this, no data loss)
Refresh the access URL in place via the DB:

```
# Find your connection id:
docker exec securo-db-1 psql -U postgres -d securo -c \
  "SELECT id, institution_name, status FROM bank_connections WHERE provider='simplefin';"

# Generate a fresh Setup Token at your SimpleFIN bridge, then:
CONN_ID='<connection-id>'
TOKEN='<fresh-setup-token>'
NEWENC=$(docker exec securo-backend-1 python -c "
import base64, httpx
from app.agents.services.crypto import encrypt
claim=base64.b64decode('''$TOKEN''').decode().strip()
url=httpx.post(claim, timeout=30).text.strip()
assert httpx.get(url.rstrip('/')+'/accounts', timeout=30).status_code==200
print(encrypt(url))
")
docker exec securo-db-1 psql -U postgres -d securo -c \
  "UPDATE bank_connections SET credentials = jsonb_set(credentials::jsonb,'{access_url_enc}', to_jsonb('$NEWENC'::text))::json, status='active' WHERE id='$CONN_ID';"
# Then Sync in the UI — accounts, transactions, and categorizations are preserved.
```

RomuloGatto:
## Summary
- handle stale SimpleFIN access URLs as typed reauthorization failures instead of generic sync 500s
- mark connections as errored when provider credentials require user action so the Accounts page surfaces reconnect affordance in the connection action row
- add a non-destructive token reconnect path for SimpleFIN that refreshes stored credentials in place and preserves accounts/transactions/categorizations
- route token-provider reconnects to the setup-token dialog instead of the Pluggy widget reconnect path

Closes #332

## Screenshots

SimpleFIN stale credential state now uses the existing refresh/reconnect action: the refresh icon turns amber and carries a warning badge in the top-right instead of adding a separate action icon or banner.

SimpleFIN reconnect warning badge on the refresh action

Reconnect opens a fresh setup-token dialog and explains that existing data is preserved:

SimpleFIN token reconnect dialog

## Tests
- `unset PYTHONPATH PYTHONHOME && . /Users/romulogatto/Fontes/securo/backend/.venv/bin/activate && python -m compileall -q app tests && pytest tests/test_connections_api.py tests/test_connection_service.py tests/test_providers_simplefin.py -q` — 112 passed
- `pytest tests/test_connections_api.py tests/test_connection_service.py tests/test_connection_service_coverage.py tests/test_providers_simplefin.py -q` — 132 passed
- `pytest -q` — 2269 passed, 7 skipped
- `ruff check app/api/connections.py app/schemas/bank_connection.py app/services/connection_service.py tests/test_connections_api.py tests/test_connection_service.py tests/test_connection_service_coverage.py` — passed
- `npm ci`
- `npm run build` — passed after the integrated-icon UI follow-up
- `npm run lint` — passed with existing warnings only (40 warnings, 0 errors)
- `npm test` — 36 passed
- `alembic heads` — `064 (head)`
- fresh `pgvector/pgvector:pg16` Postgres + `alembic upgrade head` — passed, current `064 (head)`


tassionoronha:
Hey @RomuloGatto, thanks for this fix. 
Did you test with a real SimpleFin connection? 

@FugginBeenus can u confirm if this resolves the issue?

RomuloGatto:
@tassionoronha Yes, just tested with the SimpleFin demo account. Connected to it, then changed the access_url directly in the DB to  'https://bad:bad@beta-bridge.simplefin.org/simplefin and asked to sync on the UI. It failed, showed the warning on the refresh icon, and then clicking on it asked for a new token, and then everything got synced again

FugginBeenus:
Finally tested and the issue is resolved. Thanks.

PR #397, opened 2026-07-06
Fix asset wallet chart curves

### LINKED ISSUE #341: [Bug] Graph on assets page still not showing the correct curves
## Description

When showing the assets graph grouped by wallet, the lines shown don't match the correct values. Even after the correctionn that closed the #163.

## Steps to Reproduce

1. Go to Assets
2. Click on the "By wallet" option
3. See the error in the graph lines

## Expected Behavior

The lines should reflect the value in each wallet.

## Actual Behavior

The lines don't align with the values in each wallet.

## Environment

- Deployment: Docker
- Browser: Safari 26.5
- OS: macOS Tahoe 26.5.1

## Screenshots

Image

## Additional Context

In the screenshot both lines appear very close, but the proportions would be:

Itaú: 97% of the total amount
Banco do Brasil: 3% of the total amount

RomuloGatto:
## Summary
- render the Assets portfolio chart as unstacked line curves so each wallet/asset line is plotted at its own value
- keep the existing wallet/asset grouping, legend ordering, and tooltip totals while removing the cumulative `stackId` behavior that made small wallets appear near the portfolio total
- add a regression test documenting the #341 stacked-area failure mode and guarding the PortfolioChart against reintroducing stacking

Fixes #341

## Screenshot

The seeded evidence below uses two wallets at the same proportions from the bug report: Itaú = R$97,000 (97%) and Banco do Brasil = R$3,000 (3%). Banco do Brasil now renders near the baseline instead of clustered near the Itaú/total curve.

Assets wallet curves

## Verification
- `npm ci` ✅
- `npm test -- assets-chart-rendering` ✅ `2 passed`
- `npm test` ✅ `38 passed`
- `npm run build` ✅
- `npm run lint` ✅ exits 0; existing warnings remain
- `git diff --check` ✅
- Manual browser screenshot with seeded 97% / 3% wallet data ✅


tassionoronha:
Thanks for opening this one @RomuloGatto. 
As we are discussing on the #341 , this is not an actual bug so removing the stacked chart would be the best solution here. What I proposed is to have a toogle(or another mechanism) to switch between individual lines or stacked. 

Since this solution is a considerable change, I'll close this PR.
Be free to open another one and engage in the discussion in the issue in case you want to resolve issue.
Thanks! :)

PR #401, opened 2026-07-07
Add cash-flow transaction calendar view

### LINKED ISSUE #400: [Feature] Add a cash-flow transaction calendar
## Problem

The transactions page is optimized for list/table workflows, but it is hard to plan cash flow by day. Users need a calendar-style view that shows daily ending balances, actual and projected transaction activity, and quick entry actions without losing context or scrolling away from the selected day.

The existing filters also make calendar balance semantics confusing when arbitrary multi-account subsets are selected. For a cash-flow calendar, the view should be either all accounts / unified balance or one selected account.

## Proposed Solution

Add a transaction calendar view to the transactions page with:

- Month grid showing compact end-of-day balances per calendar cell.
- Markers for income, expenses, transfers, and projected recurring items.
- Marker tooltips that explain what happened without exposing amounts.
- A selected-day detail sidebar on desktop/tablet widths, with transactions listed on the right instead of below the calendar.
- Compact add income, add expense, and transfer shortcuts in the selected day cell that do not cause row/cell growth.
- Selected-day summary panels for income, expenses, and transfers, each with its action button attached to the panel.
- Calendar account scope limited to all accounts or exactly one account.
- Previous/next month controls following the existing month-stepper design.
- URL persistence for calendar mode, selected date, month range, and account scope.
- Dark-mode-safe styling for markers, selected cells, and shortcut buttons.

## Alternatives Considered

- Keep transactions in the normal table and rely on date filters only. This does not give a month-level cash-flow view or quick per-day planning.
- Put selected-day details below the calendar. This works on mobile, but on desktop it forces scrolling and makes the selected-day context feel disconnected.
- Allow arbitrary multi-account calendar subsets. That keeps parity with list filters, but makes the primary balance number harder to reason about than all accounts or one account.
- Show a cash-flow line graph inside the grid. This was explored visually but was too noisy/misleading for the calendar cells, so the calendar keeps the compact balance value instead.

## Additional Context

This is intended as a UX enhancement for planning transactions and projected cash flow directly from the transactions page. The implementation should reuse the existing transaction and transfer dialogs for mutations rather than introducing new write paths.

RomuloGatto:
## What

Adds a cash-flow transaction calendar view to the transactions page.

Highlights:

- Month grid with compact end-of-day balance values.
- Calendar mode uses the page-level month/account controls as the source of truth instead of duplicating navigation/scope controls inside the card.
- Compact inline legend for income, expenses, and transfers.
- Negative/low-balance days keep a stronger red/rose cell treatment without an extra low-balance marker.
- Projected recurring rows stay labeled in detail/preview rows, but are no longer a standalone calendar-grid marker/background.
- Persistent Compact/Detailed calendar density toggle:
  - Compact keeps the balance/marker-focused layout.
  - Detailed shows the top 1–3 transaction previews plus `+N more`.
- In-cell credit/debit/transfer shortcuts are hover/focus-only, so they are available from the day cell without being permanently visible.
- Desktop/tablet selected-day details stay in a right sidebar instead of below the calendar.
- Sidebar focuses on selected day, balance, and transaction list; duplicate income/expense/transfer action panels were removed.
- Calendar account scope is constrained to all accounts or exactly one selected account.
- URL persistence for `view=calendar`, selected `day`, date range, and account scope.
- Dark-mode-safe selected-cell, marker, low-balance, density toggle, detailed preview, and hover-shortcut styling.
- PR screenshots committed under `docs/transaction-calendar/` using seeded demo data.

## Why

The transactions table is useful for reviewing rows, but it is hard to plan cash flow by day. A calendar view makes the end-of-day balance and the daily transaction mix visible at a glance.

Forecasting is the main value of this view, but projected status works better as row/detail metadata than as another grid-level marker. The final UI keeps the month grid focused on balances plus actual income/expense/transfer signals, while detailed mode and the selected-day sidebar still distinguish projected rows from posted transactions.

## How to Test

Verified locally:

1. `cd backend && uv run --extra dev pytest tests/test_transaction_calendar_service.py -q`
   - `4 passed, 3 warnings`
2. `cd frontend && npm run build`
   - passed
3. `cd frontend && npm test -- --run`
   - `36 passed`
4. `cd frontend && npm run lint`
   - passed with `0 errors`; existing warnings remain
5. Locale JSON validation
   - passed
6. `git diff --check`
   - passed
7. Manual browser check with seeded demo data:
   - seeded via `backend/scripts/seed_perf.py --scale 0.02 --accounts 2 --categories 8 --assets 3 --start-date 2026-01-01`
   - opened `/transactions?view=calendar&from=2026-07-01&to=2026-07-31&day=2026-07-07`
   - verified selected day restores from the URL and the sidebar stays on the right
   - verified all-account and single-account calendar scopes
   - verified no duplicated calendar-card navigation/account controls
   - verified Compact/Detailed toggle, detailed previews, `+N more`, hover-only in-cell shortcuts, clean legend, and projected demotion
   - verified dark and light mode screenshots

CI after the latest push:

- Backend Tests: pass
- Frontend Checks: pass

## Screenshots

Dark mode, all accounts, Detailed calendar density:

Transaction calendar dark mode, all accounts, detailed density

Light mode, single account, Compact calendar density:

Transaction calendar light mode, single account, compact density

## Related Issue

Closes #400

## Checklist

- [x] Backend calendar tests pass (`python -m pytest tests/test_transaction_calendar_service.py`)


frontend/src/locales/es.json
```diff
@@ -482,7 +482,22 @@
...
+    "calendarAllAccounts": "All accounts",
+    "calendarEndBalance": "End balance",
+    "calendarItemCount": "{{count}} item",
+    "calendarItemCount_plural": "{{count}} items",
+    "calendarProjectedCount": "{{count}} projected",
+    "calendarProjected": "Projected",
+    "calendarAddIncome": "Add income",
+    "calendarAddExpense": "Add expense",
+    "calendarSelectedDay": "Selected day",
+    "calendarNoItems": "No transactions on this day"
```
DannyFeliz: Check the translations, the values are not translated.

frontend/src/pages/transactions.tsx
```diff
@@ -1487,14 +1603,15 @@ export default function TransactionsPage() {
           </div>
         </div>
       )}
+        </>
```
DannyFeliz: Do we need this?

RomuloGatto: Not at all, removed.

frontend/src/locales/es.json
RomuloGatto: it is all covered now

tassionoronha (commented):
Really nice work on this. The calendar is a genuinely useful addition, the projected recurring logic is the right idea, and the dark mode craft is lovely. I took it for a spin with a multi account, multi currency dataset and read through the service. A few things I'd love to see addressed before we merge, one correctness issue plus some UX refinements. None of it changes the core approach, which is solid. 🙌

## 1. Balance is off when scoped to a single account in a non primary currency (blocking)

When you filter the calendar to one account whose currency is not the workspace primary currency, the running balance comes out wrong. The reason is that the starting balance and the daily deltas end up in different currencies and get added together.

`_balance_at` returns the native currency total without converting when there's only one currency in play, so the baseline stays in the account's own currency. The per day deltas, meanwhile, are converted to the primary currency. Adding those two produces a number that is neither.

The cleanest fix is to make `_balance_at` convert to primary even in the single currency case when that currency differs from primary, so the baseline and the deltas always share a unit. A quick regression test with a single foreign currency account scope would lock this in. Everything works correctly in the all accounts case, so this only shows up in the single account path.

## 2. The navigation controls are duplicated

Right now the month and the account scope each appear in more than one place at once. The page already has a month stepper and an account filter, and the calendar card renders its own month stepper and its own account dropdown on top of that. It's not obvious which one is authoritative.

I'd let the calendar consume the page level month stepper and account filter, and drop the copies inside the card. One control per concept keeps the header calm and removes the "which one do I use" question.

## 3. The day cells could carry a lot more

Each cell currently shows a balance and a few small dots, so you can't tell what actually happened on a day without clicking into it. Since the item list is already in the payload and you already have the category icon component, the cells are a great place to surface the top one to three transactions with their category icon and amount, then a "+N more" when a day is busy. The balance can move to a small footer chip. This is what makes a calendar worth having over the list: you scan the month, you don't click through it.

## 4. The in cell add buttons feel redundant

The add income, add expense, and transfer shortcuts inside the selected cell overlap with the same actions in the side panel and in the page header, so the same three actions live in three places. They also reserve a bit of vertical space in every cell even when it isn't selected. I'd remove them from the cell and keep those actions in the side panel where there's more room and context, which frees the cell up for the content in point 3.

## 5. Give the markers a legend, and make projected and low balance states louder

The colored dots are hard to decode without a legend, and the projected marker is subtle even though projections are what move the forecast. Two small additions would go a long way: a compact legend for the markers, and a distinct treatment for projected days versus posted ones. A cash flow calendar's best trick is warning you about an upcoming low balance day before it arrives, so a dedicated warning state for days that dip below zero (or below a chosen buffer) would really make the feature sing.

## Minor notes (non blocking)

- The recurring projection query doesn't filter out closed accounts, while the actual transactions do, so a recurring item on a closed account can still affect the projected balance. Worth aligning.
- `transferDefaultDate` isn't cleared on a successful transfer, only on close, so the next transfer opened from the header can prefill a stale date.
- The service recomputes per transaction FX in a Python loop. There's an existing helper that does this as a single grouped query and would keep the calendar consistent with the rest of the app. Optional, but nice for multi currency workspaces.

Happy to pair on any of these if useful. Thanks again for the feature, it's a great direction. 🙏

tassionoronha:
Hey @RomuloGatto, posted this with Claude help but also contains my own thoughts and perception about the feature itself. I think in general is a good idea, but still missing some UI refinement. 

Think also to this kind of feature to do some research before and have some benchmark for the feature functionality itself but for the UX as well. In case you do/did that, be free to share the sources. 

Thanks!

RomuloGatto:
Good catches, thanks for the detailed pass. I’ll take a look later today, especially at the currency-scoped balance issue and the smaller correctness notes.

About research: yes, I did some. This pattern is usually called a cash flow calendar, and it often exists as a separate app or feature focused on forecasting balances over time. I had built something similar for myself before, and since I eventually stopped using that in favor of Securo, I thought bringing the useful part of that workflow into Securo would make sense.

On the UX points, I have a few thoughts worth discussing:

> 2. The navigation controls are duplicated
> 4. The in cell add buttons feel redundant

I agree this adds some visual noise. My thinking was that, in calendar mode, users may expect the calendar itself to feel self-contained: month navigation close to the calendar, and day actions available directly from the day they are interacting with.

That said, I see the concern about having more than one control for the same concept. One possible compromise could be to keep the page-level month/account controls as the source of truth, but make the calendar navigation behave more like quick jumps, for example, previous/next 4 weeks, instead of duplicating the same month selector. That could also be useful later in the transaction list, but maybe that’s a separate discussion.

For the in-cell actions, I still think they can be intuitive when hovering over a specific day, but I’m open to removing them if we want the cell to stay focused on information only and keep creation actions in the side panel. (Maybe turning it into a right-click menu? Just giving ideas)

> 3. The day cells could carry a lot more

I’m a bit cautious about putting transaction rows directly inside every cell because the layout can become cramped quickly, especially on smaller screens or busy accounts. But I agree that the current cells may be too minimal.

Maybe we could make this configurable/persistent per user, something like a compact calendar view versus a detailed calendar view(Something similar to a toggle we are discussing on the assets). Compact would keep the current balance/marker-focused layout, while detailed could show the top 1–3 transactions plus “+N more.” What do you think?

> 5. Give the markers a legend, and make projected and low balance states louder

Agreed. A small legend and stronger visual treatment for projected/low-balance days make sense, especially because forecasting is the main value of this view.

--

I’ll start by fixing the correctness issue and the non-blocking bugs; then we can align on how far to take the UX changes in this PR versus a follow-up.

tassionoronha:
Thanks for taking it in consideration. 
Let me know when is ready for review again :)

RomuloGatto:
@tassionoronha I just assumed my counter-arguments were aligned with your thoughts and proceeded to fix it. Here is the updated short summary:

Related to the bug, the currency-scoped balance issue had already been addressed: calendar balances now use the shared daily balance delta helper scoped by account IDs and converted through the workspace primary currency. I also added regression coverage for a single foreign-currency account and for ignoring recurring projections from closed accounts.
    
On the UX side, I cleaned up the duplicated controls and reduced the calendar noise:
    - kept the page-level month/account controls as the source of truth
    - removed the duplicated calendar-card month/account controls
    - removed the permanent in-cell action buttons and made them hover/focus-only
    - added a persistent Compact/Detailed calendar toggle
    - Compact keeps the balance/marker-focused layout
    - Detailed shows the top transaction previews plus “+N more”
    - kept the legend compact
    - kept low-balance treatment visual via the red/rose cell styling
    - demoted projected from a grid-level marker/background to row/detail metadata, since it felt noisy as a fourth calendar marker

tassionoronha:
Screenshot 2026-07-13 at 19 40 55

Hey @RomuloGatto, worked on my side to bring some ideas I had to this page. 
Can you check if it looks fine for you?

RomuloGatto:
@tassionoronha ran tests here, on me, that look perfect. I miss the transactions shortcut in the day cells, but maybe we can revisit it after some community review. Feel free to merge when you want

PR #405, opened 2026-07-08
fix(a11y): add hover/accessible labels to icon-only buttons

rodrimaia:
## Problem

Many icon-only action buttons across the UI had no \`title\` or \`aria-label\`, so hovering showed nothing and users couldn't tell what an icon did. Reported for the connected-accounts screen (settings / sync / disconnect), but the issue was app-wide.

## Changes

Added accessible labels to **22 icon-only controls across 14 files**:

- **accounts**: connection Settings / Sync / Disconnect buttons (were fully unlabeled)
- **recurring, budgets, rules, goals**: edit / delete row actions
- **payees**: favorite toggle (add/remove favorite)
- **setup**: light / dark theme toggles
- **assets**: edit / delete asset, delete manual value, delete tx entry
- **agents-list** (view Link), **agent-detail** (back), **agent-connections** (edit/delete)
- **group-detail** (delete settlement), **admin/settings** (delete user)
- **knowledge-section** (delete doc), **transaction-drill-down** (close)

Also fixed a bug in \`assets.tsx\`: the edit/delete buttons' \`title\` was conditional and resolved to \`undefined\` exactly when the button was actionable.

### i18n

- Added \`common.view\`, \`payees.addFavorite\`, \`payees.removeFavorite\` to all 8 locales.
- Everything else reuses existing keys (\`common.edit/delete/close/back\`, \`accounts.sync/disconnect\`, \`connections.settings\`, \`settings.themeLight/Dark\`, \`assets.syncedReadOnly\`).

Approach kept simple per request: native \`title\` (+ \`aria-label\` on custom \`<Button>\`/\`<Link>\`). Note native tooltips keep the browser's ~1s hover delay.

## Follow-up (not in this PR)

Consider adding **\`eslint-plugin-jsx-a11y\`** to prevent regressions:
- \`jsx-a11y/control-has-associated-label\` (icon-only buttons)
- \`jsx-a11y/anchor-has-content\` (icon-only links)
- configure \`controlComponents: ['Button']\` since the shadcn \`<Button>\` isn't recognized by default
- prefer \`aria-label\` over \`title\` to satisfy the rule (title isn't reliably counted as an accessible name)

## Verification

- \`npm run build\` passes (includes tsc).
- Manual: hover each affected icon shows a translated label.


rodrimaia:
**Question:** Do guys think it is worth it to add the mentioned ESLint rules to avoid this issue globally?

tassionoronha:
Hey @rodrimaia , the PR LGTM! Thanks for that.

And regarding the lint rules, IMO, worth it. Be free to do that in a second PR :)

PR #416, opened 2026-07-14
fix: support CAMT.052 intraday reports in CAMT importer

Hoovercraft:
## Problem

`parse_camt()` only navigates `Document > BkToCstmrStmt > Stmt > Ntry`, which is the CAMT.053 (end-of-day statement) structure. A valid CAMT.052 file (intraday report, root path `Document > BkToCstmrAcctRpt > Rpt > Ntry`) parses without error but silently yields zero transactions, since the `Ntry` search never finds a matching `Stmt` container to look inside.

There's no error surfaced anywhere - the XML is well-formed, `ET.fromstring` succeeds, the loop over statements is just empty. From the UI this looks identical to an empty account: 0 transactions found, nothing to import.

Several European banks only issue CAMT.052, not CAMT.053 - in my case a German Volksbank (Genossenschaftsbank), where CAMT.052 is the only export format currently offered through online banking.

## Fix

Add `BkToCstmrAcctRpt/Rpt` as a fallback path when `BkToCstmrStmt/Stmt` yields no statements. Both CAMT.052 and CAMT.053 share the same `Ntry` sub-schema (amount, credit/debit indicator, booking/value date, remittance info), so the existing per-entry parsing logic needed no changes - only the statement-container lookup.

CAMT.053 behavior is unchanged (the fallback only triggers when the primary path finds nothing).

## Testing

- Added `test_parse_camt052_bktocstmracctrpt` to `TestParseCamt`, mirroring the existing CAMT.053 fixture pattern.
- Verified against real CAMT.052 XML exports from a German bank (self-hosted test instance) - transactions now show up in the import preview as expected; previously 0.

Happy to adjust naming/structure if you'd prefer this handled differently (e.g. a dedicated `parse_camt052` vs. a shared function).

### DIFF
--- a/backend/app/services/import_service.py
+++ b/backend/app/services/import_service.py
@@ -226,8 +226,10 @@ def parse_camt(content: bytes) -> list[TransactionImport]:
             # reported again as BOOK once it settles. Skip anything that
             # isn't BOOK to avoid importing it twice. Status is either a
             # plain code (<Sts>BOOK</Sts>) or wrapped (<Sts><Cd>BOOK</Cd></Sts>)
-            # depending on the schema version.
-            status = find_text(ntry, 'Sts') or find_text(ntry, 'Sts/Cd')
+            # depending on the schema version. Check the wrapped form first:
+            # in pretty-printed XML the <Sts> element's own text is the
+            # whitespace before <Cd>, which would otherwise mask the real code.
+            status = find_text(ntry, 'Sts/Cd') or find_text(ntry, 'Sts')
             if status and status.strip().upper() != 'BOOK':
                 continue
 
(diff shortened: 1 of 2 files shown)


tassionoronha:
Nice catch @Hoovercraft . 

One thing before merge: CAMT.052 includes pending entries (Sts=PDNG). 
Can you skip non-BOOK entries? 
Otherwise dupes later.

Hoovercraft:
Good catch, thanks - pushed a fix in 979c933.

`parse_camt()` now skips any `Ntry` whose `Sts` isn't `BOOK` (handles both the plain `<Sts>BOOK</Sts>` and the newer `<Sts><Cd>BOOK</Cd></Sts>` schema variants). Verified against a real CAMT.052 export that had a PDNG entry followed by the same transaction booked a day later - only the BOOK one is imported now. Added `test_parse_camt052_skips_pending_entries` to cover it.

tassionoronha:
Thanks @Hoovercraft! 
Just fixed a small point otherwise LGTM!

PR #419, opened 2026-07-14
fix(passkeys): bind WebAuthn to the origin the browser is on

### LINKED ISSUE #403: [Bug] Passkey registration fails when RP ID is set to a raw IP address
## Description

Adding a passkey fails with "Could not add passkey." The `/api/auth/passkeys/register/options` endpoint returns WebAuthn creation options where `rp.id` is set to a raw IP address (e.g. `192.168.4.xxx`) instead of the domain the app is served from.

The browser rejects the credential creation because, per the WebAuthn spec, the RP ID must be a [registrable domain suffix of, or equal to, the origin's effective domain](https://www.w3.org/TR/webauthn-2/#relying-party-identifier). A bare IP address is never a valid RP ID, so `navigator.credentials.create()` throws a `SecurityError` and no passkey is stored.

The RP ID appears to be derived from the server's internal bind address rather than the request's host/origin.

## To Reproduce

1. Deploy Securo behind a reverse proxy so it's reached via a domain (e.g. `securo.example.com`) while the backend binds to an internal IP.
2. Go to Settings → Passkeys.
3. Enter a passkey name and click **Add passkey**.
4. See error: "Could not add passkey."

## Expected behavior

The passkey should register successfully. `rp.id` should be set to the domain the user is accessing the app from (or be configurable), not the backend's internal IP address.

## Screenshots

Image

## Response from `/api/auth/passkeys/register/options`

```json
{
    "challenge_id": "<redacted>",
    "options": {
        "rp": {
            "name": "Securo",
            "id": "192.168.4.xxx"
        },
        "user": {
            "id": "<redacted>",
            "name": "<redacted-email>",
            "displayName": "<redacted-email>"
        },
        "challenge": "<redacted>",
        "pubKeyCredParams": [
            { "type": "public-key", "alg": -7 },
            { "type": "public-key", "alg": -257 }
        ],
        "timeout": 60000,
        "excludeCredentials": [],
        "authenticatorSelection": {
            "residentKey": "preferred",
            "requireResidentKey": false,
            "userVerification": "required"
        },
        "attestation": "none"
    }
}
```

The problem field is `options.rp.id`, it's an IP, but the app is accessed via a domain, so the origin's effective domain never matches it.

## Environment

- **Deployment:** Self-hosted (Docker), reverse proxy in front, backend bound to internal IP
- **Access URL:** custom domain (HTTPS)
- **Browser:** Chrome (latest)
- **OS:** macOS

## Additional context

Likely fix: derive `rp.id` from the request `Origin`/`Host` header (or expose an explicit `WEBAUTHN_RP_ID` / `RP_ID` config option) instead of using the server's bind address. The corresponding WebAuthn `origin` used during verification should also match the domain the browser sends.

### LINKED ISSUE #410: [Bug] Could not add passkey
## Description

A clear description of what the bug is.

## Steps to Reproduce

1. Go to 'Passkeys'
2. Click on 'Add passkey'
3. See error "Could not add passkey"

## Expected Behavior

I expected to see the passkey popup to register my faceid on iphone or the fingerprint prompt for passkey on macbook

## Actual Behavior

It just showed the error "Could not add passkey"

## Environment

- Deployment: Docker
- Browser: Firefox 152.0.5 (64-bit)
- OS: macOS 27, iOS 27 Dev Beta 3

## Screenshots

If applicable, add screenshots to help explain the problem.

Image

Image

tassionoronha:
Closes #403. Closes #410 (same root cause).

## The bug

`rp.id` was derived from `FRONTEND_URL`, so it never tracked the address the app was actually opened on. On the stock Docker stack at `http://127.0.0.1:3000` the backend returned `rp.id: "localhost"`, and the browser rejected it before any prompt appeared:

```
SecurityError: This is an invalid domain.
```

Passkeys only worked when browsing at exactly `http://localhost:3000`. A LAN IP, `127.0.0.1`, or a reverse-proxied domain all failed, and the UI reported every failure as "Could not add passkey". `WEBAUTHN_RP_ID` and `WEBAUTHN_ORIGIN` existed in `config.py` but were never passed through Compose, so Docker users had no way to set them.

Setting `FRONTEND_URL` to the real domain was not enough either: behind a TLS-terminating proxy, the browser's origin (`https://securo.example.com`) never matches a single internally-configured URL, so verification rejected the credential.

## The fix

- Resolve the RP ID and expected origin **per request**, from the origin the browser is on (`Origin`, falling back to the forwarded host). Both are stored in the challenge, so verification matches the ceremony that issued it.
- `WEBAUTHN_RP_ID` still pins passkeys to one domain, and now actually reaches the backend through Compose. `FRONTEND_URL` is overridable again.
- The frontend container now learns the real `FRONTEND_URL` too. It feeds Vite's `allowedHosts`, so a proxied domain was previously refused with "Blocked request" before passkeys were even reachable.
- Two WebAuthn rules can never be worked around: an IP address is not a valid RP ID, and plain HTTP is not a secure context off localhost. Both are now detected up front and explained, in the dialog and on the login page, instead of surfacing as an opaque error.
- Documented passkeys in `.env.example` and the README, where they were absent entirely.

## UX

The dialog now matches the rest of the app: inline delete confirmation instead of `window.confirm` (it was the only one in the frontend), toasts instead of raw red text, a proper empty state, and a distinct message per failure (cancelled, already registered, wrong domain, insecure origin, IP address).

## Tests

10 new backend tests covering RP-ID resolution, which had no coverage: origin derivation, `http://localhost`, IP and loopback-IP rejection, insecure origins, configured-RP-ID subdomain matching and mismatch rejection, forwarded-host fallback, and that verification uses the origin the challenge was issued for.

Exercised end to end in a real browser with a CDP virtual authenticator (real `navigator.credentials.create()`/`get()`, real `py_webauthn` signature and origin verification, real DB rows), in two setups:

- **`http://localhost:3000`** — register, sign in with the passkey and no password, delete. All pass.
- **HTTPS on a proxied domain**, behind a TLS-terminating proxy, which is the #403 setup — all pass.

As a control, the same run against the pre-fix backend fails to register and fails to sign in, on both setups. The IP origin (`127.0.0.1`) is correctly blocked up front with an explanation instead of a silent `SecurityError`.


tassionoronha:
Can you @DannyFeliz and @chisato04, check if this PR resolves the problem?

DannyFeliz:
@tassionoronha It works perfectly, I was able to create it and use it 👌🏻 

CleanShot 2026-07-16 at 23 43 34@2x

chisato04:
lgtm!

tassionoronha:
Thanks @chisato04 and @DannyFeliz for checking!

PR #412, opened 2026-07-13
fix(nginx): resolve 508 loop on cloud and add cloud deployment guide

dionisiojr:
## What

This PR introduces an optimized cloud deployment path alongside critical Nginx routing fixes to support zero-cost hosting on platforms like Render and Supabase, without breaking local Docker environments.

### 📂 Files Changed / Added:
1. `backend/start.sh` *(Added)*: A lifecycle script that automates database infrastructure on deploy, running `alembic upgrade head` before booting the Uvicorn server (with logs and comments internationalized to English).
2. `frontend/default.conf.template` *(Added/Modified)*: Replaced the static Nginx configuration with a dynamic environment-driven routing template.
3. `docker-compose.yml` *(Modified)*: Injected `NGINX_RESOLVER: 127.0.0.11` into the frontend container to guarantee zero friction for local development.
4. `CLOUDHOST.md` *(Added)*: A complete, frictionless guide for hosting the entire stack 100% for free on Render and Supabase.
5. `README.md` *(Modified)*: Added a quick-access link on the homepage pointing users to the new cloud deployment documentation.

## Why

When deploying the application to multi-tenant cloud providers (like Render), the static Nginx configuration injected `proxy_set_header Host $host;`, causing an infinite redirection loop (**Error 508 Loop Detected**). Additionally, hardcoded DNS resolvers broke Nginx's ability to locate the dynamic upstream backend container on cloud networks. 

These changes are needed to allow users to host the application in the cloud easily and completely for free, ensuring the project is highly accessible.

## How to Test

Steps to verify the changes work:

1. **Local Regression Test:** Run `docker-compose up --build` in your local environment. Verify that the frontend and backend boot normally and the application functions with zero friction, proving backward compatibility.
2. **Nginx Verification:** Inspect the generated Nginx config or verify that the container correctly utilizes the injected `NGINX_RESOLVER` (127.0.0.11) locally.
3. **Cloud Deployment Test (Optional):** Follow the steps provided in `CLOUDHOST.md` to deploy the stack on Render. Verify the application bypasses the 508 Loop Detected error and establishes a secure connection to the Supabase PostgreSQL instance.

## Related Issue

Addresses cloud deployment and Nginx routing limitations for PaaS providers.

## Checklist

- [x] Translations updated (if user-facing strings changed) -> *Documentation and script comments internationalized to English.*
- [x] For a large or core change, I discussed it first -> *Adding a standalone cloud deployment guide to expand project adoption.*
[...]

PR #421, opened 2026-07-15
Fix nginx paas loop #412

dionisiojr:
I wanted to thank you so much for the recommendations and feedback on the previous PR.
I just opened the new Pull Request. This implementation is essential for making Securo compatible with PaaS environments. It ensures the project runs natively and seamlessly in the cloud (even on free tiers) while keeping the local development environment completely untouched.

## What

This PR introduces Nginx dynamic routing templates and a fallback startup script to support zero-cost cloud hosting (PaaS) without breaking local environments.

### 📂 Files Changed / Added:
1. `frontend/Dockerfile` *(Modified)*: Added `ENV NGINX_RESOLVER=127.0.0.11` as a fallback with inline comments explaining the PaaS upstream resolution fix.
2. `frontend/default.conf.template` *(Modified)*: Dynamic template for Nginx routing, resolving the 508 Loop and fixing SNI handshakes.
3. `backend/start.sh` *(Added)*: A startup script to apply Alembic migrations sequentially before starting Uvicorn, which is strictly necessary for free-tier PaaS environments.

## Why

Multi-tenant PaaS providers like Render trigger a **508 Loop Detected** error with static Nginx configs due to `proxy_set_header Host $host;`. Furthermore, hardcoded DNS resolvers prevent containers from communicating correctly over dynamic cloud networks.

Since free-tier PaaS environments do not support automated "Pre-Deploy Hooks" (which are typically locked behind paid tiers), having a physical `start.sh` script is required to run database migrations sequentially before starting the FastAPI/Uvicorn server.

## How to Test

1. **Local Boot Check:** Run `docker-compose up --build` on your local machine. Verify that the application starts perfectly without specifying `NGINX_RESOLVER`, verifying that the `127.0.0.11` fallback works.
2. **Dynamic Resolve Check:** Confirm Nginx dynamically resolves backend upstreams without configuration failures.
3. **Cloud Run:** Successfully tested both locally and deployed in a cloud PaaS environment.

## Related Issue

Addresses Nginx routing loops and database migration automation for cloud PaaS, building upon the feedback received on PR #412.

## Checklist

- [x] Translations updated (if user-facing strings changed) -> *All comments, environment fallbacks, and scripts are internationalized to English.*
- [x] For a large or core change, I discussed it first -> *Aligned directly with the maintainer's feedback on PR #412.*

Adds PaaS startup support and dynamic Nginx routing. Adds a Render and Supabase deployment guide.

- Runs database migrations before starting the backend.
- Preserves shutdown signals for Uvicorn.
- Supports dynamic proxy routing, DNS resolution, and upstream TLS verification.
- Documents database, Redis, secrets, architecture, and deployment limits.

Risk: database migration.


tassionoronha:
Nice diagnosis, and the resolver fallback works (tested both prod images). 

One ask: dropping `Host $host` breaks redirects for existing deploys (they point at `backend:8000`). Could you template it instead, `ENV PROXY_HOST=\$host` (escape needed)? Also `start.sh` isn't wired up anywhere and isn't +x.

dionisiojr:
Thanks for the feedback! I have updated the branch with all requested adjustments:

1. **Templated `PROXY_HOST`:** Restored `proxy_set_header Host ${PROXY_HOST};` in `frontend/default.conf.template` and added `ENV PROXY_HOST=\$host` fallback in `frontend/Dockerfile`. This preserves default `$host` behavior for existing deployments while allowing PaaS/cloud platforms to override it via environment variables if needed.
2. **Permissions & Script Wiring:** Granted `+x` execution permission to `backend/start.sh` and updated `backend/Dockerfile` with `RUN chmod +x ./start.sh` so it is fully ready for startup execution.

Ready for review!

tassionoronha:
Thank you for coming back with this, and for taking the #412 feedback on board so carefully. Making Securo easy to self-host on a free PaaS tier is genuinely valuable, and I appreciate the legwork. 🙏

I pulled the branch and tested it: built the prod frontend image and ran it with only `BACKEND_URL` set, the way `docker-compose.prod.yml` does. Rendered config comes out identical to today (`resolver 127.0.0.11`, `Host $host`), `/api/health` returns 200, and the Helm chart is untouched since it ships its own template. The ENV-default approach works exactly as intended, nice call.

A few things before merge:

**1. `start.sh` needs `exec`.** Without it uvicorn is a child of bash and never gets SIGTERM. I reproduced it: `docker stop` hangs the full 10s grace period, then SIGKILL (exit 137), so in-flight requests get cut on every deploy.

```bash
exec uvicorn app.main:app --host 0.0.0.0 --port 8000
```

**2. The guide has your live hostnames in it.** `securo-oic9.onrender.com` and `securo-1-iy2i.onrender.com` are real instances of yours, could you swap them for placeholders? Same for `SECRET_KEY`, worth telling people to generate one with `openssl rand -hex 32` since it signs JWTs.

**3. Gaps in the guide:** the "Deployment Architecture" section is empty; there's no step for provisioning PostgreSQL even though `DATABASE_URL` is required; and there's no Celery worker or beat. Without those last two, bank sync, recurring transactions, FX sync and asset valuations silently never run. Either add them or state explicitly that they're skipped and what that costs.

**4. Smaller, take or leave:** move the guide to `docs/deploy-render.md` to match the other docs; `NGINX_RESOLVER=8.8.8.8` as the documented default sits awkwardly with the project's privacy pitch (fine as an option, just not the recommendation); `RUN chmod +x ./start.sh` is redundant since the file is already committed as `100755`; a few trailing spaces in the template and Dockerfile comments. Optional hardening: `proxy_ssl_verify on;` + `proxy_ssl_trusted_certificate /etc/ssl/certs/ca-certificates.crt;` next to your SNI line, now that the proxy can reach an upstream over the public internet. I checked, that CA bundle ships in the nginx unprivileged alpine image and the handshake works.

One heads-up: `docker compose up --build` doesn't actually exercise this change, local dev builds the frontend from `Dockerfile.dev` (Vite), not the nginx `Dockerfile`. `docker build ./frontend` or `docker-compose.prod.yml` is the real check.

Solid work, the backward-compatible design here is the right instinct. Ping me once the above is in and I'll merge. 🚀

dionisiojr:
Hi @tassionoronha ! Thanks again for the thorough review, feedback, and testing guidance! 🙏

I have updated the branch with all the requested refinements:

1. **`start.sh` & SIGTERM Handling:** Added `exec` before `uvicorn app.main:app --host 0.0.0.0 --port 8000` so that SIGTERM signals are properly passed to Uvicorn, eliminating deployment timeouts during container shutdown.
2. **Guide Relocation & Security Cleanup:**
   - Moved the deployment guide to `docs/deploy-render.md` and cleaned up the legacy file.
   - Replaced all live hostnames with generic placeholders (`your-backend-service.onrender.com`). 👍 Thanks for tip
   - Added instructions to generate `SECRET_KEY` via `openssl rand -hex 32`.
   - Updated the default `NGINX_RESOLVER` recommendation to `1.1.1.1` to better align with the project's privacy focus.
3. **Gaps in the guide:**
   - Added a clear **Mermaid diagram** for the Deployment Architecture section.
   - Detailed the step-by-step setup for **Supabase (PostgreSQL)**, including Session Pooler setup and the `+asyncpg` driver formatting.
   - Added an explicit note regarding Render's Free Tier 15-minute HTTP spin-down behavior, explaining why background workers are omitted and detailing the background features affected.
5. **Smaller, take or leave:**
   - Removed the redundant `RUN chmod +x ./start.sh` line from `backend/Dockerfile`.
   - Applied the recommended Nginx upstream SSL hardening (`proxy_ssl_verify on;`, `proxy_ssl_trusted_certificate`, and `proxy_ssl_name`).
   - Cleaned up trailing whitespace across the modified files.

Everything is tested and deployed successfully on Render. Ready for merge whenever you can! 🚀

dionisiojr:
Hi @tassionoronha, Great to see all checks passed and green. Just following up to see if everything looks good on your end or if any further adjustments are needed before merge. Thanks! 🚀

PR #422, opened 2026-07-15
fix(a11y): enforce accessible labels via jsx-a11y lint rules

rodrimaia:
## What

Adding an eslint rule to prevent issues like https://github.com/securo-finance/securo/pull/405 :) 

Add eslint-plugin-jsx-a11y with control-has-associated-label (depth: 3) and anchor-has-content rules to catch missing accessible labels at lint time. Fix remaining icon-only buttons missing aria-label: user avatar menu (dedicated common.userMenu key), admin toggles, reports pagination, import undo, goals color picker, and edit/delete buttons in budgets and recurring. Add common.userMenu translation across all 8 locales.

## Why

Continuation of #405 -- icon-only controls without accessible names are invisible to screen readers and fail WCAG. The lint rules prevent regressions; depth: 3 handles controls whose visible label is nested deeper than the default depth: 2.

## How to Test

1. Run `npm run lint` -- zero jsx-a11y errors
2. Inspect any icon-only button with DevTools -- every control now has an accessible name
3. Confirm `npm run build` passes

## Related Issue

Follow-up to #405

## Checklist


tassionoronha:
Thanks for pushing on this subject @rodrimaia !
Added some fix but is good overall :)

PR #432, opened 2026-07-16
fix(dashboard): keep excluded rows out of category drill-down total

Lucas-Euzebio:
## What

The dashboard category drill-down side panel summed **every** returned transaction into its subtotal, including rows that don't count toward P&L (ignored transactions and settlement legs). As a result the panel's total didn't match the category total that opened it.

This shows those rows for transparency — greyed, struck-through, with an "ignored" badge — but keeps them **out of the total**, and adds an "Excluded" line reporting their sum. It mirrors the treatment the main Transactions list already uses (`summaryExcluded`).

## Why

The dashboard category figure uses `counts_as_user_pnl()` (excludes ignored / settlement / transfer-category rows), but the drill-down called the generic transactions list — which intentionally returns ignored rows — and summed them client-side. So clicking a category that contained any ignored transaction opened a panel whose total contradicted the number just clicked.

## How to Test

1. In a given month, mark one transaction of a spending category as ignored (row menu → ignore).
2. On the dashboard, note that category's total (it excludes the ignored one).
3. Click the category to open the drill-down side panel.
   - **Before:** the ignored transaction was listed and added into the panel total, so it diverged from the category figure.
   - **After:** it's listed but greyed / struck-through with an "ignored" badge, an "Excluded" line reports its sum, and the panel total matches the category figure.

## Related Issue

None — small self-contained bug fix.

## Checklist

- [x] Backend tests pass (`pytest`) — no backend changes
- [x] Frontend lints clean (`npm run lint`) — no new warnings vs. baseline
- [x] Translations updated (all 8 locales)
- [ ] For a large or core change, I discussed it first — N/A (small bug fix)

### DIFF
--- a/frontend/src/components/transaction-drill-down.tsx
+++ b/frontend/src/components/transaction-drill-down.tsx
@@ -3,7 +3,7 @@ import { useTranslation } from 'react-i18next'
 import { useDisplayLocale, useDateLocale } from '@/hooks/use-display-locale'
 import { useQuery } from '@tanstack/react-query'
 import { transactions as transactionsApi, dashboard, admin } from '@/lib/api'
-import { AlertTriangle, Info, Paperclip, X } from 'lucide-react'
+import { AlertTriangle, EyeClosed, Info, Paperclip, X } from 'lucide-react'
 import { CategoryIcon } from '@/components/category-icon'
 import { useAuth } from '@/contexts/auth-context'
 import { usePrivacyMode } from '@/hooks/use-privacy-mode'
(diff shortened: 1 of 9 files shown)


tassionoronha:
Hey @Lucas-Euzebio, thanks for this PR. 

Althought, this was already fixed on main by #379, which moved this server-side (user_pnl_only). 
Verified the drill-down total now matches. Closing this one here, but be free to open another one if you think that the current implementation could be improved regarding this.

Cheers :)

PR #434, opened 2026-07-17
Add OIDC-only local auth toggle

### LINKED ISSUE #433: [Feature] Setting to require OIDC-only login (disable local password/passkey auth)
## Problem

Securo supports OIDC login (great!) alongside local username/password and passkey auth, but there's no way to require OIDC exclusively once it's set up. The `/api/auth/login` (password) and passkey endpoints are unconditionally mounted with no config option, so anyone who knows (or guesses) a local account's password can still bypass SSO entirely, even on an instance where the admin wants OIDC (such as Authentik) to be the **only** way in (e.g. for centralized user management, deprovisioning, or MFA policy enforcement at the IdP).

`registration_enabled` already solves half of this (blocking *new* local signups), but it doesn't touch login for accounts that already exist, including the original admin account created during `/setup`.

## Proposed Solution

A new setting, e.g. `LOCAL_AUTH_ENABLED` (default `true` for backward compatibility), that when set to `false`:
- Returns 403 from `POST /api/auth/login` (password grant) regardless of credentials
- Returns 403 from the passkey authenticate/verify endpoints
- Frontend hides the password/passkey login form entirely when OIDC is the only available method (can be surfaced via the existing `/api/auth/oidc/config` response, e.g. an added `local_auth_enabled` field)

This would only apply when `OIDC_ENABLED=true`, since disabling all auth methods with no OIDC configured would lock everyone out — probably worth a startup validation error in that case, similar to the existing `OIDC_EXISTING_USER_LINK_MODE` validation in `oidc_auth.py`.

## Alternatives Considered

- **Reverse-proxy level blocking** (e.g. blocking `POST /api/auth/login` at Traefik/nginx in front of Securo) — works, but it's a blunt network-layer workaround rather than something the app itself understands, and it's easy to forget/misconfigure or leave a gap (e.g. passkey endpoints) since it's not documented or tested as part of Securo itself.
- **Leaving it as-is** — acceptable for most instances, but not for anyone standardizing on SSO-only access for compliance/security reasons.

## Additional Context

Happy to take a first pass at implementing this if the idea aligns with where auth is headed — the codebase already has a clean precedent to follow in `registration_enabled` (`admin_service.is_registration_enabled`, `check_registration_enabled` dependency in `main.py`), and the OIDC settings/validation live in `core/config.py` + `api/oidc_auth.py`. Would want to confirm naming/scope before opening a PR per the contributing guide.

RomuloGatto:
## What

Adds a `LOCAL_AUTH_ENABLED` setting so admins can require OIDC-only login after configuring an OIDC provider.

Fixes #433.

## Changes

- Adds `LOCAL_AUTH_ENABLED=true` by default for backward compatibility.
- Validates that `LOCAL_AUTH_ENABLED=false` can only be used when `OIDC_ENABLED=true`, preventing accidental lockout with every login method disabled.
- Returns `403 LOCAL_AUTH_DISABLED` from password login when local auth is disabled.
- Returns `403 LOCAL_AUTH_DISABLED` from passkey authentication and passkey second-factor endpoints when local auth is disabled.
- Extends `/api/auth/oidc/config` with `local_auth_enabled` so the frontend can reflect the configured login modes.
- Hides email/password, passkey login, and registration controls from the login page when OIDC-only mode is active.
- Documents/wires the setting in `.env.example`, backend env example, Docker Compose, production Compose, Helm values, and README.

## Verification

- `cd backend && uv run --extra dev pytest tests/test_config.py tests/test_auth_api.py::test_login_forbidden_when_local_auth_disabled tests/test_oidc_auth.py::test_oidc_config_disabled_by_default tests/test_oidc_auth.py::test_oidc_config_exposes_local_auth_disabled tests/test_passkeys.py::test_passkey_login_forbidden_when_local_auth_disabled tests/test_passkeys.py::test_passkey_second_factor_forbidden_when_local_auth_disabled -q`
- `cd backend && uv run --extra dev pytest tests/test_auth_api.py tests/test_passkeys.py tests/test_oidc_auth.py tests/test_config.py -q`
- `cd backend && uv run --extra dev ruff check .`
- `cd frontend && npm run lint` — passes with existing warnings only
- `cd frontend && npm test -- --run`
- `git diff --check`
- Browser smoke test with mocked `/api/auth/oidc/config` confirmed OIDC-only mode renders only the provider button (`Continue with Pocket ID`) and hides local login/passkey/register controls.


tassionoronha:
Hey @RomuloGatto, thanks for this! Really appreciate that you wired it through config, compose, Helm values, both `.env.example` files and the README. The `model_validator` guard was the right instinct too.

Two things:

**1. The guard doesn't close the hole it was written for.**

It checks `oidc_enabled`, but `/api/auth/oidc/config` computes `enabled` as `oidc_enabled AND oidc_client_id AND oidc_discovery_url`. So `OIDC_ENABLED=true` + `LOCAL_AUTH_ENABLED=false` with a missing client id or discovery URL passes validation, boots cleanly, and renders a login page with zero login methods. I brought that state up locally: logo, heading, subtitle, nothing else. Total lockout on a self-hosted box, and an easy variable to forget.

The README specifically promises this can't happen. Fix is small: have the validator require the same conditions the config endpoint does, and give the login page an explicit "no auth method configured" state so the failure is legible.

**2. OIDC-only doesn't cover register or password reset.**

Verified against a running instance with local auth off: `/api/auth/login` correctly 403s, but `POST /api/auth/register` returns 201 and `POST /api/auth/forgot-password` returns 202.

Not directly exploitable, and credit where due, the OIDC linking logic is careful and rejects email-matched local accounts by default. But it does undercut `OIDC_AUTO_REGISTER=false` + `OIDC_EXISTING_USER_LINK_MODE=email`, where an open register endpoint lets anyone create the account the IdP identity then links to. I'd gate both routers on the same flag and spell out in the README what the setting does and doesn't cover.

**Smaller:**

- `PATCH /api/users/me` still accepts `password` via `BaseUserUpdate`.
- `app-layout.tsx` defaults `localAuthEnabled` to `true` while the query is in flight, so the password/2FA/passkey items flash in on load. `undefined` reading as "enabled" is the less safe default here.
- The `oidc_only_settings` fixture mutates the cached settings singleton across every key. The narrower one in `test_passkeys.py` is the safer pattern.
- #469 and #477 both add a migration `066`, just flagging across branches.

Drove it in Chrome and the happy path is genuinely nice: single primary "Continue with Pocket ID" button, no stray divider, and the user menu correctly drops password/2FA/passkeys. That part I'd ship as-is.

RomuloGatto:
@tassionoronha tks for the points, just fixed:

1. Lockout prevention
    - Config validation now uses the same availability predicate as /api/auth/oidc/config.
    - LOCAL_AUTH_ENABLED=false requires OIDC_ENABLED=true, OIDC_CLIENT_ID, and OIDC_DISCOVERY_URL.
    - The login page now has explicit loading, config-fetch error, and “no authentication method configured” states instead of rendering an empty card.
    
2. Registration and password reset
    - Added the shared local-auth guard to the registration router.
    - Added it to both forgot-password and reset-password endpoints.
    - Updated the README and environment examples to document exactly which paths are disabled and which data remains intact.
    
Smaller items
    - Password mutations through PATCH /api/users/me and the other FastAPI Users update path now return 403 LOCAL_AUTH_DISABLED; ordinary profile updates remain available.
    - app-layout.tsx now treats an unresolved OIDC config as local auth disabled, preventing the password/2FA/passkey menu flash.
    - Narrowed oidc_only_settings so it restores only the four settings it changes.
    - Centralized the local-auth guard shared by password and passkey routes.
    
I also added regression coverage for incomplete OIDC configuration, registration, forgot/reset password, password updates, and allowed non-password profile updates.

tassionoronha:
@RomuloGatto great round, thanks for turning it around so fast. All four points are properly closed: sharing the `oidc_login_available` predicate between the validator and the config endpoint was exactly the right move. I verified live that register, forgot/reset-password, `PATCH /users/me` password and the passkey routes all 403 now, while ordinary profile updates keep working. And the OIDC-only login card looks great:

OIDC-only login card

Drove the branch in Chrome and found one blocker plus a few smaller things:

**1. Login is impossible if `/api/auth/oidc/config` alone fails.** I failed just that one request, backend healthy, `LOCAL_AUTH_ENABLED=true`: no email/password fields, no submit button, no passkey button, no register link.

Login page with only the oidc/config request failing

Before this PR that call was swallowed by `.catch(() => {})` and login kept working. The backend already 403s local login when the flag is off, so the client can safely fall back to `local_auth_enabled: true` on error and show a non-blocking warning. Fail closed on the server, fail open on the client.

**2. Same fail-closed pattern in `app-layout.tsx`.** `oidcConfig?.local_auth_enabled === true` means a failed or still-retrying query permanently hides Change password / 2FA / Passkeys. `?? isError` gets you what I was asking for in the last round: hidden while pending, shown when the config is unknowable.

**3. `POST /api/setup/create-admin` is still open.** Unauthenticated, creates a superuser with a local password whenever the users table is empty. On a fresh OIDC-only deploy that is the same account-seeding path we discussed, and the README now says registration is disabled without carving it out. Either gate it (and document that the first admin comes from `OIDC_SYNC_ROLES` + `OIDC_ADMIN_ROLES`) or spell out the exception. Same wording issue for admin password sets, `admin_service.update_user` hashes directly and never hits the overridden `update`.

**4. TOTP enrollment is not gated while passkey enrollment is.** `POST /api/auth/2fa/setup` returns 200 in OIDC-only mode, `passkeys/register/options` returns 403. Same class of action, two behaviors.

**5. `/register` still renders the full signup form** if you hit the URL directly, and submitting just gives a generic error. A redirect to `/login` closes it.

Register page still rendering in OIDC-only mode

Nits: the login page now shows ~300ms of "Loading…" on every load (measured locally, longer over a network), worth a deliberate call. And `oidc_only_settings` / `local_auth_disabled` are near-duplicates, one fixture in `conftest.py` would do.

Only 1 is a blocker for me. Thanks again for the care you are putting into this one.

RomuloGatto:
@tassionoronha closed the remaining round.

- The login page now fails open only when /api/auth/oidc/config actually fails, showing local controls with a non-blocking warning while keeping the backend authoritative. Pending configuration still hides local controls to avoid an incorrect flash.
- app-layout.tsx follows the same pending/error semantics.
- Fresh setup, admin user creation, admin password updates, and password-backed workspace user creation are now blocked in OIDC-only mode.
- TOTP setup and enablement are blocked consistently with passkey enrollment; disabling existing TOTP remains available as a cleanup path.
- Direct /register and /setup access redirect when OIDC-only mode is known.
- A fresh OIDC-only instance now stays on the OIDC login flow instead of redirecting to local setup. The README documents first-admin provisioning through OIDC_AUTO_REGISTER, OIDC_SYNC_ROLES, and OIDC_ADMIN_ROLES.
- Local credential controls were also removed from admin and workspace-invite screens.
- The duplicate settings fixtures were consolidated in conftest.py.

The short loading state is deliberate: it prevents local credential controls from flashing on an OIDC-only instance. A genuine config-request failure now recovers to local login instead of blocking access.

tassionoronha:
@RomuloGatto this one is in good shape. Went through the last round live and it all holds up.

I pushed one commit instead of sending you back for another pass, since it was follow-through on my own notes:

- The local-credential screens only fell back on isError, and a failed config request does not reliably get there (the retry parks and the query stays pending), so Change password / 2FA / Passkeys stayed hidden. Pinned retry: false + networkMode: 'always' behind a shared useLocalAuthEnabled hook.
- Timeout on /auth/oidc/config, a hung request left the login card stuck loading.
- Login and passkey guards moved to route dependencies, so a bad body gets 403 LOCAL_AUTH_DISABLED instead of a 422.
- Inviting an unknown email in OIDC-only mode no longer asks for a password that cannot be given.

Shout if anything reads wrong, especially the hook extraction. 
Thanks for the patience across three rounds 💜

PR #439, opened 2026-07-18
fix: make category picker scrollable inside dialogs on touch/wheel input

### LINKED ISSUE #428: [Bug] on mobile web, selecting the category inside a rule is not scrolling down
## Description

Using securo with my cellphone. Created a rule,  at the actions, tried to select a category.  The drop-down is not responsive,  can't scroll down to select the category I need. 

## Environment

- Deployment: Docker
- Browser: brave
- OS: android

itodevio:
Fixes #428

## Why

Radix's body scroll lock (`react-remove-scroll`) registers a bubble-phase `wheel`/`touchmove` listener on `document` that calls `preventDefault()` on events whose target isn't inside the Dialog's own DOM subtree. The `Command` list used by combobox pickers (e.g. selecting a category on a rule's action) portals to `document.body`, so it falls outside that subtree, and its wheel/touch events get suppressed before the list itself can scroll — this leaves the picker frozen on touch devices (and non-scrollable via mouse wheel) when opened from within a Dialog.

React also attaches `onWheel`/`onTouchMove` as passive listeners, so calling `preventDefault()` from them isn't an option.

## How

`CommandList` now attaches its own non-passive native `wheel`/`touchstart`/`touchmove` listeners directly to the list's DOM node (via a ref) and drives `scrollTop` manually, unconditionally. This works the same whether or not a Dialog's scroll lock is currently active.

## Test plan

- [x] `npm run lint` — clean
- [x] `npm run build` — clean
- [x] Open the rule dialog's category picker on a touch device/emulator and confirm the list scrolls with a finger drag
- [x] Confirm mouse-wheel scrolling also works in the same picker
- [x] Repeat for the transaction dialog's category picker

### DIFF
--- a/frontend/src/components/ui/command.tsx
+++ b/frontend/src/components/ui/command.tsx
@@ -4,6 +4,55 @@ import { SearchIcon } from "lucide-react"
 
 import { cn } from "@/lib/utils"
 
+// When this list is nested inside a modal Dialog (e.g. category picker
+// inside the transaction dialog), Radix's body scroll lock (react-remove-scroll)
+// registers a bubble-phase `wheel`/`touchmove` listener on `document` that
+// preventDefault()s events whose target isn't inside the dialog's own DOM
+// subtree — which is exactly this list's situation, since its Popover portals
+// to document.body. That listener only fires once the event has bubbled all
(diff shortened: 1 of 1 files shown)


tassionoronha:
The PR fixs the problem.
Thanks @itodevio !

PR #440, opened 2026-07-19
fix(frontend): use local dates for form defaults

ADolkun:
## Summary

- format user-facing form defaults using the browser local calendar day
- share one date helper across transaction, transfer, account, asset, settlement, and recurring forms
- leave UTC download filename timestamps unchanged

## Why

`new Date().toISOString()` produces a UTC day, so users near midnight outside UTC can get yesterday or tomorrow in date inputs.

## Checks

- 37 frontend tests passed
- focused regression passed with outer UTC and Tokyo timezones
- frontend lint passed with existing warnings
- frontend production build passed
- independently reviewed and approved


tassionoronha:
LGTM! Thanks @ADolkun :)

PR #441, opened 2026-07-19
fix(auth): tolerate adjacent TOTP windows

ADolkun:
## Summary

- accept TOTP codes from the immediately previous or next 30-second window
- route enable, disable, and login verification through one helper
- rate-limit the second-factor verification endpoint
- cover adjacent-window behavior and the sixth-attempt 429 response

## Why

This tolerates normal authenticator/server clock drift without widening beyond one adjacent time step. Rate limiting keeps the larger acceptance window bounded against repeated guesses.

## Checks

- 14 targeted backend tests passed with RuntimeWarning promoted to errors
- Ruff passed on all touched files
- independently security-reviewed and approved


tassionoronha:
Great addition, LGTM!

PR #443, opened 2026-07-19
fix: enforce workspace payee uniqueness

ADolkun:
## What
- make payee uniqueness workspace-scoped, case-insensitive, and whitespace-normalized
- require sync/import callers to provide the workspace explicitly
- recover concurrent payee creation through a savepoint instead of aborting the sync/import transaction
- migrate existing normalized duplicates while preserving transactions, mappings, rules, favorites, and notes
- trim names consistently on create and update

## Why
Payees are workspace-owned, but the database still enforces the legacy user/name constraint. Concurrent syncs can also race between lookup and insert, producing a duplicate-key failure that aborts the whole sync.

## Verification
- full backend: 2293 passed, 7 skipped; 92.90% coverage
- focused payee/sync/import/API: 204 passed
- Ruff passed
- Alembic 065 -> 066 upgrade and 066 -> 065 downgrade SQL rendered cleanly
- independent review: caller-scope blocker fixed; no remaining blockers

## Stack

Review and merge in this order — each PR builds on the one above it:

1. #374 — connection sync correctness (base of the stack)
2. #443 — workspace payee uniqueness
3. #444 — sync retry reliability (merges last)

GitHub requires a PR's base branch to live in this repository, and these
branches live in a fork, so all three necessarily target `main` rather than
each other. #444 contains this PR's commits; they disappear from its diff once this one lands.

## Rebase note

Merged current `main` (2026-08-09). While resolving, an earlier merge on this
branch was found to have reverted the payee lookup to `lower(name)`, which no
longer matched the `lower(btrim(name))` unique index that migration 066
creates — a legacy untrimmed name would miss the lookup and hit an
IntegrityError. The trim-normalized, workspace-scoped lookup is restored.

Re-verified after the merge: Ruff clean, `ty check` clean, full backend suite
2411 passed / 7 skipped.


ADolkun:
Resolved against current `securo-finance/securo:main` at `5daf3b2` (2026-08-23). Upstream now contains the basic workspace-scoped constraint; this PR retains its additional case/whitespace-normalized uniqueness and race-safe get-or-create behavior. Rebased the data migration to `074` after current head `073`, preserving fiscal IDs, mappings, nested rule references, favorites, notes, and contact fields. Seeded PostgreSQL upgrade validation passed.

tassionoronha:
Thanks for this, the normalized uniqueness is the right invariant and the service side reads well. All three existing callers already pass `workspace_id`, so making it required is safe.

One hard blocker: the revision id collides. `main` already has `074_hide_default_categories` and `075_account_institution`, and this adds a second revision `074` pointing at `073`. After merging main, `alembic upgrade head` fails outright:

```
UserWarning: Revision 074 is present more than once
FAILED: Multiple head revisions are present for given argument 'head'
```

Renumbering it to `076` with `down_revision = "075"` clears the collision, and the existing chain then runs up to `075` cleanly.

Two things I couldn't validate and would want covered before merge:

- The migration body is ~150 lines of Postgres that never runs in CI, since the test DB is SQLite. Merging notes, hoisting tax ids, repointing transactions/mappings/rules, then deleting, is a lot to get right unattended. Could you describe what you ran it against, ideally with row counts before and after?
- The rules `conditions` rewrite does a `replace()` on the whole JSON as text, so any other field holding the same UUID string gets rewritten too. Narrowing that to the `payee_id` nodes you already match with `jsonb_path_exists` would be safer.

Smaller: `test_get_or_create_payee_returns_case_insensitive_concurrent_winner` mocks the session, so `begin_nested` is a mock and the savepoint behavior it's named after isn't actually exercised.

ADolkun:
Thanks — addressed in `0e64124` and merged current `main` first.

- Renamed/rechained the migration to `076` with `down_revision = "075"`; `alembic heads` now reports exactly `076 (head)`.
- Replaced the whole-document text substitution with a structure-aware rewrite. It changes only matching `field == "payee_id"` leaves, including the supported one-level nested condition groups; sibling fields containing the same UUID are preserved.
- Reworked the concurrent-winner test to use the real `AsyncSession`, nested transaction, conflicting insert, and winner query. The exercised SQL path is `SAVEPOINT` → failing `INSERT` → `ROLLBACK TO SAVEPOINT` → successful lookup.

For the migration validation, I used a disposable `pgvector/pgvector:pg16` database, ran the full Alembic chain to `075`, seeded three case/whitespace variants plus their dependent data, then ran `alembic upgrade head` to `076`.

| Row/check | Before | After |
|---|---:|---:|
| payees | 3 | 1 |
| payee tax IDs | 3 | 2 |
| transactions pointing at duplicate IDs | 2 | 0 |
| transactions pointing at keeper | 0 | 2 |
| payee mappings | 2 | 4 |
| rules | 1 | 1 |

Post-upgrade assertions confirmed:

- the oldest payee won;
- favorite, merged notes, type, email, phone, address, and website survived;
- the keeper's tax ID plus the earliest non-conflicting duplicate tax ID survived;
- transactions and mappings point to the keeper;
- `set_payee` actions and top-level/nested `payee_id` conditions were repointed;
- unrelated condition/action values containing the duplicate UUID were unchanged.

Local checks passed: Ruff, `ty`, 3,205 backend tests with 92% coverage, focused payee tests, `git diff --check`, and the seeded PostgreSQL migration assertions. An independent no-edit review found no blocking issues.

tassionoronha:
Re-checked `0e64124`. All three are addressed:

- `alembic heads` reports `076 (head)` and nothing else.
- The conditions rewrite is now structural, only touching leaves where `field == "payee_id"` matches the source id, and it handles nested groups. Sibling fields holding the same UUID are left alone.
- The concurrency test runs against a real `AsyncSession` and only stubs the first lookup, so the savepoint, the failing insert against the real unique index, and the recovery lookup are all exercised.

Full backend suite passes (3160), branch is current with `main`, CI green. Thanks for the migration validation write-up, that was what I was missing.

No further blockers from me.

tassionoronha:
Thanks for that @ADolkun !

PR #449, opened 2026-07-21
fix: SimpleFIN pending transactions no longer date to the Unix epoch

### LINKED ISSUE #447: [Bug] Pending SimpleFIN transactions get date 1970-01-01 instead of their real date
## Description

Pending transactions synced from SimpleFIN get stored with date `1970-01-01` (the Unix epoch) instead of their actual transaction date. This makes them invisible in any date-filtered view (e.g. viewing "this month"), even though the sync itself completes successfully and the transaction is genuinely there.

## Steps to Reproduce

1. Connect a bank/credit account via SimpleFIN that currently has pending (not-yet-posted) transactions.
2. Sync the connection.
3. Go to Transactions, filtered to the current month.
4. The pending transactions from that account don't appear.
5. Query the transaction directly (e.g. via the API or DB) — its `date`/`effective_date` is `1970-01-01`.

## Expected Behavior

Pending transactions should show their actual (pending) date, sourced from `transacted_at`.

## Actual Behavior

They get `1970-01-01`. In `app/providers/simplefin.py`, `_build_transaction()` does:

```python
posted = _epoch_to_date(raw.get("posted"))
transacted = _epoch_to_date(raw.get("transacted_at"))
txn_date = posted or transacted
```

SimpleFIN sends `"posted": 0` for transactions that haven't posted yet (there's no real posted timestamp). `_epoch_to_date(0)` returns `date(1970, 1, 1)` — not `None` — because it only treats `None`/`""` as "no value":

```python
def _epoch_to_date(value):
    if value is None or value == "":
        return None
    return datetime.fromtimestamp(int(value), tz=timezone.utc).date()
```

Since a real `date` object is truthy even when it's the epoch, `posted or transacted` picks the wrong date and never falls through to the correct `transacted_at` value sitting right next to it.

## Suggested Fix

Treat a falsy/zero `posted` as "no value" before falling back to `transacted_at`, e.g.:

```python
posted = _epoch_to_date(raw.get("posted")) if raw.get("posted") else None
```

(or make `_epoch_to_date` itself reject `0`, since it's never a meaningful timestamp for this API).

## Environment

- Deployment: Docker
- Version: v0.13.10

brandonjones24:
## Summary
- `_epoch_to_date(0)` legitimately returns `1970-01-01` (it's a valid conversion of a real epoch timestamp), but SimpleFIN sends `"posted": 0` — not `null` — specifically to mean *not posted yet*, not "posted at the Unix epoch."
- `posted = _epoch_to_date(raw.get("posted"))` was therefore resolving to a truthy `date(1970, 1, 1)` for every pending transaction, so `posted or transacted` picked that wrong date and never fell through to the correct `transacted_at` value sitting right next to it.
- Net effect: pending transactions synced from SimpleFIN silently get dated to 1970-01-01, making them invisible in any date-filtered view even though the sync completes without error.

## Fix
Treat a falsy `posted` (covers `0`, `None`, missing) as "no posted date yet" before converting, so the fallback to `transacted_at` actually kicks in:

```python
posted = _epoch_to_date(raw.get("posted")) if raw.get("posted") else None
```

## Test plan
- [x] Added `test_get_transactions_pending_tx_uses_transacted_at_not_epoch`, reproducing the exact bug: a pending transaction with `"posted": 0` and a real `transacted_at`, asserting the parsed date matches `transacted_at` and is not `1970-01-01`.
- [x] Added `test_epoch_to_date_zero_is_technically_the_epoch` documenting that `_epoch_to_date(0)` itself is correct — the bug is in the caller treating that as meaningful rather than in the helper.
- [x] Full `backend/tests/test_providers_simplefin.py` suite passes (20/20) locally.
- [x] `ruff check` clean on both changed files.

Fixes #447

### DIFF
--- a/backend/app/providers/simplefin.py
+++ b/backend/app/providers/simplefin.py
@@ -380,7 +380,11 @@ class SimpleFinProvider(BankProvider):
             return None
         txn_type = "debit" if amount_raw < 0 else "credit"
         amount = amount_raw.copy_abs()
-        posted = _epoch_to_date(raw.get("posted"))
+        # SimpleFIN sends "posted": 0 (not null) for transactions that haven't
+        # posted yet — a falsy epoch, not a real one, so treat it as unset
+        # rather than letting it resolve to 1970-01-01 and shadow the correct
+        # transacted_at date below.
+        posted = _epoch_to_date(raw.get("posted")) if raw.get("posted") else None
         transacted = _epoch_to_date(raw.get("transacted_at"))
         txn_date = posted or transacted
         if not txn_date:
--- a/backend/tests/test_providers_simplefin.py
+++ b/backend/tests/test_providers_simplefin.py
@@ -79,6 +79,13 @@ def test_epoch_to_date_parses_seconds():
     assert _epoch_to_date(1672531200) == date(2023, 1, 1)
 
 
+def test_epoch_to_date_zero_is_technically_the_epoch():
+    # _epoch_to_date(0) legitimately returns 1970-01-01 — it's callers that
+    # must decide whether 0 means "no value" for their field. See
+    # test_get_transactions_pending_tx_uses_transacted_at_not_epoch below.
+    assert _epoch_to_date(0) == date(1970, 1, 1)
+
+
 # ----- claim flow -------------------------------------------------------------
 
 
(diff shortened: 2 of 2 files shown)


tassionoronha:
Hey @brandonjones24, welcome here and thanks for the PR :)

Althought, I'll rather close this one as superseded: #375 landed the same fix inside _epoch_to_date (rejects seconds <= 0), which also covers string "0" and the holdings created field. Thanks for the clear diagnosis.

PR #450, opened 2026-07-21
fix: guard SimpleFIN holding currency against non-ISO tickers

### LINKED ISSUE #448: [Bug] SimpleFIN sync fails entirely when a crypto holding's ticker (e.g. DOGE) overflows the currency column
## Description

If a SimpleFIN-connected investment account holds a cryptocurrency whose ticker is longer than 3 characters (e.g. DOGE, SHIB, AVAX), the background sync for that **entire connection** crashes with a Postgres `StringDataRightTruncationError: value too long for type character varying(3)`, and the connection is left showing an "error" badge — even though every other account/holding in that connection synced successfully first.

## Steps to Reproduce

1. Connect a SimpleFIN account that holds a cryptocurrency with a 4+ character ticker (e.g. Dogecoin).
2. Trigger a sync.
3. The sync crashes partway through; the Accounts page shows the connection with an "error" badge; the crypto holding is never created (the whole transaction rolls back).

## Expected Behavior

The holding should be stored — with its ticker in the dedicated `ticker` column and a valid fiat currency (or a sensible default) in `currency` — and the sync should complete successfully for the rest of the connection.

## Actual Behavior

In `app/providers/simplefin.py`, `get_holdings()` does:

```python
currency=raw.get("currency") or acc_currency,
```

For crypto holdings, SimpleFIN's raw payload puts the ticker itself (e.g. `"DOGE"`) in the `currency` field, since the holding isn't denominated in a fiat currency. That value is written straight into `HoldingData.currency`, which maps to the `assets.currency` column — `VARCHAR(3)` — and a 4-character ticker blows the constraint, crashing the whole sync transaction and taking the rest of that connection's accounts down with it (all-or-nothing).

The code already captures the real ticker separately (`metadata={"symbol": raw.get("symbol"), ...}`), but never writes it to the schema's dedicated `assets.ticker` column, which exists for exactly this purpose and currently goes unused by `get_holdings()`.

## Suggested Fix

- Populate `ticker=raw.get("symbol")` on the `HoldingData`.
- Only use `raw.get("currency")` for `currency` when it's a valid ≤3-character code; otherwise fall back to `acc_currency` (or the workspace's default currency).

## Environment

- Deployment: Docker
- Version: v0.13.10

brandonjones24:
## Summary
- Some SimpleFIN connectors put a crypto/asset ticker (e.g. `"DOGE"`) in a holding's `currency` field instead of an ISO 4217 code.
- `HoldingData.currency` is stored in a `VARCHAR(3)` column, so a longer ticker overflows the write and crashes the *entire* account sync — not just that one holding.
- Added `_holding_currency()` in `backend/app/providers/simplefin.py`, which only accepts the holding's `currency` if it's a plausible 3-letter alphabetic code, otherwise falling back to the account-level currency. The unrelated `symbol`/ticker metadata is untouched.

Fixes #448

## Test plan
- [x] Added `test_get_holdings_crypto_ticker_currency_falls_back_to_account_currency` reproducing a `DOGE`-ticker holding and asserting it resolves to the account's `USD` currency instead of crashing/overflowing.
- [x] `pytest tests/test_providers_simplefin.py -v` — 19/19 passed.
- [x] `ruff check` on both changed files — clean.

### DIFF
--- a/backend/app/providers/base.py
+++ b/backend/app/providers/base.py
@@ -136,6 +136,11 @@ class HoldingData:
     purchase_price: Optional[Decimal] = None
     purchase_date: Optional[date] = None
     isin: Optional[str] = None
+    # Exchange/asset symbol (e.g. "AAPL", "DOGE") for the dedicated
+    # `assets.ticker` column. Kept separate from `currency`: connectors that
+    # report a crypto position put the ticker in both, and only one of them
+    # belongs in a 3-char ISO currency field.
+    ticker: Optional[str] = None
     maturity_date: Optional[date] = None
     is_withdrawn: bool = False  # provider signaled the position was sold/transferred
     metadata: Optional[dict] = None
--- a/backend/app/providers/simplefin.py
+++ b/backend/app/providers/simplefin.py
@@ -121,21 +121,35 @@ def _to_decimal(value: Any) -> Optional[Decimal]:
         return None
 
 
-def _holding_currency(raw: dict, acc_currency: str) -> str:
-    """Pick a currency code for a holding, guarding against non-ISO tickers.
+def _iso_currency(value: Any, fallback: Optional[str]) -> Optional[str]:
+    """Normalize a connector-supplied currency, guarding against non-ISO values.
 
-    SimpleFIN's per-holding ``currency`` field is populated by the bank
-    connector, and for crypto/brokerage holdings some connectors put the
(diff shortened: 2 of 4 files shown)


tassionoronha:
Thanks for this one @brandonjones24 !

I extended the guard to the account and transaction currency paths, and populated assets.ticker. 
Also rebased for CI.

LGTM!

PR #459, opened 2026-07-25
fix: convert enable banking consent validity from seconds to days

pjrm:
## What

Enable Banking returns maximum_consent_validity in seconds (as documented in the ASPs data reference [here](https://enablebanking.com/docs/api/reference/#aspspdata)), but the code was incorrectly treating the raw value as days. This change converts the value using timedelta(seconds=...).days before storing it.

Additionally, this PR ensures that valid_until_days is passed to getOAuthUrl, so the configured consent validity is correctly applied when generating the OAuth URL.

## Why

While attempting to add Trade Republic through Enable Banking, I encountered an error indicating that the consent validity could not exceed 90 days.

After investigating the issue, I found that the validity value was being stored incorrectly. Additionally, we were not passing the configured validity when invoking the Enable Banking API, causing it to fall back to the default value of 180 days. This PR fixes both issues by correctly storing the configured validity and ensuring it is included in the API request.

## How to Test

1. Set up the Enable Banking sandbox application
3. Define Enable Banking environment variables in the .env file
4. Add a bank integration by selecting a mock bank
6. Test the failure by adding one more day, like `max_consent_days=maximum_consent_validity+1,`

## Related Issue

Closes #___ (or link the issue this addresses).

## Checklist

### DIFF
--- a/backend/app/providers/enable_banking.py
+++ b/backend/app/providers/enable_banking.py
@@ -299,6 +299,9 @@ class EnableBankingProvider(BankProvider):
             inst_country = (item.get("country") or "").upper()
             if inst_country:
                 countries.add(inst_country)
+            if (maximum_consent_validity := item.get("maximum_consent_validity", None)) is not None:
+                maximum_consent_validity = timedelta(seconds=maximum_consent_validity).days
+
             institutions.append(
                 InstitutionData(
                     name=item.get("name") or "",
@@ -307,7 +310,7 @@ class EnableBankingProvider(BankProvider):
                     logo=item.get("logo"),
                     bic=item.get("bic"),
                     psu_types=list(item.get("psu_types") or []),
-                    max_consent_days=item.get("maximum_consent_validity"),
+                    max_consent_days=maximum_consent_validity,
                 )
             )
         institutions.sort(key=lambda i: (i.country, i.display_name.lower()))
--- a/backend/tests/test_providers_enable_banking.py
+++ b/backend/tests/test_providers_enable_banking.py
@@ -167,7 +167,7 @@ async def test_list_institutions_maps_and_dedupes_countries(eb_keys):
                         "logo": "https://l/revolut.png",
                         "bic": "REVOLT21",
                         "psu_types": ["personal"],
-                        "maximum_consent_validity": 180,
+                        "maximum_consent_validity": 15552000,
                     },
                     {"name": "Sparkasse", "country": "DE"},
                     {"name": "BNP Paribas", "country": "FR"},
--- a/frontend/src/components/oauth-connect-dialog.tsx
+++ b/frontend/src/components/oauth-connect-dialog.tsx
@@ -15,6 +15,7 @@ type Institution = {
   name: string
   display_name: string
   country: string
+  max_consent_days?: number | null
   logo?: string | null
 }
 
@@ -108,6 +109,7 @@ export function OAuthConnectDialog({ open, onClose, provider, supportsAssetSync 
       const url = await connections.getOAuthUrl(provider, {
         country,
         institution_name: institution.name,
+        valid_until_days: institution.max_consent_days,
         ...(supportsAssetSync ? { sync_assets: syncAssets } : {}),
       })
       window.location.assign(url)


tassionoronha:
Thanks @pjrm !

I checked this against the live EB data and the values look right now (7/89/90/180 days). 

One small ask: could you revert the `or` → `.get(k, default)` swaps on L295/299/307/312? 
A null in the response would break the whole list there, no? Even if Idk they would send it.

pjrm:
> Thanks @pjrm !
> 
> I checked this against the live EB data and the values look right now (7/89/90/180 days).
> 
> One small ask: could you revert the `or` → `.get(k, default)` swaps on L295/299/307/312? A null in the response would break the whole list there, no? Even if Idk they would send it.

Hi @tassionoronha! If I'm understanding your concern correctly, I think both versions behave the same if `item`  is `None`, since Python will fail on the first `.get()` call:

```
>>> a = {"a": 1}
>>> a.get("a")
1
>>> a.get("b", "")
''
>>> a.get("b") or ""
''
>>> b = None
>>> b.get("test") or ""
Traceback (most recent call last):
  File "<python-input-10>", line 1, in <module>
    b.get("test") or ""
    ^^^^^
AttributeError: 'NoneType' object has no attribute 'get'
>>> b.get("test")
Traceback (most recent call last):
  File "<python-input-9>", line 1, in <module>
    b.get("test")
    ^^^^^
AttributeError: 'NoneType' object has no attribute 'get'
```

So if we want to guard against that case, would it make sense to validate the item before accessing it? Something like:

```
if not isinstance(item, dict):
    continue
```

Is that what you had in mind?

BTW, I'm happy to revert to the previous `or` version. I just thought this was a bit more readable.

tassionoronha:
You're right that if `item` itself is `None` both versions blow up the same way. That wasn't the case I meant, my wording was unclear.

The difference is when the key is **present with a `null` value**. `.get(k, default)` only falls back when the key is *absent*, so a JSON `null` comes through as `None`:

```python
>>> item = {"country": None}
>>> item.get("country", "").upper()
AttributeError: 'NoneType' object has no attribute 'upper'
>>> (item.get("country") or "").upper()
''
```

Same shape for the others: `data.get("aspsps", [])` and `list(item.get("psu_types", []))` raise `TypeError: 'NoneType' object is not iterable`, where the `or` form gives `[]`. `name` is the sneaky one, it doesn't raise in the loop but later at `institutions.sort(key=lambda i: ... i.display_name.lower())`.

There's no per-item try/except, so any one of those takes down the whole `/institutions` response (the router turns it into a 502), not just the one bank.

To be fair, EB documents `name`, `country` and `psu_types` as required, so it likely never fires in practice. That's why I'm not treating it as a bug, just that the swap gives up robustness for nothing and isn't part of the conversion fix. So a plain revert of those four lines is all I'm after, no need for the `isinstance` guard.

The seconds→days conversion itself is spot on, verified against live data.

pjrm:
> You're right that if `item` itself is `None` both versions blow up the same way. That wasn't the case I meant, my wording was unclear.
> 
> The difference is when the key is **present with a `null` value**. `.get(k, default)` only falls back when the key is _absent_, so a JSON `null` comes through as `None`:
> 
> ```python
> >>> item = {"country": None}
> >>> item.get("country", "").upper()
> AttributeError: 'NoneType' object has no attribute 'upper'
> >>> (item.get("country") or "").upper()
> ''
> ```
> 
> Same shape for the others: `data.get("aspsps", [])` and `list(item.get("psu_types", []))` raise `TypeError: 'NoneType' object is not iterable`, where the `or` form gives `[]`. `name` is the sneaky one, it doesn't raise in the loop but later at `institutions.sort(key=lambda i: ... i.display_name.lower())`.
> 
> There's no per-item try/except, so any one of those takes down the whole `/institutions` response (the router turns it into a 502), not just the one bank.
> 
> To be fair, EB documents `name`, `country` and `psu_types` as required, so it likely never fires in practice. That's why I'm not treating it as a bug, just that the swap gives up robustness for nothing and isn't part of the conversion fix. So a plain revert of those four lines is all I'm after, no need for the `isinstance` guard.
> 
> The seconds→days conversion itself is spot on, verified against live data.

I see your point, and you're totally right. You're talking about not using the default value if the value is None, which could break downwards. 

I've reverted those lines.
Thank you for clarifying 🙏

tassionoronha:
Thanks for the changes @pjrm !! LGTM now :)

PR #455, opened 2026-07-24
add French (fr) translation

tdechanterac:
## What

Adds a complete French (`fr`) locale and wires it up everywhere the other languages are registered. Used #373 as an inspiration.

## Why

French was the only major Western-European language missing from the UI (de/es/it/pl/pt-BR/ru/uk were already supported), leaving French-speaking users on the English fallback. This closes that gap. Per CONTRIBUTING, translation updates are welcome without prior discussion.

## How to Test

Steps to verify the changes work:

1. In the UI, click on the User Profile Menu (bottom-left avatar/name) -> **Language** and switch it to **Français** (or change the language under **Workspace Settings** / **Admin Settings**).
2. Create a **new workspace** (e.g., "Test FR") while the UI is set to French.
3. Navigate to the **Categories** tab in the newly created workspace. Verify that the default category groups (*Logement*, *Alimentation & Restaurants*, *Revenus*, etc.) and default categories (*Courses*, *Santé*, *Abonnements*, *Autres*, etc.) are generated in French.

## Checklist

- [x] Backend tests pass (`pytest`) — added `test_create_default_groups_french` and `test_create_default_categories_french`, mirroring the German tests from #373. 
- [x] Frontend lints clean (`npm run lint`)
- [x] Frontend builds (`npm run build` / `tsc -b` clean)
- [x] Translations updated (if user-facing strings changed) — new `frontend/src/locales/fr.json`; key + placeholder parity with `en.json` verified by the `i18n.test.ts` locale test (all locales pass)

PR #461, opened 2026-07-26
add European Portuguese (pt-PT) translation

11VitorAlves11:
## What

Adds European Portuguese (`pt-PT`) as a supported language, alongside the existing `pt-BR`.

**Frontend**
- New `frontend/src/locales/pt-PT.json` — full locale, key and placeholder parity with `en.json` (1559 keys).
- Registers the bundle in `lib/i18n.ts`. `resolveSupportedLang` maps an explicit `pt-PT` tag to the new bundle; a bare `pt` keeps its historical `pt-BR` mapping, so nothing changes for existing users.
- Adds the date-fns `pt` (European Portuguese) locale in `lib/date-fns-locale.ts`.
- Adds a **Português (PT)** option to all five language switchers (`app-layout`, `workspace-switcher`, `workspace-settings`, `admin/settings`, `setup`). The existing entry is now qualified as **Português (BR)** in the two switchers where it was just "Português", since a bare "Português" is ambiguous once both variants are offered.

**Backend** — mirrors the French localization from #455, so `pt-PT` users get localized defaults instead of the English fallback:
- `pt-PT` key added to `DEFAULT_GROUPS_I18N` (`category_group_service.py`) and `DEFAULT_CATEGORIES_I18N` (`category_service.py`).
- Coverage tests mirroring the French ones.

The `lang` value already flows from the user's UI language through `create_default_groups` / `create_default_categories`, and missing langs fall back to English, so the backend part is additive and safe.

## Why

Portuguese was only available as `pt-BR`. The two variants differ enough in everyday finance vocabulary that Brazilian Portuguese reads as noticeably foreign to European Portuguese speakers — and in a few cases it is actively misleading.

This is a genuine European Portuguese translation, not a `pt-BR` copy:

- **Register**: `tu`-form address, matching the informal register `es`, `it` and `de` already use (`workspace.you` is `tú` / `tu` / `Du` in those locales).
- **Vocabulary**: `ficheiro`, `utilizador`, `palavra-passe`, `definições`, `guardar`, `eliminar`, `ecrã`, `percentagem`, `subscrições`.
- **Grammar**: progressive `a` + infinitive instead of the gerund (`A carregar…` rather than `Carregando…`), definite article before possessives (`cria a tua conta`), `precisam de ser` rather than `precisam ser`.
- **False friends avoided**: Income is **Rendimentos**, not "Renda" (which means *rent* in pt-PT); Nickname is **Nome personalizado**, not "Apelido" (which means *surname*).
- `excluir` was kept only where the English source means *exclude*; where it means *delete* it became `eliminar`.

## How to Test

1. `cd frontend && npm run dev`
2. Open the user menu → **Language** → **Português (PT)**. The UI switches to European Portuguese and the choice persists in localStorage.
3. Check that dates render with the European Portuguese date-fns locale (e.g. the month stepper and transaction dates).
4. Verify the option also appears in the workspace switcher, workspace settings, admin settings, and the first-run setup screen (as `PT-PT`).
5. Create a new workspace with the UI language set to Português (PT) — the seeded default category groups and categories come out in European Portuguese (`Habitação`, `Rendimentos`, `Supermercado`, `Subscrições`).
6. `cd frontend && npx vitest run src/locales/i18n.test.ts` — parity, duplicate-key and placeholder checks for the new locale.

## Related Issue

No issue — per [CONTRIBUTING](../CONTRIBUTING.md#where-to-start), translation updates are welcome without prior discussion.

## Checklist

- [x] Backend tests pass (`pytest`) — 2342 passed, 7 skipped. One unrelated flake, `test_agents_jwt.py::test_short_ttl_respected`, fails only under full-suite load on a slow filesystem and passes in isolation; it does not touch i18n.
- [x] Frontend lints clean (`npm run lint`) — 0 errors (41 pre-existing warnings, untouched by this PR)
- [x] Frontend builds (`npm run build`) — `tsc -b` also type-checks the new `SupportedLang` member against the date-fns locale map
- [x] For a large or core change, I discussed it first — n/a, translation-only addition


tassionoronha:
The pt-pt translation is defo welcoming here! 
The changes LGTM! Obrigado :)

PR #469, opened 2026-07-28
feat: choose archived account history visibility

gabrielvcbessa:
## What changed

- Adds an explicit choice when archiving an account:
  - **Keep history in reports** (recommended/default)
  - **Exclude the entire account history**
- Keeps archiving operationally consistent in both modes: synchronization, current balances, and future projections stop using the account.
- Applies one shared history-visibility predicate across transactions, search, dashboard history, and reports instead of duplicating feature-specific rules.
- Hides excluded history without deleting any transactions, and provides an explicit transaction-list audit override (`include_excluded_history=true`).
- Reopening an account restores its history.
- Shows the selected history policy on closed-account rows and adds translations for all supported locales.

## Why

Closing an account currently conflates two different decisions:

1. Stop treating the account as active.
2. Remove every past transaction from financial history.

That rewrites legitimate history for normal closed accounts, but retaining history is also undesirable for duplicate, test, or incorrectly imported accounts. Separating account lifecycle from history visibility gives users control without destructive deletion.

The root cause was repeated `Account.is_closed == false` conditions across independent reporting queries. This change centralizes the historical rule so transaction history, search, dashboards, and reports stay consistent.

## Behavior

| Behavior | Keep history | Exclude history |
| --- | --- | --- |
| Sync / current balance / future projections | Stopped | Stopped |
| Past transactions, search, dashboards, reports | Included | Hidden |
| Stored transactions deleted | No | No |
| Reopen restores visibility | Yes | Yes |

## Compatibility and migration

- Migration `066` adds the non-null `exclude_from_history` preference.
- Existing archived accounts are marked as excluded during migration, preserving current behavior.
- Active accounts default to retained history.
- New archive requests default to keeping history, while API clients can explicitly request exclusion with `exclude_history=true`.

## Tests

- Full backend suite: `2343 passed, 7 skipped`
- Focused account-closure/report suite: `119 passed, 4 skipped`
- Frontend tests: `40 passed`
- Production Vite bundle: passed
- Ruff: passed
- Alembic: `066 (head)`
- ESLint on changed frontend files: 0 errors (2 pre-existing warnings in the account editor)

The repository-wide TypeScript build still reports two existing formatter type errors in the unchanged Reports page (`reports.tsx:839` and `reports.tsx:1299`); the production Vite bundle succeeds.


tassionoronha:
Hey @gabrielvcbessa, thanks for taking this on! Separating account lifecycle from history visibility was overdue, and centralizing the rule into one predicate is the right move.

Two blockers before merge:

**1. Balance history no longer reconciles with the total balance tile.**

`_daily_deltas` now uses `account_history_is_visible()`, but its anchor (`_balance_at` -> `_get_open_accounts`) still filters on `is_closed`. Under the new "keep history" default the archived account is missing from the starting balance but present in the deltas, so the curve drifts and its final point stops matching the total balance card on the same screen. Confirmed locally, and the same scenario reconciles fine when the account is excluded, so it's new here.

Related: the dashboard test was edited to add `exclude_from_history=True` to stay green, which moves it off the path the code now takes by default. The keep-history branch ends up with no coverage.

**2. The migration doesn't preserve current behavior for the list, search and export.**

On `main` the only closure filter lives inside `if user_pnl_only:`, so `/api/transactions`, search and CSV export all return closed-account rows today. Migration 066 flips every existing closed account to excluded, so those rows disappear on upgrade day.

Export bothers me most: neither call site passes `include_excluded_history`, including the `transaction_ids` branch where the user hand-picked the rows. A CSV that silently returns a subset is rough in a project built on data ownership.

**Worth fixing too:**

- `include_excluded_history` has no frontend caller, so `account-detail.tsx` renders an empty transaction list for archived accounts with no explanation.
- `get_budget_vs_actual` never joins `Account`, so excluded history still inflates budget actuals. Contradicts the "hidden from reports" copy.
- `user_pnl_only` lost its `is_closed` guard.
- No way to change the policy after archiving, and `reopen_account` resets it silently.
- `pt-PT.json` didn't get the new keys.
- #477 also adds a migration `066`. Second to merge branches the Alembic history.

Drove the UI and the dialog is solid: good default, clean selection state, i18n renders. Only nit is that the exclude option is styled identically to the recommended one despite being the consequential choice.

Happy to pair on the balance history fix.

tassionoronha:
Welcome here @gabrielvcbessa ! 

I made this first review with Claude which highlighted some critical points.
For a first PR I would suggest to start with a small fix/improv instead of touching deep business rules of the project. 

You're free to continue this one as well, but don't hesitate to create an issue before-hand or reach people on the Discord group to confirm the right approach for this nature of PR.

Cheers and thank you for collaborating :)

tassionoronha:
Closing PR by lack of activity and team validation beforehand. 
You can always reach us out on Discord and we're happy do discuss more about this point before jump in a PR :)
Cheers! :)

Issue #98, opened 2026-04-20
[Bug]  OFX monthly credit card imports skip later installments as duplicates + Brazilian installment workflow support


ruanvamaral:
Title: OFX monthly credit card imports skip later installments as duplicates + Brazilian installment workflow support

**Description**:

I import one OFX file per month from my credit card statement.

In Brazil, credit card installment purchases are extremely common. A single purchase is divided into monthly charges ("parcelas"), and each installment appears on future monthly statements until fully paid.

**Current issue:**

When I import the OFX file from the month of purchase, only the first installment appears.

Example:

December import (purchase date):

* Nimbus Stay * Q7K2P9L1 - Parcela 1/6

Then, after importing later monthly OFX files, the following expected installments do not appear:

* January:  * Nimbus Stay * Q7K2P9L1  Parcela 2/6
* February: * Nimbus Stay * Q7K2P9L1  Parcela 3/6
* March:  * Nimbus Stay * Q7K2P9L1  Parcela 4/6
* etc.

Observed behavior:

Only the first installment is visible. Future installments from later monthly imports seem to be skipped, likely as duplicates.

Expected behavior:

Each installment shown in each monthly OFX statement should be imported as an independent transaction.

Why this is especially important in Brazil:

Brazilian credit card systems commonly allow purchases in multiple installments (3x, 6x, 10x, 12x, etc.). This is a core consumer behavior, not an edge case.

For many users, a large share of monthly expenses comes from installment charges. Therefore, proper installment handling is essential for financial visibility.

Without this support, reports become inaccurate because future monthly obligations disappear.

**Date expectation / statement cycle context:**

The first import may contain the original purchase date (December).

However, later installments should appear in subsequent months according to the credit card billing cycle, typically when a new monthly statement opens.

**So users expect:**

* December statement → installment 1/6
* January statement → installment 2/6
* February statement → installment 3/6
* March statement → installment 4/6

Even if dates differ depending on bank/card cycle.

**Technical suspicion:**

Duplicate detection may be matching by description/value and ignoring:

* installment number (1/6, 2/6, 3/6...)
* statement month
* OFX FITID unique identifiers

Suggested long-term feature:

Native installment management for Brazilian-style credit card purchases.

**Possible workflow:**

* Detect installment transactions automatically
* Generate future projected installments
* Show upcoming monthly obligations
* Reconcile projected installments with real imported OFX entries later

Also useful when manually creating a transaction:

* Add transaction
* Enable installments
* Select number of months
[...]

PR #474, opened 2026-07-29
fix: improved csv import resilience and added support to map payee external id and notes

jorishaenseler15:
## What

Added support for mapping and importing `Payee`, `External ID`, and `Notes` from CSV files. Also improved the CSV parser's resilience against irregular date formats, empty/missing columns, and overly long text inputs. Finally, fixed a frontend caching bug where newly created payees didn't appear in dropdowns immediately after an import.

## Why

Users need to be able to map custom columns (like notes and external IDs) when importing CSVs from various banks. Also it was not possible to map the payee column. Additionally, the old CSV parser was too fragile: it would crash the database on very long payee names (throwing a PostgreSQL 255-character string truncation error), it failed to parse dates separated by dots (e.g., `15.01.2026`), and the UI didn't automatically invalidate the `payees` cache after a successful import, forcing users to reload the page to assign payees.

## How to Test

1. Go to the Import Transactions page and upload a CSV file containing `Payee`, `External ID`, and `Notes` columns.
2. Verify that the UI allows you to map these columns, and that the imported transactions correctly save these fields in the database.
3. Edit a newly imported transaction immediately after import and verify that the newly created Payee appears in the dropdown without needing a manual page reload.
4. Upload a CSV with an extremely long string (over 255 characters) mapped to the Payee column; verify it imports safely (truncating to 255 characters) instead of crashing with a `500 Internal Server Error`.
5. Upload a CSV with dotted date formats (e.g., `15.01.2026`) and verify they are successfully parsed.

## Checklist


tassionoronha:
Nice work, this fixes real papercuts. One blocker though.

Auto-detecting an id column as external_id can silently drop transactions. external_id drives dedup on (account, external_id, date), ignoring amount and description (from #98), so row numbers become the dedup key:

import A -> imported=2 skipped=0   (ids 1,2)
import B -> imported=0 skipped=2   (ids 1,2, same date)
descriptions in DB: ['Coffee', 'Taxi']

Export B was a Rent payment and a Salary. Both discarded as "duplicates". Without this change the content-based fallback handles that case fine.

Can we drop the auto-detect and keep external_id as an explicit mapping only?

Also pt-PT is missing the three new keys, so CI goes red on rebase.

Minor: a row with fewer fields than the header still crashes on .replace, and PayeeUpdate.notes didn't get the max_length that PayeeCreate.notes did.

jorishaenseler15:
Thanks for your review @tassionoronha!

I adressed all your found issues/improvements - can you have a look again at it?

Thanks!

tassionoronha:
Great improvs @jorishaenseler15 , LGTM now :)
Just dropped the uv since we are not using it yet.

PR #475, opened 2026-07-29
feat: added bulk delete for payees

jorishaenseler15:
## What

- Implemented bulk deletion for payees in both the frontend and backend.
- Added a bulk delete button in the payees overview that appears when multiple rows are selected (more than 1).
- Added per list entry delete button icon.
image

## Why

Users need an efficient way to clean up multiple unwanted payees at once. Additionally the user should be able to delete one entry directly on the list payee view.

## How to Test

1. Navigate to the Payees page in the frontend.
2. Check the boxes next to multiple payees to select them.
3. Click the red "Delete (X)" button that appears at the top right of the table.
4. Click the trash bin icon on a single payee row and verify it deletes the payee directly, without a native popup dialog.
6. Verify the backend handles bulk deletions correctly by running the tests: `uv run pytest tests/test_payee_service.py tests/test_payees_api.py`.

## Checklist


frontend/src/pages/payees.tsx
```diff
@@ -220,6 +222,21 @@ export default function PayeesPage() {
...
   })
 
+  const bulkDeleteMutation = useMutation({
+    mutationFn: (ids: string[]) => payeesApi.bulkDelete(ids),
+    onSuccess: (result) => {
+      invalidateFinancialQueries(queryClient)
+      queryClient.invalidateQueries({ queryKey: ['payees'] })
+      setSelectedIds(new Set())
+      setLastSelectedId(null)
+      if (summaryPayee && selectedIds.has(summaryPayee)) {
```
MarlonPassos-git: Could we use the mutation variables here instead of reading the current `selectedIds` state?

The selection can change while the request is pending, so `selectedIds` may no longer represent the payees that were actually deleted when `onSuccess` runs.

For example:

```tsx
onSuccess: (result, deletedIds) => {
  if (summaryPayee && deletedIds.includes(summaryPayee)) {
    setSummaryPayee(null)
  }
}
```

This would also be consistent with `mergeMutation`, which uses the mutation variables inside `onSuccess`.

frontend/src/pages/payees.tsx
```diff
@@ -275,10 +292,16 @@ export default function PayeesPage() {
...
-                <Button variant="outline" onClick={() => { setMergeTargetId(''); setMergeDialogOpen(true) }}>
-                  <Merge size={16} className="mr-1.5" />
-                  {t('payees.merge')} ({selectedIds.size})
-                </Button>
+                <div className="flex items-center gap-2">
+                  <Button variant="outline" onClick={() => { setMergeTargetId(''); setMergeDialogOpen(true) }}>
+                    <Merge size={16} className="mr-1.5" />
+                    {t('payees.merge')} ({selectedIds.size})
+                  </Button>
+                  <Button variant="destructive" onClick={() => bulkDeleteMutation.mutate(Array.from(selectedIds))} disabled={bulkDeleteMutation.isPending}>
```
MarlonPassos-git: Could we add a confirmation dialog before executing this mutation?

Deleting a payee also clears its association from existing transactions, so this is more destructive than simply removing an item from the list. An accidental click currently performs the operation immediately.

The same confirmation should probably also be applied to the new per-row delete action below, which directly calls `deleteMutation.mutate(payee.id)`.

frontend/src/locales/pt-BR.json
```diff
@@ -1200,7 +1200,8 @@
     "typeCompany": "Empresa",
     "allTypes": "Todos os tipos",
     "filterByType": "Filtrar por tipo",
-    "favoritesOnly": "Apenas favoritos"
+    "favoritesOnly": "Apenas favoritos",
+    "deletedMultiple": "{{count}} beneficiários excluídos"
```
MarlonPassos-git: The branch is currently behind main, which recently added the pt-PT locale. After updating the branch, could we also add payees.deletedMultiple to pt-PT.json?

backend/tests/test_payee_service.py
```diff
@@ -282,6 +283,40 @@ async def test_delete_payee_nulls_transaction_refs(session: AsyncSession, test_u
...
+    assert await get_payee(session, p3.id, test_workspace.id) is not None
+
+    await session.refresh(tx)
+    assert tx.payee_id is None
+
+
+@pytest.mark.asyncio
+async def test_bulk_delete_payees_invalid_ids(session: AsyncSession, test_user, test_workspace):
+    deleted = await bulk_delete_payees(session, test_workspace.id, [uuid.uuid4(), uuid.uuid4()])
+    assert deleted == 0
```
MarlonPassos-git, line 317: Could we add test coverage for a request containing a payee from another workspace, and for deleting a payee that has an associated PayeeMapping? These would verify both workspace isolation and the mapping cleanup performed by the service.

jorishaenseler15: @MarlonPassos-git I added two tests to verify workspace isolation and mapping cleanup

frontend/src/locales/pt-BR.json
jorishaenseler15: @MarlonPassos-git I added the missing keys and now all languages should be up to date!

frontend/src/pages/payees.tsx
jorishaenseler15: @MarlonPassos-git I added a dialog to let the user confirm its actions!

frontend/src/pages/payees.tsx
jorishaenseler15: @MarlonPassos-git should be fixed now!

tassionoronha:
Removed the uv here, but apart of that I think we are now in a good place to merge. 
Waiting for @MarlonPassos-git check before continue.

MarlonPassos-git (approved):
LGTM

PR #477, opened 2026-07-30
Allow hiding default categories

### LINKED ISSUE #348: [Bug] Unable to remove initial categories
## Description

Starting with a fresh setup, several categories and groups are created automatically. The user is not able to remove any of these items. Only user-created categories or groups are removable.

## Steps to Reproduce

1. Create a new workspace
2. Go to categories
3. Try to remove an existing category

## Expected Behavior

All categories and category groups should be removable.

## Actual Behavior

Only user-created categories and category groups are removable. The button is not available for default items.

## Environment

- Deployment: Docker
- Browser: Firefox
- Version v0.13.6+8bf4273

## Screenshots

Image

## Additional Context

I understand this may be a limitation due to the data-model. If that's the case, we could allow hiding categories (YNAB does something similar).

RomuloGatto:
## Summary
- Adds non-destructive hide/show support for system default categories and category groups.
- Keeps user-created categories/groups deletable while preserving the existing delete guard for system defaults.
- Filters hidden defaults from normal category/category-group API lists, with `include_hidden=true` for management screens.
- Shows Hidden badges and eye/eye-off controls on the Categories management page.

Closes #348

## Screenshots

Default categories before hiding, with system defaults managed through hide controls:

Default categories visible

Hidden default category group and individual hidden category remain visible in management with Hidden badges and show controls:

Hidden default categories

## Verification
- `cd backend && uv run --extra dev ruff check .`
- `cd backend && uv run --extra dev pytest tests/test_categories_api.py tests/test_category_groups_api.py tests/test_category_service.py tests/test_category_group_service.py -q` — 56 passed
- `python3 -m compileall -q backend/app`
- `cd frontend && npm run lint` — passed with existing warnings
- `cd frontend && npm test -- --run` — 45 passed
- `git diff --check`

## Notes
- Screenshots were captured from the real frontend running locally with deterministic route stubs for auth/workspace/category APIs, so no secrets or live user data are present.


tassionoronha:
Hey @RomuloGatto, thanks for this one! A visibility flag instead of trying to delete seeded rows that transactions still point at is exactly right, and the migration docstring explaining why is the kind of thing future readers appreciate. Management page is well built too: dimming, Hidden badge, eye toggle in place of delete, and full locale coverage including pt-PT.

Two things before merge:

**1. Hiding a category with transactions breaks the transaction editor.**

`category-select.tsx` resolves the selected item by looking the value up in the list it's handed, and every consumer now gets the hidden-filtered list. Hide a default category that has transactions and open one: the list row still shows the name correctly, but the edit dialog's Category field reads "Select category...", as if it were uncategorized.

Good news, I traced the save path and there's no data loss. `categoryId` is seeded from `seed?.category_id`, not from the list, so saving writes the original id back. It's a display lie rather than corruption, but it still misreports state in a finance app and invites users to "fix" something that isn't broken.

Fix is to let the picker render a current value that isn't in the list, using the Hidden treatment you already built.

**2. Migration `066` collides with #469.**

Same revision, same `down_revision`. Whichever lands second gives everyone two Alembic heads and breaks `alembic upgrade head`.

**Worth a look:**

- No `is_system` guard on `is_hidden`. `update_category` does a blanket `setattr` over `exclude_unset`, so any category in the workspace can be hidden via API, and the UI only renders the eye toggle inside the `cat.is_system ?` branch. A user category hidden that way has no path back to visible.
- "Hidden" is narrower than the label suggests. It only affects the two list endpoints; dashboard and reports query `Category` directly. I confirmed a hidden category still shows in Spending by Category, and budgets behave the same way. Defensible scope, just worth documenting or extending.
- The `setFormIgnoreTransfer(cat?.is_ignored ?? false)` line is a real unrelated bug fix. Nice catch, worth calling out in the description so it doesn't read as noise.

Drove the whole flow in Chrome and everything else behaves end to end. The `(Hidden)` suffix on a currently-assigned group in the dropdown is a thoughtful touch.

RomuloGatto:
Thanks @tassionoronha below was addressed:
    
1. Existing hidden category assignments
    - CategorySelect now accepts a current-category fallback when the selected ID is absent from the visible option list.
    - The current category remains visible with a Hidden badge, while hidden categories are still excluded from new selections.
    - This is wired into transaction, recurring-transaction, and rule editors.
    
2. Migration 066 collision
    - #477 remains the valid sole 066 -> 065 head against current main.
    - Since #469 is currently conflicting and still has unresolved blockers, the intended safe merge order is #477 first, followed by rebasing #469 to 067 -> 066.
    - If #469 is selected to land first instead, I’ll rebase this PR and renumber its migration after #469 merges. Renumbering #477 now would make its migration depend on a revision that is not yet present in main.
    
Also addressed the two follow-ups:
    - Added service-layer guards so only system categories/category groups can be hidden. Creation cannot set is_hidden, while harmless is_hidden=false updates on custom records remain allowed.
    - Documented the scope in the Categories UI: hidden defaults are removed from category pickers, while existing transactions, budgets, dashboards, and reports continue using them.
    
The is_ignored transaction-response handling is intentional: it fixes the adjacent issue where the transaction API dropped the category ignore flag, which is required for the existing report filtering to work correctly.
    
Verification:
    - Backend: 2,358 passed, 7 skipped
    - Frontend: 51 passed
    - Ruff, TypeScript build, locale parity, and browser QA passed
    - Alembic reports a single 066 head

tassionoronha:
Thanks for the thorough turnaround @RomuloGatto, and for the point-by-point writeup. Re-tested the branch locally and confirming:

- **Transaction / recurring / rule editors:** fixed. The picker now renders the assigned category with the Hidden badge instead of falling back to "Select category...".
- **`is_system` guard:** confirmed. `update_category` and `update_group` reject `is_hidden=true` on user records, creation ignores the field, and `is_hidden=false` still passes through so nothing gets stuck invisible.
- **`is_ignored`:** understood, thanks for explaining. Agreed it belongs here, no action needed.

**On the scope note.** You added copy saying hidden defaults stay in "transactions, budgets, dashboards, and reports". Dashboards and reports hold up. Budgets and rules don't yet, and they degrade in a way that looks broken:

1. **Budgets** (`budgets.tsx:119`): `getCategoryDisplay` falls back to `<span>{categoryId}</span>`. I put a budget on Educação, hid the category, and the row rendered `52103856-1b6f-41cb-863d-8d3b59d16817` with no icon. `BudgetProgress` already returns `category_name` / `category_icon` / `category_color`, so this should be a small swap.

2. **Rules list** (`rules.tsx:100`, `:289`): `actionSummary` falls back to the generic "Category" label and the category sort resolves to `''`. I hid Doações and the "Amigos do bem" rule went from `→ Doações` to `→ Category` while its neighbours kept their names. `allCategoriesList` is already fetched on that page, it just needs passing through, plus relaxing the `enabled` gate so it loads for the list and not only the dialog. `transaction-dialog.tsx:1234` has the same fallback.

Neat illustration of the gap: with the Edit Rule dialog open, the dialog correctly shows "Doações (Hidden)" while the list behind it still shows "→ Category". The fix you built is the right one, it just needs to reach the read-only summaries too.

**On migration ordering:** agreed, let's land this as 066 and rebase #469 to 067. Heads up that #469 is showing MERGEABLE again as of today, so whoever touches it next should renumber before it goes in.

One last small thing: `invalidateAll` invalidates `['category-groups']`, but rules/recurring/transactions/dashboard key groups as `['categoryGroups']`, so a newly hidden group lingers on those pages until a refetch.

RomuloGatto:
@tassionoronha closed the remaining display and cache gaps.

- Budget rows now resolve category name, icon, and color from the existing /budgets/comparison payload, so hidden assignments no longer degrade to UUIDs.
- Rule summaries and category sorting use the full management catalog for existing references.
- The transaction “add to rule” grouping now resolves existing hidden category and group names from the same display-only catalog.
- Ordinary category pickers still receive only visible categories. Hidden records are used solely to represent existing assignments and are never added as normal selectable options.
- Category mutations now invalidate categories, budget comparison metadata, and both categoryGroups / category-groups query-key conventions.

tassionoronha:
Hey @RomuloGatto, last round all checks out. Nice work.

I pushed three commits to your branch. 

The main one: hiding acategory took it out of the pickers but not out of use, the seeded rules kept assigning it, so the next sync filed transactions right back in. Every default category has a seeded rule, so it hit all of them. Now the engine skips hidden categories, and hiding one with rules asks if you want to turn them off. 

Merging it now 💪🏼

PR #480, opened 2026-07-31
fix: decode OFX 2.x/XML imports without a legacy SGML header

### LINKED ISSUE #479: [Bug] OFX import crashes with UnicodeDecodeError on OFX 2.x/XML exports containing non-ASCII characters
## Description

Importing an OFX file crashes with `UnicodeDecodeError` whenever the file is in the newer **OFX 2.0/XML** format (as opposed to legacy OFX 1.0 SGML) *and* contains a non-ASCII character (e.g. an accented letter) anywhere in a transaction memo or payee. Encountered with Erste Bank (Austria)'s "MS Money Sunset Deluxe" export preset; likely affects any bank/export tool that produces headerless OFX 2.x/XML.

## Steps to Reproduce

1. Save the following as `repro.ofx` — a minimal OFX 2.0/XML file (no legacy SGML header block, UTF-8 encoded, one non-ASCII character in the `MEMO`):

   ```xml
   <?xml version="1.0" encoding="utf-8" ?><?OFX OFXHEADER="200" VERSION="202" SECURITY="NONE" OLDFILEUID="NONE" NEWFILEUID="NONE"?><OFX><BANKMSGSRSV1><STMTTRNRS><STMTRS><CURDEF>EUR</CURDEF><BANKACCTFROM><BANKID>1</BANKID><ACCTID>1</ACCTID><ACCTTYPE>CHECKING</ACCTTYPE></BANKACCTFROM><BANKTRANLIST><DTSTART>20260101</DTSTART><DTEND>20260131</DTEND><STMTTRN><TRNTYPE>DEBIT</TRNTYPE><DTPOSTED>20260115</DTPOSTED><TRNAMT>-10.00</TRNAMT><FITID>1</FITID><MEMO>1 Aufladung(en) für Konto</MEMO></STMTTRN></BANKTRANLIST></STMTRS></STMTTRNRS></BANKMSGSRSV1></OFX>
   ```

2. Go to Accounts → Import → upload `repro.ofx`.
3. The import preview fails instead of showing the parsed transaction.

This matches real-world exports byte-for-byte in shape — e.g. Erste Bank Austria's online banking produces exactly this structure (no SGML header, `<?xml ... encoding="utf-8" ?>` prolog) under its "MS Money Sunset Deluxe" format option. Their "MS Money 2000" option produces the older SGML format instead, which imports fine today.

It can also be reproduced directly against the `ofxparse` library, without going through Securo's import endpoint at all — confirming this is an upstream bug rather than something in Securo's own parsing:

```python
from ofxparse import OfxParser
import io

xml = (
    '<?xml version="1.0" encoding="utf-8" ?><?OFX OFXHEADER="200" VERSION="202" '
    'SECURITY="NONE" OLDFILEUID="NONE" NEWFILEUID="NONE"?>'
    '<OFX><BANKMSGSRSV1><STMTTRNRS><STMTRS><CURDEF>EUR</CURDEF>'
    '<BANKACCTFROM><BANKID>1</BANKID><ACCTID>1</ACCTID><ACCTTYPE>CHECKING</ACCTTYPE></BANKACCTFROM>'
    '<BANKTRANLIST><DTSTART>20260101</DTSTART><DTEND>20260131</DTEND>'
    '<STMTTRN><TRNTYPE>DEBIT</TRNTYPE><DTPOSTED>20260115</DTPOSTED>'
    '<TRNAMT>-10.00</TRNAMT><FITID>1</FITID><MEMO>1 Aufladung(en) für Konto</MEMO></STMTTRN>'
    '</BANKTRANLIST></STMTRS></STMTTRNRS></BANKMSGSRSV1></OFX>'
)
OfxParser.parse(io.BytesIO(xml.encode("utf-8")))
# UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 468: ordinal not in range(128)
```

## Expected Behavior

The file should import successfully, with non-ASCII characters decoded correctly.

## Actual Behavior

The preview endpoint raises:

```
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 2938: ordinal not in range(128)
```

Full traceback:

```
Traceback (most recent call last):
  File "/app/app/api/import_transactions.py", line 60, in preview_import
    transactions = import_service.parse_ofx(content)
  File "/app/app/services/import_service.py", line 96, in parse_ofx
    ofx = OfxParser.parse(io.BytesIO(content))
  File "/usr/local/lib/python3.13/site-packages/ofxparse/ofxparse.py", line 396, in parse
    ofx_file = OfxPreprocessedFile(file_handle)
  File "/usr/local/lib/python3.13/site-packages/ofxparse/ofxparse.py", line 160, in __init__
    ofx_string = self.fh.read()
  File "<frozen codecs>", line 510, in read
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 2938: ordinal not in range(128)
```

The entire import is aborted — not just the transaction containing the offending character.

## Root cause (traced in code)

The bug lives in the `ofxparse` library (pinned as `ofxparse>=0.21` in `backend/pyproject.toml`), not in Securo's own parsing logic:

- `OfxFile.read_headers()` (`ofxparse/ofxparse.py:87-97`) looks for the legacy colon-delimited header block (`OFXHEADER:100`, `ENCODING:...`, `CHARSET:...`, etc.) only in the bytes *before the file's first `<`*.
- OFX 2.x/XML files have no such block — they start immediately with `<?xml ...?>` — so `read_headers()` finds nothing and `self.headers` ends up empty.
- `handle_encoding()` (`ofxparse/ofxparse.py:100-135`) then sees no `ENCODING` header and silently decodes the entire body as **ASCII**, which raises `UnicodeDecodeError` the moment it hits a non-ASCII byte.

This is a confirmed, still-open upstream issue: [jseutter/ofxparse#133](https://github.com/jseutter/ofxparse/issues/133) (open since 2017) — `ofxparse` never reads the `encoding="..."` attribute out of the XML prolog itself.

I also evaluated the community fork `ofxparse2` (published on PyPI, importable as `ofxparse`, otherwise a drop-in replacement), which specifically advertises a fix for this. It doesn't actually fix it correctly: its `extract_xml_encoding_from_file()` reads from the file handle's *current* position, which by the time it's called has already been advanced past 10KB by `read_headers()`'s own read — so it reads garbage from the middle of the document, never finds the `encoding="..."` attribute, and falls back to ASCII decoding **with `errors="ignore"`**. That doesn't crash, but it silently **drops every non-ASCII character** from memos/payees rather than decoding them — worse than the crash, since it corrupts transaction data with no error raised. So switching to that fork is not a safe fix either.

## Suggested Fix

`parse_ofx()` in `backend/app/services/import_service.py` already does OFX preprocessing before handing content to `ofxparse` (it synthesizes FITIDs for Brazilian banks whose exports omit them). The same preprocessing stage can detect a headerless OFX 2.x/XML document and prepend a synthetic legacy SGML header declaring the correct `ENCODING`/`CHARSET` (matching whichever encoding the content was actually decoded with — UTF-8 or Latin-1), routing the file through `ofxparse`'s already-correct SGML decode path instead of its broken auto-detection path.

## Environment

- Deployment: Docker (self-hosted)
- Bank/export: Erste Bank (Austria), OFX 2.0/XML export ("MS Money Sunset Deluxe" preset)
- `ofxparse` version: 0.21 (current pin in `backend/pyproject.toml`)

## Additional Context

This likely isn't Erste-specific — any bank whose OFX export tool emits OFX 2.x/XML without the legacy SGML header block, combined with any non-ASCII character in transaction text (accented names, non-English memos, etc.), will hit the same crash. I'll follow up with a PR shortly after this issue.

Jyasapara:
## What

`ofxparse` (pinned `>=0.21` in `backend/pyproject.toml`) only detects a file's encoding from a legacy colon-delimited SGML header block preceding the first `<`. OFX 2.x/XML exports — e.g. Erste Bank's "MS Money Sunset Deluxe" preset — have no such block, so `ofxparse` silently assumes ASCII and crashes with `UnicodeDecodeError` on the first non-ASCII byte (upstream: [jseutter/ofxparse#133](https://github.com/jseutter/ofxparse/issues/133), open since 2017).

This PR splits the OFX preprocessing pipeline in `backend/app/services/import_service.py` into single-purpose functions:

- `_decode_ofx_bytes` — trial-decodes UTF-8, falling back to Latin-1 (unchanged behavior, just extracted)
- `_patch_empty_fitids` — the existing Banco do Brasil empty-FITID synthesis, unchanged, just extracted from the old combined `_preprocess_ofx_for_empty_fitid`
- `_ensure_ofx_sgml_header` — **new**: detects a headerless OFX 2.x/XML document and prepends a synthetic SGML header declaring whichever encoding the content was actually decoded with, routing the file through `ofxparse`'s already-correct SGML decode path instead of its broken auto-detection
- `_preprocess_ofx` — orchestrates the three above

## Why

Importing an OFX 2.0/XML statement containing any non-ASCII character (accented names, umlauts, etc.) currently aborts the entire import with an unhandled `UnicodeDecodeError`. See #479 for the full root-cause writeup, including why I didn't just switch to the community fork `ofxparse2` — it "fixes" the crash but has its own bug (reads the encoding hint from the wrong file offset) that causes it to silently drop non-ASCII characters instead of decoding them, which is worse than a loud crash for financial data.

## How to Test

1. `cd backend && pytest tests/test_import_service.py -k ofx -v` — 9/9 pass (6 existing + 3 new: `test_parse_ofx_xml_format_without_sgml_header`, `test_parse_ofx_xml_format_latin1`, `test_parse_ofx_xml_format_ascii_only`)
2. Full backend suite: `pytest` — all pass, no regressions
3. Manual repro (also in #479): save the minimal OFX 2.0/XML snippet from the issue as `repro.ofx`, upload via Accounts → Import — previously crashed, now imports successfully with the umlaut decoded correctly
4. Verified end-to-end against the real-world Erste Bank OFX 2.0/XML export that originally triggered this — all 336 transactions parse, non-ASCII memos decode correctly

## Related Issue

Closes #479

## Checklist

- [ ] Frontend lints clean (`npm run lint`) — N/A, backend-only change
- [ ] Frontend builds (`npm run build`) — N/A, backend-only change
- [ ] Translations updated (if user-facing strings changed) — N/A, no user-facing strings changed
- [x] For a large or core change, I discussed it first (issue or Discord) so we could align before the code — filed #479 first with root-cause analysis before opening this PR

### DIFF
--- a/backend/app/services/import_service.py
+++ b/backend/app/services/import_service.py
@@ -89,20 +89,27 @@ def _patch_empty_fitids(text: str) -> str:
 def _ensure_ofx_sgml_header(text: str, encoding: str) -> str:
     """Prepend a legacy OFX 1.x SGML header for OFX 2.x files that omit it.
 
-    OFX 2.x is plain XML: an <?xml encoding="..." ?> prolog immediately
-    followed by <OFX>, with no colon-delimited SGML header block (e.g. Erste
-    Bank's "MS Money Sunset Deluxe" export). ofxparse only looks for
-    encoding hints in the bytes preceding the file's first "<"; when that's
-    empty it silently assumes ASCII and crashes on any non-ASCII byte.
-    Prepending a synthetic SGML header routes the file through ofxparse's
-    existing, correctly-working SGML decode path instead of its broken
-    auto-detection (see https://github.com/jseutter/ofxparse/issues/133).
+    OFX 2.x is plain XML that goes straight into its first tag, with no
+    colon-delimited SGML header block (e.g. Erste Bank's "MS Money Sunset
+    Deluxe" export). ofxparse only looks for encoding hints in the bytes
+    preceding the file's first "<"; when that's empty it silently assumes
+    ASCII and crashes on any non-ASCII byte. Prepending a synthetic SGML
+    header routes the file through ofxparse's existing, correctly-working
+    SGML decode path instead of its broken auto-detection (see
+    https://github.com/jseutter/ofxparse/issues/133).
+
+    The trigger is therefore "nothing precedes the first tag", not "starts
+    with <?xml": the XML declaration is optional in XML 1.0, so an OFX 2.x
+    file may open with just its <?OFX ... ?> instruction, or with <OFX>
+    itself, and those hit the same ofxparse bug. Anything else already has a
+    legacy header, and a file with no tag at all is left for ofxparse to
+    reject on its own terms.
 
     `encoding` must match whatever the caller will re-encode `text` with,
     so the declared header and the actual bytes stay consistent.
     """
-    stripped = text.lstrip("\ufeff \t\r\n")
-    if not stripped.startswith("<?xml"):
+    preamble, first_tag, _ = text.lstrip("\ufeff \t\r\n").partition("<")
+    if not first_tag or preamble.strip():
         return text
     if encoding == "latin-1":
         enc_lines = "ENCODING:USASCII\r\nCHARSET:8859-1\r\n"
--- a/backend/app/services/import_service.py
+++ b/backend/app/services/import_service.py
@@ -38,7 +38,18 @@ _OFX_BALANCE_ROW_DESCRIPTIONS = (
 )
 
 
-def _preprocess_ofx_for_empty_fitid(content: bytes) -> bytes:
+def _decode_ofx_bytes(content: bytes) -> tuple[str, str]:
+    """Decode OFX content to text, trying UTF-8 then falling back to Latin-1.
+
+    Returns (text, encoding) so callers can re-encode consistently later.
+    """
+    try:
+        return content.decode("utf-8"), "utf-8"
+    except UnicodeDecodeError:
+        return content.decode("latin-1"), "latin-1"
+
+
+def _patch_empty_fitids(text: str) -> str:
     """Synthesize a FITID for STMTTRN blocks that have an empty/missing one.
 
     Banco do Brasil (and a few other Brazilian banks) emit balance-summary
@@ -47,13 +58,6 @@ def _preprocess_ofx_for_empty_fitid(content: bytes) -> bytes:
     each affected block with a deterministic synthetic FITID so parsing
     succeeds; balance rows are filtered out later by description.
     """
-    try:
-        text = content.decode("utf-8")
-        original_encoding = "utf-8"
-    except UnicodeDecodeError:
-        text = content.decode("latin-1")
-        original_encoding = "latin-1"
-
     def _replace(match: re.Match) -> str:
         block = match.group(0)
         fitid_match = re.search(r"<FITID>([^<\r\n]*)", block, re.IGNORECASE)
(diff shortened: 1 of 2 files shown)


tassionoronha:
Great research on this, thanks. Pushed one commit: the XML declaration is optional, so OFX 2.x files opening on <?OFX ?> or <OFX> hit the same crash. Now keyed on the missing header instead.

PR #482, opened 2026-07-31
fix: prevent date field self-shadowing in AssetTransactionUpdate

### LINKED ISSUE #481: [Bug] PATCH /api/assets/transactions/{tx_id} rejects any date value
## Description

`AssetTransactionUpdate` (in `backend/app/schemas/asset.py`) defines the field `date: Optional[date] = None`. Because the field is named `date` and its type annotation also references the name `date`, Python binds the default value `None` to the class attribute `date` before evaluating the `Optional[date]` annotation. The annotation lookup then resolves to the just-bound `None` instead of `datetime.date`, so Pydantic builds the field as "must be `None`". As a result, the PATCH endpoint for updating an asset transaction can never actually change (or even accept) a `date` value — any request that includes one is rejected with a 422.
The UI by default sends the transaction date with a PATCH request for a transaction that has a date, even if it didn't change, blocking updates to other fields as well.

## Steps to Reproduce

1. Create an asset transaction:
   ```
   POST /api/assets/{asset_id}/transactions
   { "kind": "buy", "quantity": 10, "price": 20, "date": "2026-01-01" }
   ```
2. Fetch the transaction's `id` via `GET /api/assets/{asset_id}/transactions`.
3. Attempt to update it, including a `date` field:
   ```
   PATCH /api/assets/transactions/{tx_id}
   { "kind": "buy", "quantity": 119, "price": 43.47, "fee": 0, "date": "2025-04-15" }
   ```

## Expected Behavior

The request should succeed (200) and update the transaction's date to `2025-04-15`.

## Actual Behavior

The request fails with `422 Unprocessable Entity`:
```
date
  Input should be None [type=none_required, input_value='2025-04-15', input_type=str]
```
Any non-null `date` value in the PATCH payload is rejected, so a transaction's date can never be updated through this endpoint.

## Environment

- Deployment: Manual (local dev)
- Backend: FastAPI + Pydantic v2 (Python 3.13)
- Endpoint: `PATCH /api/assets/transactions/{tx_id}` (`backend/app/api/assets.py`)
- Schema: `AssetTransactionUpdate` (`backend/app/schemas/asset.py`)

## Screenshots

N/A (API-level bug, no UI screenshots)

## Additional Context

Root cause is a Python name-shadowing quirk: for an annotated class attribute `date: Optional[date] = None`, the default value is assigned to the name `date` in the class namespace before the `Optional[date]` annotation expression is evaluated, so the annotation resolves to `None` rather than the imported `datetime.date` type. This only affects fields whose name collides with the type name used in their own annotation *and* have a default value on the same line (e.g. `AssetTransactionUpdate.date`); sibling fields like `AssetTransactionCreate.date: date` (no default) are unaffected.

Jyasapara:
## What

Fixes `AssetTransactionUpdate` (`backend/app/schemas/asset.py`) so `PATCH /api/assets/transactions/{tx_id}` can actually accept a `date` value, and adds a regression test covering a full-field transaction update.

## Why

The field `date: Optional[date] = None` shadowed itself: Python binds the default `None` to the class attribute `date` before evaluating the `Optional[date]` annotation, so the annotation resolved to `None` instead of `datetime.date`. Pydantic then required the field to always be `None`, meaning any PATCH request that included a `date` was rejected with a 422 — transaction dates could never be updated through this endpoint. Aliasing the import (`from datetime import date as _date`) and using `_date` throughout the schema file avoids the name collision.

## How to Test

1. Run the backend test suite: `pytest backend/tests/test_asset_transactions_api.py -v`
2. Manually:
   - `POST /api/assets/{asset_id}/transactions` with `{"kind": "buy", "quantity": 10, "price": 20, "date": "2026-01-01"}`
   - `PATCH /api/assets/transactions/{tx_id}` with `{"kind": "buy", "quantity": 119, "price": 43.47, "fee": 0, "date": "2025-04-15"}`
   - Confirm the response is `200` and the transaction's date is updated to `2025-04-15` (previously returned `422`).

## Related Issue

Closes #481

## Checklist

- [ ] Frontend lints clean (`npm run lint`) - N/A, backend-only change
- [ ] Frontend builds (`npm run build`) - N/A, backend-only change
- [ ] Translations updated (if user-facing strings changed) - N/A, no user-facing strings changed
- [x] For a large or core change, I discussed it first (issue or Discord) so we could align before the code (see [CONTRIBUTING](../CONTRIBUTING.md#before-large-or-core-changes)) - filed #481 first with root-cause analysis before opening this PR

### DIFF
--- a/backend/app/schemas/asset.py
+++ b/backend/app/schemas/asset.py
@@ -1,5 +1,5 @@
 import uuid
-from datetime import date, datetime
+from datetime import date as _date, datetime
 from decimal import Decimal
 from typing import Optional
 
@@ -12,15 +12,15 @@ class AssetCreate(BaseModel):
     currency: str = "USD"
     units: Optional[Decimal] = None
     valuation_method: str = "manual"
-    purchase_date: Optional[date] = None
+    purchase_date: Optional[_date] = None
     purchase_price: Optional[Decimal] = None
-    sell_date: Optional[date] = None
+    sell_date: Optional[_date] = None
     sell_price: Optional[Decimal] = None
     current_value: Optional[Decimal] = None  # convenience: creates initial AssetValue
     growth_type: Optional[str] = None
     growth_rate: Optional[Decimal] = None
     growth_frequency: Optional[str] = None
-    growth_start_date: Optional[date] = None
+    growth_start_date: Optional[_date] = None
     is_archived: bool = False
     position: int = 0
     group_id: Optional[uuid.UUID] = None
@@ -28,7 +28,7 @@ class AssetCreate(BaseModel):
     # fetches the live quote on create and seeds the first AssetValue.
     ticker: Optional[str] = None
     ticker_exchange: Optional[str] = None
-    maturity_date: Optional[date] = None
+    maturity_date: Optional[_date] = None
     # Per-unit price for the opening buy of a market-priced holding (preço
     # médio model, consistent with the transaction ledger). When omitted, the
     # service seeds the buy at the live quote ("bought at market now").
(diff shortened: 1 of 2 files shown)


tassionoronha:
Hey @Jyasapara! Welcome here and thanks for this PR. LGTM!

PR #485, opened 2026-08-02
feat: add activity mode to the cash-flow transaction calendar

### LINKED ISSUE #484: [Feature] Activity mode for the cash-flow transaction calendar
## Problem

The cash-flow calendar from #401 is built around the running ending balance, so it can't answer "how much came in and went out on this day?". You have to open a day and add the rows up by hand.

The balance is also a poor proxy for activity: a day with 3,000 in and 3,000 out looks identical to a day with no movement.

## Proposed Solution

A `Balance | Activity` toggle in the calendar header, independent from `Compact | Detailed`. Balance stays the default and behaves exactly as today.

In Activity mode the calendar shows daily income and expenses instead of the running balance, keeping settled amounts visually separate from projected recurring occurrences. Transfers keep their rows but don't count as income or expenses.

Backend-wise this needs `actual_*` and `projected_*` buckets on the calendar day payload, with the existing combined totals kept as the sum of both. No migrations, no new endpoints.

## Alternatives Considered

- A separate Activity tab. Duplicates month/account/day state and loses your place when switching.
- Reusing `Compact | Detailed`. Density and metric are independent, so overloading it removes a useful combination.
- Deriving it client-side. Would re-implement the transfer, ignored-category and FX rules the service already applies.

## Additional Context

I have a working implementation (backend, UI, tests, all 10 locales) I can open as a PR. Two things I'd like your call on first:

1. Preference in `localStorage` (like the density toggle) or in the URL (like the month/account filters)?
2. Should the income/expense dot markers stay visible in Activity mode? They're redundant with the amounts, but keeping them means a day reads the same in both modes.

DannyFeliz:
## What

Adds a `Balance | Activity` toggle to the calendar from #401, independent from `Compact | Detailed`. Balance is the default and is unchanged.

In Activity mode:

- The trend strip becomes daily bars, with actual income above the axis, expenses below, and projections as dashed violet outlines. Columns are selectable by mouse and keyboard.
- Day cells show real income/expense amounts (or "No movements") instead of the running balance, with projections on their own violet line.
- The day panel leads with the actual net, then income/expense totals, plus a projection block when relevant.
- Transfers keep their marker and rows but never count as income or expenses.

Backend: `TransactionCalendarDay` gains `actual_*` and `projected_*` buckets. The existing `income`/`expense`/`transfer_net` stay as the sum of both, so current consumers are unaffected. No migrations, no new endpoints.

The preference persists in `localStorage`. Switching modes never changes the selected day, month, or filters.

**First commit is a separate pre-existing bug fix.** The day cell's selection ring is drawn outside the element, so `overflow-hidden` clipped it on edge rows, and selecting a day in the last row hid the calendar's bottom border. It's isolated so it can be reviewed or dropped independently.

## Screenshots
CleanShot 2026-08-02 at 14 17 38
CleanShot 2026-08-02 at 14 17 22

## Why

Every surface of the calendar answers "what will my balance be?". The equally common question, "how much came in and went out?", meant opening each day and adding rows up by hand, easy to miscount since projections sit in the same list as settled transactions. See #484.

## How to Test

1. **Transactions → Calendar**, pick a month with both real transactions and recurring bills.
2. Toggle `Balance | Activity`. Balance should look exactly as before.
3. Day with a projection: shown as a dashed outline in the chart and on its own violet line, never mixed into the actual figures.
4. Day with a transfer: rows and marker remain, amount isn't counted as income/expense.
5. Month with no activity: chart shows an empty state, cells read "No movements".
6. Reload and the toggle persists. Switching modes doesn't move the selected day or filters.
7. Select chart columns with keyboard (Tab, then Enter/Space).
8. Repeat on mobile and dark mode.
9. Ring fix: select a day in the last row and in the bottom corners, and the bottom border stays visible.

**Worth your call:** projections with an ignored category are excluded from activity totals but still move the projected balance. That's asymmetric with ignored *actuals*, which don't affect the balance either. It's pre-existing behavior inherited from `_daily_balance_deltas_by_date`, and I'm happy to align them if you'd prefer.

## Related Issue

Closes #484

## Checklist

- [x] Translations updated, 9 new keys in all 10 locales
- [x] For a large or core change, I discussed it first. Opened #484 and happy to adjust before review


tassionoronha:
Thanks @DannyFeliz, really nice work. 

I pushed two commits. Backend: ignored projections still moved the projected balance, so the calendar promised a number that reverts once they post. Frontend: the repeated "No movements" is now a centred dash, and the projected totals follow the panel's row style instead of the dashed box.

PR #490, opened 2026-08-03
feat: add read-only Trading 212 connector

Sgubela:
## Summary
- Add a read-only, token-based Trading 212 connector on top of current upstream `main`.
- Sync one primary-currency brokerage cash account, current positions, deposits and withdrawals, dividends, and completed order fills.
- Preserve provider metadata and make repeated syncs idempotent; cash settlement rows are ignored rather than imported as duplicate spending.
- Preserve the upstream Accounts UI and masked-number behavior.

## Safety
- Credentials use Securo's existing encrypted connection-credential path.
- The connector accepts `<api-key>:<api-secret>` for live accounts and `demo:<api-key>:<api-secret>` for paper trading.
- It uses an explicit read-only endpoint allowlist and never places or cancels orders, modifies Pies, or creates exports.
- Reconnecting with credentials for a different Trading 212 account is rejected, as is creating the same Trading 212 connection twice in one workspace.
- Unknown cash-transaction and order-fill types fail closed instead of being silently misclassified.

## Migration
- Alembic revision `077` extends upstream revision `076` and keeps a single migration head.
- The migration is additive and schema-aware for installations that previously had equivalent local Trading 212 columns.
- Its downgrade is intentionally blocked because it cannot safely distinguish newly added schema from compatible pre-existing local schema; restore a backup instead of downgrading.

## Live validation
- Backed up and deployed to a real Securo installation, upgrading PostgreSQL from revision `068` to `077`.
- Verified account creation, cash and holdings import, transactions, dividends, completed orders, and repeat sync behavior.
- Fixed three issues found during live validation: Alembic async transaction ownership, a duplicate account-summary request that caused HTTP 429, and documented Trading 212 cash-interest transaction types.
- Final scheduled sync completed successfully with one account-summary request and the connection remaining active.

## Verification
- Affected Trading 212, migration, connection, and account suite: `253 passed`.
- Ruff and `ty`: passed.
- Independent reviews of all three live-discovered fixes: passed with no high, medium, or low findings.
- GitHub Migration Chain, Frontend Checks, and Helm Checks: passed.
- The current Backend Tests failure is the unrelated upstream month-boundary test in `tests/test_dashboard_api.py`; the same failure affects other backend PRs and already has a dedicated fix in #739.
- The PR contains no deployment configuration, private configuration, or credentials.


Sgubela:
@tassionoronha Hello, I wasn't going to create this PR at all here; my agent mistakenly did it here instead of in my fork. So I moved this to a draft. I'm not sure if this is of interest for the project or for other people, but it is for me. But I decided to ask you anyway, what are your thoughts on trading platform integrations?

negue:
> Sync brokerage accounts, holdings, and order history; preserve metadata and ignore cash-settlement records.

Would the credit card cash part be an account ? meaning this goes to Accounts 
and the brokers to assets => portfolios ?

Sgubela:
> > Sync brokerage accounts, holdings, and order history; preserve metadata and ignore cash-settlement records.
> 
> Would the credit card cash part be an account ? meaning this goes to Accounts and the brokers to assets => portfolios ?

Yes, cash is an account.

Let me put screenshots from my running instance. 

Screenshot_20260826_190202
Screenshot_20260826_190100
Screenshot_20260826_190036

Here's Accounts page, Assets page, and Accounts section in left navigation bar.

negue:
Awesome , thank you for the screenshots!!

looking at the screenshots, great to see cash to buy stocks and the cash on the card in their own sections !! will definitely help instead of doing it all manually or splitting the csvs manually / code^^

I don't know if the API allows it (havent tried it yet), but is it possible to read

1. which cash is in different currencies
2. same for the cards side (I think there is also possible to have different currencies?)


is there anything I should keep in mind when trying it locally? happy to try it out

negue:
another idea - what do you think about having each Pie in their own Trading212: Pie name group ? 
if its even possible from the api as well

Sgubela:
> Awesome , thank you for the screenshots!!
> 
> looking at the screenshots, great to see cash to buy stocks and the cash on the card in their own sections !! will definitely help instead of doing it all manually or splitting the csvs manually / code^^
> 
> I don't know if the API allows it (havent tried it yet), but is it possible to read
> 
>     1. which cash is in different currencies
> 
>     2. same for the cards side (I think there is also possible to have different currencies?)
> 
> 
> is there anything I should keep in mind when trying it locally? happy to try it out

Thanks! Right now the connector imports one cash account in the primary currency. Trading 212’s API doesn’t expose separate currency balances or Trading 212 Card data.

For local testing, use a read-only key with account, portfolio and history access. Use a backup ordisposable database because migration 077 can’t be downgraded.

Pies are a nice idea, but their API is deprecated and the mapping isn’t straightforward, so I’d keep that as a separate follow-up.

PR #493, opened 2026-08-04
fix: align calendar weekday headers with the dates below them

### LINKED ISSUE #492: [Bug] Calendar weekday headers are shifted one day in timezones behind UTC
## Description

In **Transactions → Calendar**, the weekday header row is off by one day for users in timezones behind UTC.

For August 2026 the header row reads `SAT SUN MON TUE WED THU FRI` instead of `SUN MON TUE WED THU FRI SAT`, so the 1st appears to fall on a Friday when it was actually a Saturday.

The day cells themselves are in the right places, the grid is a correct Sunday-start month grid, only the labels above them are shifted, which makes every date in the month look like it landed on the wrong weekday.

## Steps to reproduce

1. Set the machine timezone to one behind UTC (e.g. `America/Santo_Domingo`, UTC−4)
2. Go to Transactions and switch to the Calendar view
3. Compare the weekday headers against the dates in the grid

## Expected

Headers read `SUN → SAT` and line up with the dates below them.

Image

## Actual

Headers are shifted one column back, so August 1 2026 (a Saturday) sits under `FRI`.

Image

## Notes

Not reproducible in UTC or in timezones ahead of UTC, the further behind UTC the client is, the more likely it shows.

DannyFeliz:
## What

Fixes the weekday header row in the transaction calendar so the labels line up with the dates underneath them.

The header labels were built from a UTC-anchored reference week (`Date.UTC(2024, 0, 7)`, a Sunday) but formatted with `toLocaleDateString` in the viewer's own timezone. For anyone behind UTC that instant reads as the previous day, so the entire row shifted one column: at UTC−4 it rendered `SAT SUN MON TUE WED THU FRI` instead of `SUN MON TUE WED THU FRI SAT`.

The builder now formats in UTC as well, and moved to `weekdayShortLabels()` in `frontend/src/lib/date-utils.ts` so a timezone regression test can cover it.

## Why

With the labels off by one, every date in the month appeared to fall on the wrong weekday, August 2026 showed the 1st under `FRI` even though it was a Saturday. The day cells themselves were never wrong: the backend grid in `transaction_calendar_service.py` builds a correct Sunday-start month grid, so this was purely a display bug in the header, but it made the whole calendar untrustworthy at a glance.

The impact scales with distance from UTC, invisible in UTC and ahead of it, wrong for every user in the Americas.

## How to Test

1. Set your machine timezone to one behind UTC (e.g. `America/Santo_Domingo`, UTC−4)
2. Run the frontend and open **Transactions → Calendar**
3. The header row reads `SUN → SAT`, and August 2026 shows the 1st in the Saturday column
4. Switch the app language and confirm the labels are still localized and still Sunday-first
5. `cd frontend && npm test`, `weekdayShortLabels` is asserted under both a west-of-UTC (`America/Santo_Domingo`) and an ahead-of-UTC (`Pacific/Kiritimati`) timezone

## Related Issue

Closes #492

## Checklist

- [ ] Backend tests pass (`pytest`) - not run; no backend files are touched by this change
- [x] Frontend lints clean (`npm run lint`) - 0 errors (41 pre-existing warnings elsewhere in the repo, none from these files)
- [x] Translations updated (if user-facing strings changed) - n/a, the labels come from `toLocaleDateString` and stay localized
- [x] For a large or core change, I discussed it first (issue or Discord) - filed as #492

### DIFF
--- a/frontend/src/components/transaction-calendar-view.tsx
+++ b/frontend/src/components/transaction-calendar-view.tsx
@@ -6,6 +6,7 @@ import type { TransactionCalendarDay, TransactionCalendarItem, TransactionCalend
 import { Skeleton } from '@/components/ui/skeleton'
 import { CategoryIcon } from '@/components/category-icon'
 import { activityChartData, dayActivity } from '@/lib/calendar-activity'
+import { weekdayShortLabels } from '@/lib/date-utils'
 import { cn } from '@/lib/utils'
 
 function parseLocalDate(value: string) {
@@ -103,13 +104,7 @@ export function TransactionCalendarView({
   }, [calendar, onSelectedDateChange, selectedDate])
 
   const selectedDay = calendar?.days.find((day) => day.date === selectedDate)
-  const weekDays = useMemo(
-    () => Array.from({ length: 7 }, (_, index) => {
-      const date = new Date(Date.UTC(2024, 0, 7 + index)) // Sunday-start reference week
-      return date.toLocaleDateString(dateLocale, { weekday: 'short' })
-    }),
-    [dateLocale],
-  )
+  const weekDays = useMemo(() => weekdayShortLabels(dateLocale), [dateLocale])
 
   if (isLoading) {
     return (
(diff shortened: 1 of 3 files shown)


tassionoronha:
Great work @DannyFeliz , LGTM! :)

PR #501, opened 2026-08-05
chore(deps): bump pytest to >=9.0 and pytest-asyncio to >=1.0

pjrm:
## What

Update dev testing dependencies to their latest major versions. pytest-asyncio 1.0 drops the <1.0 upper bound, and pytest moves to the 9.x line.

## Why

When running ty, there's an issue with pytest < 9 that threw the error "error[invalid-argument-type]: Argument to bound method `skip` is incorrect"

## How to Test

Run `mise backend:test`

## Related Issue

Closes #___ (or link the issue this addresses).

## Checklist

### DIFF
--- a/backend/pyproject.toml
+++ b/backend/pyproject.toml
@@ -36,8 +36,8 @@ dependencies = [
 
 [project.optional-dependencies]
 dev = [
-    "pytest>=7.4.0",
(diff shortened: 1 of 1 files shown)


tassionoronha:
Thanks for that @pjrm !

PR #506, opened 2026-08-05
feat: adjust recurring dates around weekends

### LINKED ISSUE #496: [Feature] Allow recurring transactions to move before or after weekends
## Problem

Recurring transactions are currently projected on their configured calendar date,
even when that date falls on a weekend.

In practice, some payments are moved to the last banking day before the weekend,
while others are processed on the next banking day.

For example, I have a monthly payment intended for the beginning of the month. Its
recent transaction dates were:

- 03.06.
- 06.07.
- 30.07. for the August payment

The existing matching tolerance can still associate these real transactions with
the recurrence, but the projected calendar may show the payment on the wrong day.
This can make the projected account balance look higher than it will actually be,
especially when a payment is taken before the weekend.

## Proposed Solution

Add an optional weekend adjustment to recurring transactions:

- No adjustment
- Move to the Friday before the weekend
- Move to the Monday after the weekend

This adjusted date should be used for projected calendar entries and automatically
generated transactions.

The existing matching tolerance should remain, since bank processing dates may
still differ slightly from the adjusted date.

## Alternatives Considered

I can choose a middle date that fits Securo's current matching window, but this only
helps reconciliation. It does not make the projected transaction date accurate.

I could also manually change recurring dates whenever a weekend occurs, but that
would require maintaining every recurring transaction month by month.

## Additional Context

This would be especially useful for rent, salaries, loan payments, account
transfers, and other transactions where the exact timing affects the projected
account balance.

Public-holiday support could be handled separately. A weekend-only option would
already cover the most common and predictable date shifts.

seraphblade2010:
## What

- Add per-recurrence weekend adjustment with `none`, `previous_friday`, and `next_monday`.
- Apply effective dates consistently to recurring projections, projected balances and cash flow, transaction-calendar entries, pending generation, generated transaction dates, and matching while preserving the nominal schedule.
- Expose the setting through the REST API, frontend create/edit form and types, translations, and MCP recurring lifecycle tools.
- Add Alembic revision `066` with a backward-compatible `none` default and strict persisted/API validation.

## Why

Recurring transactions scheduled on weekends can be processed on the preceding Friday or following Monday. Using only the nominal date makes calendar and balance projections inaccurate. This implements issue #496 without adding holiday calendars or changing the existing matching tolerance.

## Implementation Approach

- Keep `start_date`, `next_occurrence`, cadence advancement, and end-date evaluation nominal so schedules never drift.
- Derive an effective occurrence date through one shared weekend-adjustment helper.
- Scan projection boundaries with a bounded two-day nominal margin, then filter on effective dates so cross-month occurrences appear exactly once.
- Gate pending generation by the effective cutoff while advancing only the nominal occurrence.
- Persist the validated policy with a non-null column, server default, and check constraint.

## How to Test

1. Create or edit recurring transactions using each weekend policy and confirm the selected value persists.
2. Confirm an August 1 Saturday occurrence using `previous_friday` appears on July 31, not in August.
3. Confirm an August 2 Sunday occurrence using `next_monday` is not generated on Sunday and is generated on Monday.
4. Confirm generated transaction dates use the effective date while `next_occurrence` advances on the nominal cadence.
5. Exercise create, read, and update API paths, including invalid and explicit-null policy rejection.

## Validation

- Fork PR CI passed all three jobs: Backend Tests, Frontend Checks, and Helm Checks.
- Backend CI: 2,432 passed, 7 skipped.
- Frontend CI: 56 tests passed; lint and production build passed.
- Ruff passed on the backend changes.
- Alembic `065 -> 066 -> 065 -> 066` passed against disposable PostgreSQL, including default and constraint checks.
- Manually verified all three policies through the recurring create/edit UI.
- Manually regression-tested quarterly recurrence and exact recurring-badge linkage.

## Related Issue

Closes #496

## Checklist


seraphblade2010:
@tassionoronha I do not want to stress or anything but do you think this is ready to merge as is? Unsure if you are the right person to ask this but you seem to be the most active maintainer here :)

tassionoronha:
Hey @seraphblade2010, I still the only maintainer here for now(maybe it will change soon :), so sometimes it takes longer that I also wish to review, respond some issues and also develop some side tasks :) 

The changes LGTM, it's a nice improvement to extend our recurring system. :)
Thanks for that! ✨

Issue #505, opened 2026-08-05
[Feature] Payables/Receivables workflow


panga:
## Overview

A comprehensive payables/receivables workflow for personal finance. This issue tracks the full lifecycle: from registering upcoming bills as pending, through installment payments, to automatic settlement when bank syncs confirm the charge.

## Problem

For personal finance use cases like payables/receivables, there's no clean way to distinguish between a bill that's been settled and one that's still pending. All transactions are created as `posted`, and the transaction list has no status filter. On mobile, the table-based layout doesn't scale well for daily use (too many scrollbars). 

### 1. Mobile transaction cards

Replace the data table on small screens (<768px) with card-based rows grouped by date. Each card shows:
- Category icon + description
- Account with icon
- Amount (color-coded by type)
- Status indicators

Desktop table remains unchanged on md+ screens.

### 2. Pending/posted status

- Add `status` field (`posted`/`pending`) to transaction create and edit forms
- Default manual transactions to `posted` (matching existing behavior)
- Show status on synced transactions (disabled — status determined by sync)
- Pending indicator on mobile cards: subtle clock icon in the description row
- Status filter in transaction list (backend API + frontend filter bar + MCP)

### 3. Manual installments

Allow users to split a purchase into N installments manually, independent of credit-card bank sync.

- **Transaction form**: "Split into installments" section (count, total amount, frequency, first installment status)
- **Backend**: Endpoint to create installment series
- **Edit/delete scope**: "Apply to all" / "Apply to this and future" / "This one only"
- **Display**: Existing "1/N" badge (already works for synced installments)
- **Dedup**: Existing `installment_purchase_date` fingerprint for sync matching

### 4. Recurring generate-ahead as pending

Environment flag that makes auto-generated recurring transactions create as `pending` one period ahead.

- **Behavior**: Every `auto_generate` rule creates occurrences as `pending` one period ahead
- **Settlement**: Pending rows flip to `posted` only when a real synced/imported charge merges into them
- **Feature flag**: Exposed via API, surfaced in UI help text
- **No schema change**: ENV var only, no migration

### 5. Form input visibility

- Light mode: standardize `bg-card` on dialog form fields (currently only transaction dialog was fixed)
- Affects ~15 files across transfer dialog, group form, member form, import review, workspace settings, and more

## Additional Context

- ~10 years of historical data from MinhasFinancas/Mobilis/Organizee
- Family use case (multiple accounts, shared visibility)
[...]

PR #517, opened 2026-08-06
fix: white form field backgrounds in light mode

panga:
## What

Standardizes form input backgrounds in light mode. Raw form fields (`<input>`, `<select>`, `<textarea>`) inside dialogs and forms now use `bg-card` (white) instead of `bg-background`/`bg-transparent`, which rendered as gray-on-gray against the dialog backdrop.

Two layers of change:

- **Per-file fields (15 files)**: swapped `bg-background` → `bg-card` on 38 raw form fields across the dialog/form pages.
- **Shared components (systemic fix)**: changed the default background from `bg-transparent` to `bg-card` in `ui/input.tsx`, `ui/select.tsx`, and `category-select.tsx`, fixing the recurring and account dialogs whose fields sat on a gray `DialogContent` surface.

Dark mode is intentionally untouched: it already had good contrast (inputs are visually distinct from their container), so all `dark:bg-input/30` variants are preserved. The problem was specific to light mode.

## Why

In light mode the dialog backdrop is gray (`bg-background`, #F8F9FB). Fields that used `bg-transparent` (or the old `bg-background`) blended into that surface and were not visually distinct as input areas. Using `bg-card` (white) makes every form field read clearly as an editable control. Dark mode was already fine and is not touched by this change.

## How to Test

1. Run the app in light mode.
2. Open the New Transaction dialog and the New Recurring dialog.
3. Verify inputs, selects, and textareas render white and stand out from the gray dialog surface.
4. Open the Account form and confirm the same.
5. Switch to dark mode and confirm inputs remain visually distinct from their container (no regression).

### Recurring before
before

### Recurring after
after

## Related Issue

https://github.com/securo-finance/securo/issues/505

## Checklist

- [x] Backend tests pass (`pytest`) — no backend changes
- [x] Frontend lints clean (`npm run lint`) — 0 errors on changed files (only pre-existing warnings)
- [x] Frontend builds (`npm run build`) — `tsc -b` and `vite build` pass
- [ ] Translations updated (if user-facing strings changed) — no user-facing strings changed


tassionoronha:
Nice work @panga !
LGTM :)

PR #523, opened 2026-08-06
feat(recurring): add RECURRING_GENERATE_AHEAD flag

panga:
## What

Adds a `RECURRING_GENERATE_AHEAD` feature flag (env var, off by default) that makes `generate_pending` also materialize the next occurrence of `auto_generate` recurring transactions one period ahead, always as `pending`. 

## Why

Issue #505 (payables/receivables workflow) includes "recurring generate-ahead as pending": auto-generated bills should show the upcoming occurrence ahead of time without settling it.

- Pending rows never auto-flip to posted; they upgrade only when a real synced/imported charge merges into them (or the user edits them).
- `next_occurrence` stays at the next due date instead of jumping past the pending row, so the recurring list shows the date users actually care about.
- Flag exposed via `/api/info` and surfaced in the recurring form help text.

## How to Test

1. Set `RECURRING_GENERATE_AHEAD=true` in the root `.env`.
2. Recreate the backend/workers: `docker compose up -d --force-recreate backend celery-worker celery-beat`.
3. Confirm `/api/info` returns `"recurring_generate_ahead": true`.
4. With an `auto_generate` recurring rule, run generation (hourly worker or `python -m app.cli generate-recurring`).
5. Verify the next occurrence appears as `pending`, and `next_occurrence` in the UI stays at the next due date.

## Related Issue

Part of #505 (payables/receivables workflow, item 4: recurring generate-ahead as pending).

## Checklist

- [x] Backend tests pass (`pytest`) — 78 recurring-related tests
- [x] Frontend lints clean (`npm run lint`) — 0 errors
- [x] Translations updated (if user-facing strings changed) — 11 locale files


tassionoronha:
Nice work on the flag plumbing, it's clean.

Two bugs first: once sync merges into the ahead row its source flips to sync, so the next run writes a second placeholder. And nothing ever posts these rows, so users without bank sync lose them from the dashboard while balances still count them.

**Bigger thing though**: we now have three ways to say "this money is coming". Projections (#526), installment rows (#524), and pre-materialized rows here. I'd like to settle one rule first: projections are the default, rows only exist when the user commits to one, and pending keeps meaning "the bank said so", not "we forecast it".

Can we align on that before merging any of the three?

panga:
I've aligned with @tassionoronha on Discord about pending behavior. #526 will be split into multiple PRs to tackle the semantic behavior change on pending transactions. I will fix the bug of the duplicate placeholder on sync.

PR #524, opened 2026-08-06
feat: repeat manual transactions as installment series

panga:
## What

Implements the **Manual installments** section of issue #505: users can now create a single manual transaction and have it fanned out into N installments.

## Why

Supports the payables/receivables workflow for purchases paid in installments that don't come from credit-card bank sync.

Why not just use recurring? It is a distinct representation and mental model. 

## How to Test

1. Open the transaction form, enable **Repeat as installments**
2. Pick count (2–360) and frequency (monthly/weekly/quarterly/yearly)
3. Save → the UI shows the first parcel posted, the rest pending, dates advancing by the chosen cadence
4. Edit or delete any parcel of a multi-installment series → a scope prompt asks **this / this + future / all**
5. Status-only edits skip the prompt and apply to the single row

### New transaction
new

### Edit transaction
edit

### Delete transaction
delete

### List transaction
list

## Related Issue

Part of #505 (Payables/Receivables workflow).

## Checklist

- [x] Backend tests pass (`pytest` — 2447 passed)
- [x] Frontend lints clean (`npm run lint` — 0 errors, 42 pre-existing warnings)
- [x] Frontend builds (`npm run build` — success)
- [x] Translations updated (EN + PT-BR + 6 other locales for new user-facing strings)
- [x] Discussed first — tracked on issue #505 since inception


tassionoronha:
Nice feature @panga , but I found two blockers while testing locally.

Two purchases on the same card, same day, both 3x get treated as one series. I edited one with "All" and it overwrote the other. Needs a real series id.

Also, Workspace settings is a blank crash now, maybe from the useEffect refactor. 
Unrelated to installments, worth splitting out.

panga:
I will take a look now. It is hard to manage multiple active branches (I have 3 open PRs).

tassionoronha:
Hey @panga,

pushed two fixes on top : the scope prompt was gated on source, which breaks once sync absorbs a parcel, and the series total wasn't recomputed after an amount edit. Tested the same-day 3x case end to end, no cross-series damage anymore.

LGTM now, lets move on :) 
Well done!

PR #536, opened 2026-08-08
Feat/explicit destination amount

### LINKED ISSUE #529: [Feature] Allow explicit destination amount for cross-currency transfers between accounts
## Problem

Cross-currency transfers between accounts currently accept a source amount and an optional fx_rate. This might be undesirable when both transaction amounts are already known. In that case, the user would have to calculate fx_rate manually. That makes fx_rate the source of truth instead of the actual amounts that occurred on each account. This can introduce rounding concerns for ratios that cannot be represented exactly and is particularly awkward for custom currencies that may not follow the semantic of a market FX rate at all.

## Proposed Solution

Replace the optional fx_rate with an optional explicit destination amount.

## Alternatives Considered

It would be possible to add an explicit destination amount while keeping fx_rate, but that might lead to contradictions in case both values are defined.

Other option would be to create two transactions and link them afterward. However, it requires three operations instead of one.

Mapalmeira:
## What

Replace the optional fx_rate with an optional explicit destination amount.

## Why

Cross-currency transfers between accounts currently accept a source amount and an optional fx_rate. This might be undesirable when both transaction amounts are already known. In that case, the user would have to calculate fx_rate manually. That makes fx_rate the source of truth instead of the actual amounts that occurred on each account. This can introduce rounding concerns for ratios that cannot be represented exactly and is particularly awkward for custom currencies that may not follow the semantic of a market FX rate at all.

## How to Test

1. Create a transfer between accounts using different currencies.
2. Enter a source amount and an explicit destination amount; verify each account records exactly the entered amount.
3. Create another cross-currency transfer without a destination amount; verify the destination is calculated automatically.
4. Attempt to submit a zero or negative destination amount; verify validation prevents it.

## Related Issue

Closes #529 

## Checklist

- [x] Frontend lints clean (`npm run lint`) -> lint errors appeared, but not in files edited in this PR.
- [X] For a large or core change, I discussed it first (issue or Discord) so we could align before the code (see [CONTRIBUTING](../CONTRIBUTING.md#before-large-or-core-changes))


tassionoronha:
Thanks for this @Mapalmeira , really good change. 

I merged main in and pushed two small commits on top: exact rounding, a loud error for the old fx_rate, and keeping your amounts on edit.

PR #541, opened 2026-08-09
fix: truncate transaction notes description

luiserdef:
## What

Updated the transaction card to truncate the notes description when the text exceeds the available space.

## Why
The transaction title was already truncated correctly, but the notes description could overflow when it contained long text. 
This happens in Transactions Screen.

**Screenshots**

Before:
before

After:

after


tassionoronha:
Welcome here @luiserdef! 
Short and important fix :) 
LGTM!

PR #526, opened 2026-08-06
feat: improve account detail page for mobile and desktop

panga:
## What

Overhaul of the account detail page (mobile + desktop) so every number on the
page comes from the visible transaction list, and the pending/recurring split
is consistent across the app: sidebar, account detail cards, balance chart,
and dashboard.

### Pending / recurring split (backend)

- `get_accounts` / `get_account_summary` / `serialize_account`: non-credit-card
  accounts now exclude pending transactions from the displayed balance
  (`current_balance` = posted-only for manual accounts; provider balance for
  connected accounts). Credit cards keep pending in their balance (unchanged).
- `AccountSummary` gains `opening_balance` / `opening_balance_primary`,
  computed as `current_balance − Σ(posted txs ≥ date_from)`, so the frontend
  can seed the running-balance walk and match "Saldo projetado" to period end.
- Dashboard balance APIs (`get_summary`, `get_balance_history`) accept
  `include_pending=False`: pending drops out of monthly totals and balance
  history for non-CC accounts (connected accounts subtract their net pending,
  via the new `net_pending_by_account`). Credit cards keep pending.
- `ProjectedTransaction` gains `account_id`; the projected-transactions
  endpoint accepts `account_id`, `from`, `to` for per-account, range-scoped
  projections.
- Removed `previous_balance` (schema, service, tests) — the sidebar percentage
  feature it fed was removed because it was confusing and inconsistent.

### Account detail page (frontend)

- Non-CC accounts render virtual "Previsão" rows for non-materialized recurring
  projections (non-clickable, badge, not counted in income/expenses dialogs),
  merged with real transactions for a single running-balance walk.
- Balance cards: Saldo atual (posted-only), Saldo projetado (opening + net of
  all displayed rows incl. pending + projections), Entradas/Saídas computed
  from the displayed list instead of the backend summary.
- Balance chart colored by sign (green ≥ 0, red < 0) with a zero-split
  gradient; CC chart unchanged.
- Default date range is the full current month (was: today); date navigation
  chevrons moved into the header for mobile.
- Mobile: date-grouped transaction cards (`MobileTransactionRow` with
  `showPayee`), 44px tap targets, responsive 2/4-col card grid, `clamp()` font
  sizing, pending clock icon in amber (consistent with transaction list).
- Dashboard recent-transactions rows show the account name for projected
  entries (was: date); recurring badge sizing now matches the list.
- Sidebar: removed the month-over-month percentage; `-0,00` normalized to `0`.

### Tests

[...]

PR #570, opened 2026-08-12
refactor(frontend): share currency formatter

panga:
## What

- Centralize currency formatting in the shared frontend formatter.
- Replace duplicated page and component helpers with the shared utility.
- Normalize negative zero and fall back safely when a currency code is invalid.

## Why

The account-detail work exposed several copies of the same formatter. Keeping the cleanup separate makes the functional and layout reviews smaller and prevents formatting behavior from drifting between screens.

## How to Test

1. Run `cd frontend && ./node_modules/.bin/tsc --noEmit -p tsconfig.app.json`.
2. Run `cd frontend && ./node_modules/.bin/vitest run`.
3. Open screens that display monetary values and verify currencies render as before, including zero values.

## Related Issue

Extracted from [securo-finance/securo#526](https://github.com/securo-finance/securo/pull/526).

## Checklist


tassionoronha:
LGTM now @panga.

Just fixed a detail on the decimal part.
Thanks again for that! 🙌🏼

PR #576, opened 2026-08-13
Fix/csv import improvements

### LINKED ISSUE #566: [Bug] CSV import preview ignores category_name and shows matched categories as uncategorized
## Description

When a CSV contains a `category` column whose value exactly matches an existing category in the current workspace, the import review UI still displays **No category**.

The CSV parser correctly stores the field as `category_name`, and the final import resolves it through the workspace category name-to-ID map. However, the review table only uses `selected_category_id` or `suggested_category_id` as the select value, so a category supplied by the CSV is not visible during preview.

## Steps to Reproduce

1. Import the following CSV:

   ```csv
   date,description,amount,currency,category
   2026-08-01,Salary Payment,69000.00,USD,Salary
   ```

2. Open the import review table.
3. Observe the category selector for the transaction.

## Expected Behavior

The review table should resolve the CSV's `category_name` against categories in the current workspace and preselect `Salary`.

The transaction should also be treated as categorized by the review filters.

## Actual Behavior

The category selector displays **No category**, even though the CSV value exactly matches an existing category.

Users are led to select the category manually. The final backend import may still resolve the category correctly, which makes the preview inconsistent with the persisted result.

## Environment

- Import format: CSV
- Locale/category name: English
- UI: transaction import review

## Additional Context

A possible fix is to resolve an exact CSV category-name match to an ID during preview, preferably in the backend and scoped to the current workspace. The resolved ID could be exposed separately or used to populate the preview's selected/suggested category fields. This would keep the review UI, category filters, and final import consistent.

jorishaenseler15:
## What

- Added logic to `import_service.py` to map CSV category names to `suggested_category_id` during the import preview phase (resolves Issue #566). When mapping, hidden categories are ignored, and if a matching rule applies, the rule's category takes precedence over the CSV category.
- Added support for Swiss currency thousands separators (`'`) during CSV amount normalization.
- Refactored `parse_csv` to return strong types and use structured error codes instead of hardcoded English strings.
- Added detailed tracking of unparsed rows (line number, description, raw value, and error reason) during CSV parsing and displayed them via an unparsed rows modal dialog in the UI using proper locale translations.

## Why

- **Categories:** When importing CSVs, the parser extracted the category name but didn't resolve it to a category ID for the preview API. This caused the UI to show "No category" even when an exact match existed in the workspace. The logic ensures hidden categories aren't incorrectly mapped, and that user-defined rules still take precedence.
- **Swiss Separators:** CHF numbers often use `'` as thousands separators (e.g. `1'000.00`). If not normalized out, they failed to parse as valid Decimal numbers and were skipped during import, throwing off account balances.
- **Unparsed Warning:** When rows failed to parse due to dates or amounts, they were skipped silently, leaving the user with an incorrect balance. The unparsed rows details dialog makes these rows fully transparent and properly localized.

### 1. Test Category Preselection
1. Go to your Securo dashboard and ensure you have a category named exactly `Salary` in your current workspace (create one if it doesn't exist).
2. Save the following text as a `test_import.csv` file:
   ```csv
   date,description,amount,currency,category
   2026-08-01,Salary Payment,69000.00,USD,Salary
   ```
3. Navigate to **Transactions > Import** and upload the `test_import.csv` file.
4. During the preview step, verify that the category dropdown for the transaction is automatically pre-filled with `Salary` instead of saying "No category".

### 3. Test Swiss Currency Thousands Separator (')
1. Save the following text as `test_swiss_separator.csv`:
   ```csv
   date,description,amount,currency
   2026-08-01,Large Purchase,-1'050.45,CHF
   2026-08-02,Large Salary,5'400.00,CHF
   ```
2. Upload `test_swiss_separator.csv` via **Transactions > Import**.
3. Verify that both transactions parse correctly and display amounts of `1050.45` and `5400.00` in the preview without any parser errors.

### 4. Test Details Dialog for Unparsed CSV Rows
1. Save the following text as `test_invalid_rows.csv` (contains invalid date and amount rows):
   ```csv
   date,description,amount,currency
   2026-08-01,Valid transaction,-10.00,CHF
   invalid_date,Invalid Date Row,-10.00,CHF
   2026-08-03,Invalid Amount Row,invalid_amount,CHF
   ```
2. Upload `test_invalid_rows.csv` via **Transactions > Import**.
3. Verify that an amber warning alert is shown inside the CSV Options box: `"2 row(s) in this file could not be parsed and were skipped."` with a **"View details"** button.
4. Click **"View details"** and confirm that a modal dialog is displayed with:
   - Line 3: `Invalid Date Row`, value `invalid_date`, reason `Invalid Date Format`
   - Line 4: `Invalid Amount Row`, value `invalid_amount`, reason `Invalid Amount Format`
5. To verify the localization fixes:
   - Go to your account settings and change the language (e.g., to German or Portuguese).
   - Re-open the modal and verify that the `Description` column header is correctly translated (it should NOT say the literal string `import.description`).
   - Verify that the error reasons (e.g. `Invalid Date Format`) are also correctly translated to the chosen language (they should NOT show the raw backend string `invalid_date` or the English string if another language is selected).

## Related Issue

Closes #566

## Checklist

- [x] Backend tests pass
- [x] Frontend tests pass
- [x] Frontend builds cleanly

CSV imports now match visible category names during preview and parse Swiss apostrophe thousands separators. Failed rows are retained and shown with localized error details.

- Preselects matching CSV categories and applies matching-rule precedence.
- Parses amounts such as `1’234.56`.
- Shows skipped rows with line number, description, value, and error reason.

Risk: money math.


jorishaenseler15:
Notes on Related PRs #374:

While scrolling trough the open PR's I realised the following points:
- **Duplicate Detection (PR #374):** While PR #374 handles 1-to-1 sync matching, this PR handles frequency-based matching for batch CSV imports so same-day identical items are not dropped.

tassionoronha:
Thanks for this, especially the failed-rows surface. I ran the branch end to end with a CSV holding a bad date, a bad amount and a Swiss `-1'050.45`: the amount parses correctly now, the warning banner shows, and the details dialog opens. Full backend suite passes after merging main.

Two bugs visible in that dialog:

- The "Description" column header renders as the literal string `import.description`. The PR adds `lineNumber`, `rawValue` and `errorReason` to the locale files but not `description`, and the key doesn't exist in any locale.
- `error_reason` renders the raw backend string, so a Portuguese UI shows "Invalid Date Format" / "Invalid Amount Format". These should be codes from the backend that the frontend maps to translated strings.

Structural things:

- This is five independent changes in one PR: failed-row reporting, the Swiss separator, CSV category matching, frequency-based duplicate detection and a payee lookup tweak. The dedup rewrite in particular changes behavior for OFX too, not just CSV, and deserves to be reviewed on its own.
- The payee change adds `or_(func.lower(Payee.name) == name.lower(), Payee.name == name)`, and the second clause is implied by the first, so it's a no-op. What case was it meant to fix?
- `parse_csv` now returns `Any` with a `return_failed_rows` flag. Always returning the tuple, or a small result object, would keep the callers typed.
- `enrich_with_category_suggestions` now lets a CSV category name win over rules, since `category_set = True` is set before the rules loop. Intended? It's a real precedence change and worth saying out loud in the description.
- Merging `main` conflicts in `import_service.py`, where hidden-category filtering landed. Worth deciding whether a CSV category name should be allowed to resolve to a hidden category.

tassionoronha:
One more, found after merging `main`: `nl.json` landed after this branch was written, so the seven new keys are missing there and `src/locales/i18n.test.ts` fails.

```
FAIL src/locales/i18n.test.ts > all languages contain all keys from en.json > nl
AssertionError: Keys missing in nl:
  import.failedRowsWarning, import.viewDetails, import.failedRowsTitle,
  import.failedRowsDesc, import.lineNumber, import.rawValue, import.errorReason
```

Adding the Dutch strings fixes it, along with the missing `import.description` key from my previous comment. Backend suite and `npm run build` are both green otherwise.

jorishaenseler15:
Thanks for the thorough review @tassionoronha! I have pushed some new commits addressing your feedback and resolving the merge conflicts with `main`:

- **Deduplication Revert**: I completely reverted the frequency-based deduplication logic in `import_service.py` to match the exact behavior in `main`. I'll package that feature into a separate, isolated PR as suggested so we can keep this PR focused.
- **CSV Category Logic**: I've adjusted `enrich_with_category_suggestions`. Now it ignores hidden categories when mapping CSV categories. Also, if a category exists in the CSV, it will be pre-filled, but if a Rule evaluates and matches, the Rule's category will correctly take precedence over the CSV category.
- **Payee Lookup**: Regarding the `or_` clause that looked like a no-op: I originally added it because I was hitting a `UniqueViolation` bug with Unicode characters (like `Ü`). Under the Postgres `C` locale, `func.lower()` leaves non-ASCII characters uppercase, while Python's `.lower()` folds them. The exact-match `or_` clause was a workaround to successfully match those payees and prevent the crash! However, I noticed that PR #683 is currently open and addresses this exact collation issue with dedicated tests. Since that PR handles it comprehensively, I have fully reverted my custom `or_` clause here to avoid any overlapping changes.
- **Localization and Parsing Errors**: I refactored `parse_csv` to return strong types and structured error codes instead of hardcoded English strings. I added `import.description` and translations for all the error codes to all locale files. I also synced `nl.json` to include the missing keys.

Let me know if everything looks good now!

tassionoronha:
Thanks @jorishaenseler15 !! ✨

PR #581, opened 2026-08-13
feat(dashboard): List | Calendar view for Period Transactions

### LINKED ISSUE #580: [Feature] List | Calendar view for the dashboard's Period Transactions
## Problem

The transactions page can show a month as a calendar. The dashboard can't, even though the dashboard is where you're already navigating by month.

So to see *when* in the month the money moved (which days the balance dips, where bills cluster), you have to leave for the transactions page and re-pick the month you were already looking at. The Period Transactions table answers "what happened", never "when".

## Proposed Solution

Add the same List | Calendar switch to the dashboard's Period Transactions section, reusing the existing `TransactionCalendarView`.

The point is one calendar available in two places, not two calendars that look alike:

- same component, so it behaves identically
- same storage keys for Balance/Activity and Compact/Detailed (`securo.transactionCalendar.metric` / `.density`), so the preference follows you between pages
- same `?view=calendar&day=…` URL contract, so the section survives a refresh and can be linked to

## Alternatives Considered

- **Switch in the dashboard page header**, where the transactions page has it. The calendar only replaces the section at the bottom, so a control at the top would look like it does nothing until you scroll.
- **A separate always-visible calendar card.** The dashboard is already tall; making it a view of the existing section keeps the page the same height.

## Additional Context

One wrinkle: the calendar is a full-width grid plus a sticky day panel, so it can't sit inside the list's card. Keeping the switch in one place across both views means lifting the section title and sort control into a header above the card.

I have this working locally and will open a PR.

DannyFeliz:
## What

The dashboard's **Period Transactions** section gets the same List | Calendar switch as the transactions page.

Image

It reuses `TransactionCalendarView` as-is, so:

- the calendar behaves identically, trend strip, day grid, selected-day panel, mobile day rows
- Balance/Activity and Compact/Detailed share the same storage keys, so the preference follows you between pages
- the view is carried in the URL as `?view=calendar&day=…`

Two commits: extract the segmented switch into its own component (no behavior change), then the dashboard feature. No new translation strings.

Two things worth a reviewer's eye:

- **The section header moved out of the card.** The calendar is a full-width grid plus a sticky day panel and can't nest inside the list's card, so the title and sort control moved to a header above it, that's what keeps the switch in one place across both views. Happy to switch to the in-card header the other sections use if you'd prefer.
- **The calendar query stays disabled for a wallet-only collection.** An empty account list serializes to no parameter at all, which would quietly widen the calendar to every account.

## Why

The dashboard already navigates by month, but its only view of that month's transactions is a table, it answers "what happened", never "when". Today, seeing when the money moved means going to the transactions page and re-picking the month you were already looking at. The calendar for that already exists; it just wasn't reachable from here.

## How to Test

1. Dashboard → **Period Transactions** → **Calendar**, on a month with data.
2. Toggle **Activity** and **Detailed**, then open Transactions → Calendar: it opens with the same settings. Works the other way too.
3. Step the month with the header arrows while in calendar mode, the grid refetches and the selected day moves.
4. Click a transaction in the day panel, the edit dialog opens; saving refreshes the dashboard cards.
5. Switch back to **List**, the table and sort control return, and `view` / `day` drop out of the URL.

Checked at 1440px and 375px, light and dark. The transactions page switcher still works after the extraction.

## Related Issue

Closes #580

## Checklist

- [ ] Backend tests pass (`pytest`) n/a, frontend-only diff; leaving to CI
- [x] Frontend lints clean (`npm run lint`) 0 errors, no new warnings
- [x] Translations updated (if user-facing strings changed) no new strings; the two reused keys already exist in all 10 locales
- [x] For a large or core change, I discussed it first (issue or Discord) so we could align before the code (see [CONTRIBUTING](../CONTRIBUTING.md#before-large-or-core-changes))  #580


tassionoronha:
Thanks for that @DannyFeliz , LGTM!

PR #579, opened 2026-08-13
feat(frontend): improve account detail and transaction layouts

panga:
## What

- Improve the account-detail layout across mobile and desktop breakpoints.
- Reuse the compact mobile transaction row and refine navigation, charts, summary cards, and desktop transaction presentation.
- Show pending transactions with a compact clock badge after recurring or installment badges.
- Apply the pending badge consistently to the transaction list, calendar, account detail, and the dashboard.

## Why

The account detail needed a clearer responsive hierarchy, and pending transactions were represented inconsistently between transaction views. These changes make the same transaction read consistently while preserving the bank-sync exception.

## How to Test

1. Run `cd frontend && npm test`.
2. Run `cd frontend && npm run build`.
3. Run `cd frontend && npm run lint`.
4. Open account detail at mobile and desktop widths and verify navigation, summary cards, charts, and transaction 

## Related Issue

Extracted from [securo-finance/securo#526](https://github.com/securo-finance/securo/pull/526).

## Checklist

- [x] Backend tests pass (`pytest`)
- [x] Frontend lints clean (`npm run lint`)
- [x] Frontend builds (`npm run build`)
- [x] Translations updated (if user-facing strings changed)
- [x] For a large or core change, I discussed it first (issue or Discord) so we could align before the code (see [CONTRIBUTING](../CONTRIBUTING.md#before-large-or-core-changes))

PR #582, opened 2026-08-13
feat: separate current balances from forecasts

panga:
## What

- Make current balances posted-only for manual accounts.
- Keep connected-account current balances equal to the provider snapshot.
- Treat bank pending rows, future installments, and recurring projections uniformly as forecast activity.
- Propagate the actual-versus-projected distinction through account detail, dashboard, goals, reports, budgets, and calendar.
- Add projected income, expense, and balance fields where the UI needs both views.
- Use the latest real FX rate for future-dated transactions instead of requesting a nonexistent future historical rate.
- Keep missing cross-currency conversions unknown instead of displaying or calculating them as a 1:1 fallback.
- Heal missing or legacy 1:1 FX stamps when a pending transaction becomes posted.
- Make persisted pending/future calendar transactions editable while keeping virtual recurring projections read-only.
- Standardize virtual transaction badges as `Projected` and show their account on the dashboard.
- Add regression coverage for pending, future transactions, recurring projections, calendar interaction, and FX healing.

## Why

The application previously interpreted pending transactions differently depending on the screen or how the transaction was created. The new rule is consistent: pending never enters the current balance and always enters the forecast. This applies equally to bank pending transactions, future installments, and recurring projections.

The only exception is a connected account's current balance, which remains the provider snapshot so it continues to match the bank app. The projected view therefore acts as the user's accounts-payable view.

Future foreign-currency transactions also attempted to fetch a historical rate for a date that had not happened yet. When that failed, some views treated the native amount as the primary-currency amount. The updated behavior uses the latest real rate available and never silently substitutes 1:1.

## How to Test

1. Run `cd backend && pytest`.
2. Run `cd frontend && npm test`.
3. Run `cd frontend && npm run build`.
4. Run `cd frontend && npm run lint`.
5. For a manual account, verify a posted transaction changes the current balance while pending and future-dated transactions change only the projected balance.
6. For a connected account, verify the current balance remains exactly the provider snapshot.
7. Create a future USD transaction with BRL as the primary currency and verify it uses the latest real cached/provider rate, not 1:1.
8. Verify persisted pending/future calendar rows open the transaction dialog while virtual recurring rows remain read-only.
9. Verify virtual recurring rows use the same `Projected` badge in account detail, dashboard, calendar, and drill-down, and that the dashboard shows their account.
10. Verify the same actual-versus-projected interpretation in account detail, dashboard, goals, reports, budgets, and calendar.

Validated locally:

- 532 backend tests passed and 5 were skipped across the affected suites.
- 75 frontend tests passed.
- Frontend production build passed.
- Frontend lint completed with 0 errors (41 existing warnings).

## Related Issue

- Depends on [#579](https://github.com/securo-finance/securo/pull/579).
- Extracted from [#526](https://github.com/securo-finance/securo/pull/526).

## Checklist


panga:
Ready for review.

tassionoronha:
Hey @panga. 

Pushed a few commits on top. Three reverse choices you made deliberately, so worth a look:

- generate_pending back to posted: a due occurrence is an actual, and nothing ever posted it on a manual account
- category total is posted-only now so it matches the expenses card, forecast moved to projected_total
- the card keeps pending again: the list excluded it while the bill total included it, so one card showed two numbers

Waiting for your check and approval to merge it :)

panga:
@tassionoronha I’ve tested all the chart and dashboard changes, and everything looks good on my end.

Regarding recurring transactions defaulting to pending, I introduced this behavior because, in my case, roughly 9 out of 10 recurring transactions failed to match their bank-synced counterparts. This resulted in duplicates, with no easy way to identify or filter the transactions that had failed to match.

The tradeoff, as I see it, is:
- Bank-synced accounts: When an imported transaction does not match an existing recurring transaction, both remain and effectively create a duplicate.
- Manual accounts: A recurring transaction remains pending until the user manually marks it as posted.

I consider duplicates in bank-synced accounts more disruptive. Since manual accounts already require manual maintenance, reviewing the transaction status could reasonably be part of that workflow.

If defaulting all recurring transactions to pending is not desirable, we could add a default status option to the recurring template, allowing users to choose between pending and posted.

We are already missing other things like payee that can be used for better matching. e.g: https://github.com/securo-finance/securo/issues/588

tassionoronha:
Hey @panga. 

Your data wins here, so I went with a conditional instead of flipping it back.

The row now starts pending **only when the account has a bank connection**, since that's where a real charge can arrive and duplicate. On manual accounts nothing would ever promote it, so pending would be permanent and the current balance would never catch up.

panga:
Awesome, this handles the edge case very well.

LGTM

tassionoronha:
Great, thanks for confirming @panga !

PR #586, opened 2026-08-13
feat: normalize imported transaction descriptions with rules

### LINKED ISSUE #494: [Feature] Normalize imported transaction descriptions with rules
## Problem

I use Enable Banking, and some imported transaction descriptions contain random
IDs or changing values.

For example, Amazon Prime appears as:

`D01-1234567-1234567 AMZNPrime DE ABDj947DKEL917Dj`

The first and last parts are random, while only `AMZNPrime DE` stays the same.
Because only two of the four tokens are stable, the recurring transaction matcher
does not reach the required 60% similarity.

Another transaction appears as:

`12345678901234 KV12345678901 01/26`

Only `KV12345678901` is stable. The first value changes, and `01/26` represents a
changing month and year. This means only one of the three tokens can be matched.

These are clearly recurring payments, but Securo cannot reliably associate them
with their recurring definitions.

Obviously the above numbers have all been anonymized.

## Proposed Solution

Allow rules to normalize or replace imported transaction descriptions.

For example:

- If the description contains `AMZNPrime DE`
- Set the description to `Amazon Prime`
- Set the payee to `Amazon Prime`
- Set the category to `Memberships`

Or:

- If the description contains `KV12345678901`
- Set the description to a fixed, user-defined value

The original description received from the bank should ideally remain visible in
the transaction details.

Alternatively, recurring transactions could match against the normalized payee or
a stable `contains` value instead of requiring the full imported description to be
similar.

## Alternatives Considered

I tried using only the stable parts as recurring descriptions, such as
`AMZNPrime DE` or `KV12345678901`. However, these do not reach the required 60%
similarity compared with the complete imported descriptions.

Manually editing every imported transaction or updating the recurring definition
each month would work, but defeats the purpose of bank sync and recurring
transactions.

## Additional Context

This is similar to how the C24 banking app allows transaction information to be
cleaned up while keeping the original bank data available.

seraphblade2010:
## What

- add a rule action for replacing imported transaction descriptions
- add matching against the raw imported payee
- preserve the bank-provided description and display it in transaction details
- apply description normalization before recurring matching for both imports and bank sync
- preserve imported and recurring-placeholder categories when applying other rule actions
- keep duplicate detection stable after normalization
- update the rule UI, API schemas, translations, migrations, and tests

## Why

Bank-synced transaction descriptions can contain changing IDs, dates, or other noisy values. This prevents otherwise recurring transactions from reaching the required similarity for automatic matching and makes the transaction list difficult to read.

Rules can now turn these descriptions into stable, user-defined values before recurring matching while retaining the original bank-provided text for reference.

## How to Test

1. Create a recurring expense with a stable description, such as `Amazon Prime`.
2. Create a rule that matches a stable part of an imported description and sets the description to `Amazon Prime`.
3. Import or sync a matching transaction with changing text around the stable part.
4. Confirm the normalized description is shown and the transaction is linked to the recurring definition.
5. Open the transaction details and confirm the original imported description remains visible.
6. Apply a matching rule to an existing categorized transaction with category replacement disabled and confirm its category remains unchanged.
7. Import the same transaction again and confirm duplicate detection still skips it.
8. Open the rule dialog and confirm priority editing and validation layout work on desktop and mobile widths.

## Validation

- Backend: 2523 passed, 7 skipped; 92.84% coverage
- Backend lint and type checks: Ruff and ty passed
- Alembic: single head at migration 069; PostgreSQL SQL generation passed
- Frontend: 69 tests passed
- Frontend lint: 0 errors; 42 existing warnings
- Frontend type check and production build passed
- Fork GitHub Actions: Backend Tests, Frontend Checks, and Helm Checks passed
- Manually tested imported transaction normalization, recurring matching, original-description visibility, category preservation, duplicate detection, rule priority editing, responsive layout, and validation styling

## Related Issue

Closes #494

## Checklist

- [x] Translations updated
- [x] This core rules/import change was discussed and its direction confirmed in #494


seraphblade2010:
I also tested this manually quite a bit before opening the PR:

* Imported a transaction with a changing description and confirmed the rule cleans it up before matching it with the recurring transaction, without creating a duplicate.
* Confirmed the original description, imported payee, notes, category and recurring link are all kept.
* Applied an updated rule to the existing transaction with category replacement disabled and confirmed the category stayed unchanged.
* Imported the same file again and confirmed duplicate detection still works.
* Tested priority editing and the rule validation/layout in Firefox and Brave.
* Tested the mobile layout on a Pixel 10 Pro with Firefox, including the keyboard, scrolling and reachable buttons.

Everything worked as expected for me, including the final validation style fix :)

Ready for review from @tassionoronha or somebody else willing to dig through this ^^

panga:
Recurring transactions does not have payee field. I think it is a gap and payee + date + value should be definitely used a strong signal for matching, ignore the description in that case. Created https://github.com/securo-finance/securo/issues/588

seraphblade2010:
@panga I agree that the payee could be a useful additional signal, but I would be careful about ignoring the description completely even when the other values match.

These are two examples imported from my Klarna credit card. Both use `Klarna Bank AB` as their payee, while only the description tells me that one was an Amazon purchase and the other was from Shell:

Amazon example:
Screenshot_20260814_155911


Shell example:
Screenshot_20260814_160021

The amounts happen to differ in these screenshots, so this exact pair would not collide. My point is that the amount does not identify the merchant either—two unrelated purchases can have the same amount and date while also sharing the same generic Klarna payee. In that case, the description may be the only remaining signal that distinguishes them.

I think #588 makes sense as an additional reconciliation path when a meaningful payee was explicitly selected and it produces one unique match. I would just keep the payee optional and retain description matching as a fallback or additional safeguard. That would complement the normalization in this PR instead of replacing it :)

panga:
I understand your concern. Klarna is very generic at descriptions. This can be mitigated by fallback to description similarity if two transactions found with the same amount because today this can already happen if you have two purchases at Amazon from distinct payees, so it is worse.

seraphblade2010:
That approach sounds good to me :) @panga 

This PR already makes the imported payee available before recurring matching, so #588 can build on the current implementation without requiring a redesign.

Im open to either adding it to this PR or implementing it separately as a follow-up. Both approaches work for me, so I would leave that decision to you and the maintainers based on what would be easier to review :)

gaamelu:
@seraphblade2010 @panga

Thanks for the work on #586. While testing it, I noticed a small gap in how `original_description` is populated.

Would you consider including the following in this PR?

- populate `original_description` when transactions are created through sync or import;
- preserve it when an existing placeholder transaction is matched with a provider transaction and promoted to a regular transaction. Otherwise, the original provider description can be lost at the promotion step, making auditing, debugging, and future normalization improvements harder;
- add regression tests for these paths.

This would ensure that the normalized `description` remains the user-facing value while the original provider/import description is consistently retained.

These changes don’t alter the normalization rules or matching behavior; they just complete the provenance guarantees of the new field. I have a local implementation ready to test if this direction makes sense.

gaamelu:
I added the two commits in a separate draft follow-up. If you think they fit better here, feel free to pull or cherry-pick them into this PR; otherwise, we can continue with the follow-up. And if you do not think these changes are useful, we can simply drop them.

AI-assisted message, reviewed and approved by @gaamelu.

seraphblade2010:
Thanks again @gaamelu, I pulled your two commits into this PR and kept the authorship :)

While integrating them I added one extra safeguard: the original description now only stores the provider/import value, while a separate internal flag tracks whether the displayed description was changed by a rule. This allows Reset and reapply to restore rule-managed descriptions without overwriting a description that was edited manually.

The latest revision now preserves original descriptions during regular syncs, imports and placeholder promotion, keeps the relevant imported metadata when promoting placeholders and adds regression tests for these flows. I also rebased onto the current main and moved the migration from 069 to 070 to resolve the conflict.

@tassionoronha I would consider this ready to review as-is and handle the implementation of #588 separately to not bloat this PR more than it already is. @panga I hope thats okay for you.

panga:
@seraphblade2010 thanks for doing the work. It is a great step making import / sync more robust.

tassionoronha:
Nice work @seraphblade2010, @gaamelu and @panga !
I'll take this one for review by tomorrow 👍🏼

tassionoronha:
Thanks @seraphblade2010! 

**Two fixes.** `apply_single_rule` was replacing a description the user had typed. `apply_all_rules` already protects that through your flag, but re-saving a rule with "apply to existing" had no guard. And the dialog lost the raw payee line when it gained the original description line, which for rows created before the migration means no hint at all.

**Placeholder merge, and why it took a detour.** The merge was replacing the recurring definition's description and payee with the bank's raw text. Both are values the user chose, and both were guarded before this PR.

My first attempt was just to put the guards back, and one of your tests caught why that does not work: once the description stays as the definition's wording, a rule whose condition is written against the bank's text stops matching, and the charge loses the category and notes it should have got. The overwrite was load-bearing.

So instead of re-running the rules against the placeholder, the merge now folds in the `preview` the rules already produced from the incoming charge. They still classify it, and the merge goes back to filling gaps: category and payee only when empty, notes combined so neither side is dropped, raw description kept as provenance. A merged placeholder now keeps a description that was never rule-managed, which is accurate, it is the definition's wording and not a rule's output.

If this looks good for you guys I'm ok to merge.

seraphblade2010:
I think this should cover all the bases, thanks for the fixes @tassionoronha :)

From my side its ready to merge as-is.

Edit: found a small bug while inspecting again, will push a fix in a few minutes.

seraphblade2010:
I found two more edge cases while testing the latest changes around recurring placeholder merges.

Both come down to rule results being calculated correctly on the preview, but not fully carried over when the incoming transaction is merged into an existing recurring placeholder:

set_payee could get lost and fall back to the pre-rule/imported payee ID, or stay unset completely if the incoming transaction had no raw payee.
ignore was not carried over at all.

I reproduced both against def33df with CSV imports and a normal-import control next to each case. The normal imports behaved as expected, while the matching recurring transactions lost the rule result.

The follow-up keeps the raw payee handling as-is, but takes an empty payee_id from the rule preview and also carries over is_ignored.

I retested both cases manually afterwards and they now behave the same as the normal import path.

Also added regression coverage for import + sync, including the no-raw-payee case.

With this now fixed I think the PR is ready for merge @tassionoronha

tassionoronha:
Yeah, both are fair @seraphblade2010 . The `set_payee` one is on me: I pulled category and notes off the preview but left the payee reading the raw import id, so the rule's result just vanished there. And you're right about `ignore`, no reason for the placeholder path to behave differently from a normal import.

Pulling `payee_id` off the preview is the right call. 👍🏼

PR #590, opened 2026-08-14
feat: sort account pickers by display name

### LINKED ISSUE #589: [Feature] Account pickers should be ordered by display name
## Description

Account pickers are currently rendered in the order returned by the accounts API. The API orders accounts by their canonical provider name, so accounts with a user-defined `display_name` can appear in an unexpected position even though the picker shows the display name.

This affects account selection across transaction, recurring transaction, transfer, import, rule, goal, collection, filter, and group-settlement flows.

## Steps to Reproduce

1. Create or sync two or more accounts.
2. Set a display name on one account that sorts differently from its canonical name.
3. Open an account picker, such as the transaction or recurring transaction dialog.
4. Observe that the visible account names are not alphabetically ordered.

## Expected Behavior

Every account picker should be ordered by the visible account name: `display_name` when present, otherwise the canonical `name`.

Sorting should be limited to presentation. The accounts API must continue returning the canonical `name` and the separate `display_name`, and account-listing screens should retain their existing business-specific ordering.

New transaction, recurring transaction, and transfer forms should initially select the first account shown in the display-name-sorted picker instead of an unrelated first row from the API response.

## Actual Behavior

Account pickers inherit the API order, which is based on the canonical account name rather than the name displayed to the user.

## Environment

- Deployment: All
- Browser: All
- OS: All

## Additional Context

The sorting operation should not mutate the account array because it may be sourced from the React Query cache. Edit forms and callers that provide an explicit default account must preserve that selection.

panga:
## What

- Add a shared, non-mutating helper that orders accounts by `display_name` when present and falls back to `name`.
- Apply the ordering consistently to account pickers in:
  - transaction and recurring transaction dialogs
  - transfer and linked-transfer dialogs
  - transaction filters on desktop and mobile
  - rules
  - imports
  - goals
  - collections
  - group settlements
- Update recurring and rule dialog account types to include the optional `display_name` field.
- Add unit coverage for display-name fallback ordering and input immutability.
- Keep the accounts API contract and backend ordering unchanged.

## Why

Account pickers display a user-defined account name when available, but previously inherited API ordering based on the canonical provider name. As a result, the visible options could appear out of alphabetical order.

Keeping this as presentation-only sorting avoids changing API consumers, account-listing behavior, or the canonical account name.

## How to Test

1. Create or sync several accounts with canonical names that sort differently from their display names.
2. Set display names on some, but not all, accounts.
3. Verify account options are ordered by the visible name in transaction, recurring transaction, transfer, linked-transfer, filter, rule, import, goal, collection, and group-settlement pickers.
4. Verify accounts without a display name are ordered using their canonical name.
5. Verify new transaction, recurring transaction, and transfer forms select the first visible account option.
6. Verify editing existing records and explicit transfer defaults preserve their selected account.
7. Verify the accounts API still returns canonical `name` and a separate `display_name`.

Automated verification performed:

- `npm test -- src/lib/account-utils.test.ts`
- `npm run build`
- `npm run lint` (passes with existing repository warnings)
- `git diff --check`

## Related Issue

Closes #589

## Checklist

- [x] Backend behavior is unchanged; backend tests are not required for this frontend-only change
- [x] Frontend lints complete without errors (`npm run lint`)
- [x] No translations required (no user-facing strings changed)

### DIFF
--- a/frontend/src/components/link-transfer-dialog.tsx
+++ b/frontend/src/components/link-transfer-dialog.tsx
@@ -1,5 +1,5 @@
 import { useMemo, useState } from 'react'
-import { getAccountName } from '@/lib/account-utils'
+import { getAccountName, sortAccountsByDisplayName } from '@/lib/account-utils'
 import { useTranslation } from 'react-i18next'
 import { useDisplayLocale, useDateLocale } from '@/hooks/use-display-locale'
 import { useQuery } from '@tanstack/react-query'
@@ -158,8 +158,8 @@ export function LinkTransferDialog({
     // Manual accounts (no bank connection) where we can auto-create the
     // counterpart, since a synced account would already have a real
     // transaction showing in the candidates list above.
-    const manualCounterpartAccounts = accounts.filter(
-      (a) => a.connection_id == null && a.id !== anchor!.account_id,
+    const manualCounterpartAccounts = sortAccountsByDisplayName(
+      accounts.filter((a) => a.connection_id == null && a.id !== anchor!.account_id),
     )
 
     return (
--- a/frontend/src/components/rule-dialog.tsx
+++ b/frontend/src/components/rule-dialog.tsx
@@ -1,6 +1,6 @@
-import { useState } from 'react'
+import { useMemo, useState } from 'react'
 import { useTranslation } from 'react-i18next'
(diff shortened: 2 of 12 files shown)


tassionoronha:
Nice work here @panga, thanks for that!

PR #593, opened 2026-08-15
fix: join accounts in previous balance query

chrislopez24:
## Summary

The previous-balance expression reads `Account.currency` without joining `Account`. SQLAlchemy therefore adds `accounts` as an independent `FROM`, producing a Cartesian product. With multiple accounts, every transaction is aggregated once per account, increasing query cost and producing incorrect balances when currencies differ.

## Changes

Join `Account` on `Transaction.account_id`, matching the current-balance subquery. Each transaction is now evaluated against its actual account currency.

## Validation

- `ruff check app/services/account_service.py`
- `pytest tests/test_account_service.py`: 51 passed
- PostgreSQL plan on 3 accounts: 9,600 → 3,200 intermediate rows; 24.1 ms → 14.4 ms

### DIFF
--- a/backend/app/services/account_service.py
+++ b/backend/app/services/account_service.py
@@ -83,6 +83,7 @@ async def get_accounts(session: AsyncSession, workspace_id: uuid.UUID, include_c
             Transaction.account_id,
             func.coalesce(func.sum(signed_amount), 0).label("previous_balance"),
         )
+        .join(Account, Transaction.account_id == Account.id)
         .outerjoin(Category, Transaction.category_id == Category.id)
         .where(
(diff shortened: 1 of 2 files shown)


tassionoronha:
Hey @chrislopez24 , thanks for this. 
It's indeed a real inconsistency. Can you add some tests here to avoid future regression?
Cheers :)

chrislopez24:
sure @tassionoronha! I added a focused regression test with BRL and USD accounts. I verified that it fails without the Account join due to the Cartesian product and passes with the fix.

tassionoronha:
Nice work, LGTM. Thanks @chrislopez24!

PR #594, opened 2026-08-15
fix: cap Enable Banking consent at 179 days

chrislopez24:
## Summary

Enable Banking exposes a 180-day maximum, which the frontend can pass back as `valid_until_days`. Building `valid_until` at the exact boundary leaves no margin for timestamp precision, processing delay, or provider-side validation and can cause the authorization request to be rejected as exceeding the limit.

## Changes

Cap consent validity at 179 days inside `_build_auth_payload`. This covers both the default and caller-provided values while preserving shorter institution limits.

## Validation

- `ruff check app/providers/enable_banking.py`
- `pytest tests/test_providers_enable_banking.py`: 21 passed


tassionoronha:
Hey @chrislopez24 , thanks for this one and welcome here :)
This one makes more sense as a safe cap indeed. 
LGTM :)

PR #602, opened 2026-08-17
feat: enrich payees with contact and fiscal data

tassionoronha:
Payees now carry the contact and fiscal data an invoice needs. They already
are the counterparty — a real FK from `Transaction`, auto-assigned by sync and
by the rules engine — so a separate client entity would mean the same company
existing twice, once created by sync and once typed by hand.

## Documents are rows, not columns

`payee_tax_ids (payee_id, kind, value)`, with `kind` a closed enum. Generic
numbered fields were the alternative and were rejected: `fiscal_2 =
"12.345.678/0001-95"` cannot be interpreted without knowing which configuration
wrote it, so reordering that configuration in a later release would silently
change what every stored value means, and no query could ask for "this payee's
CNPJ" without loading it first.

Two pieces, kept apart on purpose:

- **A kind defines the document** — how to normalise, validate, mask and name
  it. One definition, used everywhere, so the same number means the same thing
  in every workspace.
- **A jurisdiction chooses from them.** A pack is a country code and an ordered
  list of kinds, nothing more. It cannot weaken a validator or rename a
  document.

Two properties follow. **A pack suggests, it never restricts:** every kind stays
storable everywhere, because the counterparty's country is not the workspace's —
a Brazilian consultancy billing a Berlin client needs `vat`, and the BR pack
must not forbid it. And **no pack is still a working configuration:** an unset
jurisdiction falls back to free text under `other`, so a country nobody has
contributed a pack for is usable today.

## 42 jurisdictions

Covering every currency the product already supports, plus the euro members —
the app cannot sensibly bill in CLP and have no word for a RUT. Adding a country
is one TOML file and, if the format is checkable, one named validator. A test
fails if a currency is added later without a pack.

Nothing here touches the network. Confirming a VAT number against VIES would
disclose who a workspace bills, so validation is local and format-only where
that is all that can be checked.

## Three things worth knowing about the validators

**Check digits are asserted against real documents.** Each of the twelve
check-digit validators is tested against a public identifier of a known
organisation. Written from a prose description alone, three of them looked
correct and were wrong — and a wrong routine here silently refuses a real
client.

**Where a rule is unpublished, the validator says so.** Canada, New Zealand,
Japan, Uruguay, Costa Rica, the Dominican Republic, Ukraine, Hungary and
Indonesia check shape and size only, because the issuing authority does not
publish a check rule or it varies by year or state. Each has a test asserting
that gap deliberately, so nobody later mistakes it for an oversight.

**Masks are only given to fixed-length documents.** A mask is applied as the
user types and truncates at the template's width, so a seven-digit Chilean RUT
under an eight-digit mask reads back as `12.345.67-8`. A test fails if a mask is
added to a kind whose validator accepts a second length.

## Interface

The document picker is grouped by country and searchable — a flat list asked the
user to already know that Partita IVA is Italian. The country-to-documents map
comes from the packs rather than being duplicated in the browser.

Country names come from `Intl.DisplayNames` instead of the locale files. Forty
countries across ten locales would be four hundred strings restating what the
platform already knows, and each new pack would add ten more.

A shared kind is placed by its value. `vat` belongs to nineteen countries, so
its label cannot say whose number it is — but the number's own prefix can, which
is why one kind serves all of them. Two prefixes are not ISO codes and are
mapped explicitly: Greece bills under `EL`, Northern Ireland under `XI`.

## Naming

pt-BR renames "Beneficiários" to "Cadastros". "Beneficiário" means specifically
the recipient of a transfer, and the entity now also holds whoever pays you.
pt-PT uses "Entidades" instead, because "cadastro" reads as criminal record
there. EN keeps "Payees", the convention it already follows.

## Compatibility

Additive throughout. Every new field is nullable, existing payees keep working
with no document rows and no backfill, and the personal workspace gets the same
fields. There is no separate client type and no flag: a payee becomes a client
by having an invoice, not by being declared one.

Migration `070`, verified up, down and up again on Postgres with no residue.


mckusck:
Hi @tassionoronha. As soon as I upgraded to 0.14.3 yesterday, I went and added to almost all of many of my payees (including myself) the most known numbers such as CPF and CNPJ (specific for Brazil), but I noticed my already downloaded transactions did not identify the payees when the transaction brought one of these social numbers. Can I expect that Securo will autofill the payees for the future downloaded transactions, or do I need to set rules for that? Sorry if that behavior is already clarified in this PR, but I wasn`t able to clarify it, as of now.

PR #631, opened 2026-08-19
feat: add the missing supported currencies to the setup and register picker

ssrahul96:
## What

Adds the currencies that were missing from the shared currency selector used in the setup and register flows: INR, and along with it JPY, MXN and SEK.

## Why

All four are already supported by the backend (`SUPPORTED_CURRENCIES` and the currency metadata) and already have a formatting locale on the frontend, but they were never added to the picker, so they could not be chosen during onboarding or registration. The list now matches the backend default set, in the same order.

## How to test

1. Start the app (`docker compose up`).
2. Open the currency selector on the setup or register screen.
3. Confirm JPY (¥), MXN ($), INR (₹) and SEK (kr) appear with their flags and can be selected.

## Checklist

- [ ] Translations updated (no user-facing strings changed)


tassionoronha:
Thansk @ssrahul96! 
Doing the review I realised other currencies were also missing in the list so I updated our context here. 
Welcome :)

PR #643, opened 2026-08-20
fix: confirm destructive deletions

### LINKED ISSUE #604: [Bug] Several delete actions have no confirmation step
## Description

Several delete actions permanently remove an item as soon as the delete button is clicked:

* Recurring transactions
* Categories
* Category groups
* Rules
* Groups
* Budgets

The delete button for Groups is larger and located inside the edit dialog, but it still performs the deletion with a single click.

This makes it easy to accidentally delete something, especially since there is currently no undo or trash for these items. I originally noticed this after accidentally deleting one of my recurring transactions.

Payees already use a confirmation dialog before deletion, so that existing UI and behaviour could be reused or adapted for the affected pages.

## Steps to Reproduce

1. Open Recurring, Categories, Rules, Groups or Budgets.
2. Click the delete button for one of the affected items.
3. The item is deleted immediately without another confirmation step.

For category groups, use the delete button beside the group name. For Groups, open an existing group for editing and click the larger Delete button.

## Expected Behavior

Clicking Delete should first open a confirmation dialog following the existing payee deletion pattern.

The dialog should:

* Identify the item being deleted where possible.
* Explain that the action cannot be undone.
* Mention relevant side effects where necessary.
* Provide a Cancel button and a destructive Delete button.
* Only send the deletion request after Delete is confirmed.
* Prevent duplicate deletion requests while the request is pending.

For recurring transactions, the dialog should explain that the recurring definition will be permanently deleted while existing transactions remain and lose their recurring link.

For category groups, it should explain that the group will be deleted while its categories remain and become ungrouped.

## Actual Behavior

The affected items are deleted immediately when their delete button is clicked.

There is no confirmation step or way to undo the deletion.

Payees already behave as expected and show a confirmation dialog before the actual deletion.

## Environment

* Deployment: Kubernetes / Helm
* Browser: Firefox
* OS: Fedora 44

## Screenshots

Not applicable since the deletion happens immediately without another UI step.

## Additional Context

This should only change the frontend confirmation flow. The existing deletion APIs and backend behaviour should remain unchanged.

The payee deletion dialog appears to provide a suitable existing implementation to follow. A small shared confirmation component could also make sense if it avoids repeating the same dialog across all affected pages without turning this into a broader dialog refactor.

Disclaimer: This issue body was written with ChatGPT / Codex because I'm not crafty enough with words to describe it better. The report itself comes from real usage of Securo after I accidentally deleted one of my recurring transactions :/

seraphblade2010:
### Summary

- Add confirmation dialogs before deleting recurring transactions, categories, category groups, rules, groups, and budgets.
- Reuse a focused app-native confirmation component based on the existing payee flow.
- Add resource-specific explanations of what each deletion affects.
- Preserve existing backend deletion behavior, permissions, mutations, invalidation, and error handling.
- Improve dialog accessibility and prevent closing or submitting again while deletion is pending.

## Why

These destructive actions previously deleted items immediately, despite having no undo or trash flow.

### Testing

- `npm test` — 128 tests passed.
- `npm run lint` — passed with existing non-failing warnings.
- `npm run build` — passed.
- `git diff --check` — passed.
- Fork Backend, Frontend, and Helm Actions — passed.

### Local UI verification

- All six affected delete actions open a confirmation dialog.
- Cancelling does not delete the item; confirming deletes the selected item.
- Resource-specific titles, names, and descriptions display correctly.
- Normal and recurring budgets use the correct descriptions.
- Cancelling group deletion returns to the group edit dialog.
- The longer text wraps correctly on desktop and the tested mobile viewport.

## Related Issue

Closes #604

## Checklist

- [x] Backend CI action passed (skipped because this PR has no backend changes)
- [x] Frontend lint passed (`npm run lint`; existing non-failing warnings remain)
- [x] Translations updated
- [x] This is a focused bug fix covered by issue #604


tassionoronha:
Really nice work @seraphblade2010 , the pattern is consistent across all six and the copy is specific per resource 🙌 

Two things before merge: deleting a category that has transactions hits an FK error (500), and the dialog just sits there, so it needs an onError toast plus copy saying it can't be deleted. Same missing onError on budgets, recurring and rules. groups.tsx already does it right.

seraphblade2010:
@tassionoronha Thanks for catching this :)

I added the missing error handling for categories, budgets, recurring transactions and rules. Categories which are still in use now show a proper message, while the others fall back to the normal error message.

I did briefly push a bad commit which tried to detect technical database errors in the frontend. That was not a good solution, so I force-pushed the cleaned-up version to remove it from the history.

tassionoronha:
Tested this locally, the error path works nicely now: the toast explains why and the dialog stays open so you can back out 🙌 

Two things left: mapping HTTP 500 to "in use" will fire on any server error, so the real fix is backend-side (catch the IntegrityError, return 409 with a detail). And extractApiError in transaction-dialog.tsx already does what the five new onError blocks hand-roll. Small one: Escape on the confirmation re-opens the group edit dialog, so it takes two Escapes to leave.

seraphblade2010:
@tassionoronha  sorry its been a bit but now I’ve addressed the remaining points:

- category FK conflicts are now handled backend-side as 409 with a safe detail message
- the new delete handlers reuse the shared extractApiError helper
- group delete confirmation now exits with a single Escape/X/outside dismiss, while explicit Cancel returns to editing
- added regression coverage for the real FK/rollback path and the shared error helper

Manual dialog checks all pass, as do all normal tests.
I hope now it works as expected ^^

tassionoronha:
No worries @seraphblade2010, you did a good job. 
In the end there was some groups dismissal to be improved and other small things, which I pushed a fix. 
Thanks for that!

PR #645, opened 2026-08-20
feat(rules): mix AND and OR with nested condition groups

### LINKED ISSUE #608: [Feature] Complex rule support

## Problem

It is frustating to create complex rules due to the lack of the ability to create nested conditions. 

## Proposed Solution

Allow nested conditions. Rules are already managed as json so no DB schema changes are required.

Image

## Additional Context

I have vibe coded this on my fork and it is working fine.

tassionoronha:
A rule's conditions could only be joined by one operator, so `type is debit AND (description contains UBER OR contains 99POP)` had no way to be expressed. Each entry of `conditions` may now be either a leaf, as before, or a group carrying its own operator and its own leaves. Groups hold leaves only, capping a rule at two levels, which covers any expression in normal form while staying readable in the editor. Existing rules are flat lists of leaves, so they keep their exact meaning and nothing needs migrating.

**This is @paulovitor6's work, not mine.** He wrote it in #606, closed it himself, and the commit here keeps his authorship. What I did on top: rebased it onto current `main`, resolved the conflict in `rule-dialog.tsx` in favour of main's newer actions block (the one with `set_description` and its inline error), and fixed two things that broke CI:

- `rule-conditions.ts` / `rule-conditions.test.ts` imported types from `@/types`, but `tsconfig.node.json`, which compiles the test files, carries no path alias, so `npm run build` failed. Both now use the relative import that `rule-form-utils.ts` already uses.
- `ty` rejected `exported.conditions[0].field` in `test_rule_service.py` once the list became a union; the test narrows with `isinstance` before reading the leaf.

Closes #608


paulovitor6:
Hey @tassionoronha, I closed the PR because I saw that it was not following the proper process that you suggested. I'm glad to see that this will be on the official version.

Issue #648, opened 2026-08-20
[Bug] Automatic transfer-pair matching produces false positives when a third-party reimbursement coincidentally matches a purchase amount


vhsantos26:
## Description

Securo automatically detects "transfer pairs" (two transactions, typically in different accounts, that represent the same money moving between the user's own accounts) and excludes matched pairs from income/expense totals. The matching appears to key off amount + nearby date across accounts, without confirming the transaction actually looks like an internal movement. As a result, a PIX received from a third party (e.g. a friend paying back their share of a bill) that happens to match the amount of an earlier purchase gets misidentified as an internal transfer, and both the original purchase and the reimbursement silently disappear from income/expense and bill totals.

## Steps to Reproduce

1. Make a credit card purchase for amount X.
2. A few days later, receive a PIX/transfer from a different person (not one of your own accounts) for the same amount X, for example a friend reimbursing their share of a shared bill.
3. Check whether `transfer_pair_id` got set on both transactions, or whether the purchase disappears from the credit card's bill total / the reimbursement disappears from income for that period.

## Expected Behavior

Transfer-pair matching should only link transactions that are genuinely the same money moving between the user's own accounts, not any two transactions that happen to share an amount and a nearby date. A reimbursement from a different person shouldn't be treated as an internal transfer just because the amount lines up.

## Actual Behavior

Reproduced concretely: a R$374.66 credit card purchase (a bar tab) and an unrelated R$374.66 PIX received the next day from a different person (paying back their share of that same tab) were linked as a transfer pair. This silently removed the purchase from the card's bill total and the reimbursement from income for that period.

## Environment

- Deployment: Docker (self-hosted)
- Data source: Pluggy (Brazilian open banking)
- Securo version: v0.14.3

## Screenshots

N/A

## Additional Context

Adjacent to #647, #535, #489 and #473 in that all four are ways Pluggy-connected Brazilian account totals end up skewed, but this one is about the transfer-detection heuristic itself rather than a reporting/aggregation query, so I'd expect the fix to live somewhere else in the codebase. I haven't traced the matching logic, so I'm not attaching a suggested fix or a PR. This feels like the kind of core-mechanism change worth discussing first per the contributing guide.

suzarte:
I think this issue could also be a good opportunity to consider a broader case related to the same problem: **incoming transactions that are not actually income**.

For example, there are at least two common situations:

1. **Merchant refund**
[...]

PR #649, opened 2026-08-20
fix(accounts): don't let category flags shrink a credit card's bill total

### LINKED ISSUE #647: [Bug] Credit card bill total excludes transactions in transfer/ignored categories, understating the amount owed
## Description

The credit card detail page's "Total da fatura" (bill total) for a given cycle applies the same P&L exclusion filters used for personal budget reporting: it excludes transactions whose category has `treat_as_transfer=True` or `is_ignored=True`.

This conflates two different questions:
- "Should this count against my personal budget?" (correctly excludable, e.g. a consortium/investment-style payment shouldn't inflate discretionary spending)
- "How much do I owe the bank on this card?" (should never be reduced by category: the bank doesn't care how the user tags their own spending)

## Steps to Reproduce

1. Connect a Brazilian credit card via Pluggy.
2. Categorize one of the card's real purchases under a category with `treat_as_transfer=True` (e.g. the default "Transferências" category), for example a real estate consortium payment.
3. Open the credit card account detail page and select the cycle containing that transaction.
4. Compare "Total da fatura" against the provider's actual bill total (`credit_card_bills.total_amount` for the same due date, synced directly from Pluggy's bill data and unaffected by category).

## Expected Behavior

"Total da fatura" should match the provider's bill total for that cycle: a credit card bill shouldn't change based on how a transaction happens to be categorized for personal-budget purposes.

## Actual Behavior

Any transaction in a `treat_as_transfer` (or `is_ignored`) category disappears from "Total da fatura", understating what's actually owed.

Reproduced concretely on my own workspace: a R$2.880,18 real-estate-consortium charge on a Santander card, categorized under the default "Transferências" category, is excluded from the July cycle's "Total da fatura". The independently-synced `credit_card_bills.total_amount` for the same due date (2026-07-14) correctly shows R$23.972,93 (matches the bank's own invoice), while the account-detail page's "Total da fatura" for the same cycle shows R$20.718,07, a ~R$3.254 gap explained by `treat_as_transfer`-categorized transactions.

## Root cause (traced in code)

`get_account_summary()` in `backend/app/services/account_service.py` computes the credit-card cycle expense total using `counts_as_pnl()` (from `backend/app/services/_query_filters.py`) in its WHERE clause:

```python
if account.type == "credit_card":
    signed_for_bill = case(
        (Transaction.type == "credit", -func.abs(effective_amount)),
        else_=func.abs(effective_amount),
    )
    expenses_result = await session.execute(
        _scope(select(func.coalesce(func.sum(signed_for_bill), 0)).where(
            Transaction.account_id == account_id,
            Transaction.source != "opening_balance",
            bucket_date <= today,
            Transaction.status == "posted",
            counts_as_pnl(),
        ))
    )
```

`counts_as_pnl()` excludes rows whose category is `treat_as_transfer` or `is_ignored`. That's the right rule for personal P&L totals, but this same value feeds the credit-card-specific "Total da fatura" figure, which is presented as (and should equal) the bank's invoice amount, independent of category.

The already-synced `credit_card_bills` table (populated directly from the provider's own bill data via Pluggy) has the correct number for the same cycle and isn't affected by this filter, so the discrepancy only shows up when comparing the two.

## Suggested Fix

For credit-card accounts, the cycle "Total da fatura" figure should either:
- Prefer `credit_card_bills.total_amount` for the matching cycle when available (most accurate, since it's the provider's own bill), falling back to the transaction sum only when no synced bill exists; or
- If computed from transactions, skip the category-level `treat_as_transfer`/`is_ignored` exclusion for this specific figure (keep it for the personal-expense/dashboard totals, where it's correct). A card's bill total should only exclude a transaction when the *transaction itself* is flagged `is_ignored` (an explicit "don't count this" from the user), never based on its category, since the bank's charge doesn't disappear just because the user tagged it as a transfer/investment category.

## Environment

- Deployment: Docker (self-hosted)
- Data source: Pluggy (Brazilian open banking), Santander credit card
- Securo version: v0.14.3

## Additional Context

Found while comparing my own card's "Total da fatura" against my previous PFM tool and the bank's own bill: the numbers didn't match, and tracing it down led to the category-exclusion behavior above. Adjacent to #473 in the sense that both are about Pluggy-connected Brazilian credit card reporting being skewed by a shared calculation helper, but a distinct root cause (category exclusion vs. double-subtracted future installments).

vhsantos26:
## Description

Fixes #647.

A credit card's cycle "Total da fatura" was computed with `counts_as_pnl()`, the same filter used for personal P&L totals, which excludes transactions whose category is `treat_as_transfer` or `is_ignored`. That's the right rule for discretionary-spending totals, but not for a bill total: a charge doesn't stop being owed to the bank just because the user later tagged it under a transfer/investment category for budgeting purposes.

Concretely, on my own workspace a R$2,880.18 real-estate consortium charge on a Santander card, categorized under the default "Transferências" category, was silently excluded from the cycle's "Total da fatura", while the independently-synced `credit_card_bills.total_amount` for the same due date (fed directly from the provider's bill data) correctly showed the total including that charge.

## Change

Adds `counts_toward_bill()` in `_query_filters.py`, alongside `counts_as_pnl()`. It keeps the same paired-transfer and settlement-debit exclusions (those really are the same money moving, not new debt), but drops the category-level `treat_as_transfer`/`is_ignored` exclusion. `get_account_summary()`'s credit-card expense calculation now uses this instead of `counts_as_pnl()`.

`counts_as_pnl()` itself is untouched, so personal-expense/dashboard totals keep excluding transfer-categorized transactions as before; this only changes the credit-card bill total.

## Testing

- Full backend test suite passes locally (`uv run pytest`).
- Verified against my own Pluggy-connected Santander card: the affected cycle's "Total da fatura" went from R$20,718.07 to R$23,097.93 after this change, matching the provider's own bill total (`credit_card_bills.total_amount` = R$23,972.93) much more closely than before (the remaining ~R$875 gap is unrelated: see #648, a separate transfer-pair-matching false positive that also happened to land in this same cycle).

## Screenshots

N/A (backend aggregation fix, no UI changes).

Credit-card bill totals now include transfer-categorized charges while still excluding ignored transactions and payment credits. Forecast totals use the same bill-specific rules as posted totals.

- Charges categorized as transfers remain included in the bill total.
- Ignored charges remain excluded.
- Card payments continue to reduce the bill total.
- Personal finance totals keep their existing exclusions.

Risk: money math.


tassionoronha:
Thanks for this, and for the write-up. The diagnosis is right: `counts_as_pnl()` answers "is this discretionary spending", which is not the question a bill total asks, and #647 is a real bug.

I checked out the branch and wrote three cases against it and against `main`. They're in the file below, self-contained on the existing `session` / `test_user` / `test_workspace` / `test_connection` fixtures. Drop it in `backend/tests/` and run `pytest tests/repro_649.py`. A card that closes on the 11th and is due on the 16th, one cycle window, round numbers:

| | main | this PR |
|---|---|---|
| `test_payment_credit_must_not_shrink_the_bill` — 100 charge + 80 unpaired credit in a `treat_as_transfer` category | 100.00 ✅ | **20.00** ❌ |
| `test_pending_charge_counts_toward_the_bill` — 100 posted + 300 **pending** in a `treat_as_transfer` category | 100.00 ❌ | **100.00** ❌ |
| `test_posted_charge_counts_toward_the_bill` — 100 posted + 300 **posted** in a `treat_as_transfer` category | 100.00 ❌ | 400.00 ✅ |

The third one is your fix and it works. The other two need another pass.

**1. Card payments now net out of the bill.** The credit-card branch sums every type through `signed_for_bill`, so credits subtract. The comment this PR replaces was load-bearing there: `counts_as_pnl already excludes ... transfer-like categories, so bill payments are not double-counted`. The paired-transfer exclusion doesn't cover it, because `transfer_detection_service` only pairs on identical absolute amount, different account, within ±2 days. A payment stays unpaired whenever the paying account isn't in Securo, the amount doesn't match exactly, or the payment was partial, which is the normal case for anyone who connected only the card. Those used to be dropped by category and now subtract in full.

The underlying rule is that a bill total counts new debt, not repayments, so the category exclusion probably needs to stay on the credit side and only go away for debits.

**2. The forecast query wasn't switched.** `account-detail.tsx` renders "Total da fatura" from `projected_expenses`, which is `monthly_expenses + forecast_expenses`, and `forecast_expense_result` still uses `counts_as_pnl()` (`account_service.py:873-878`; `forecast_income` at :858 has the same asymmetry). So a charge in a transfer-flagged category is still excluded while pending and then jumps into the total once it posts. The in-progress cycle is the one people look at most, so it's worth covering in the same PR.

Two smaller things:

- Could you add tests? `test_credit_card_accounting_mode.py` already has the fixtures for this, and the first and third cases above would pin both directions.
- The new filter also stops excluding category-level `is_ignored`, which is wider than #647 needs. A category the user explicitly marked ignored is a direct "don't count this", and those rows still render with an ignored badge via `TransactionRead.reflect_ignored_category` while counting toward the total. I'd keep `Category.is_ignored` in the exclusion and drop only `treat_as_transfer`.

Happy to look again once the first two are covered.

vhsantos26:
Pushed a follow-up commit covering both gaps:

- `counts_toward_bill()` now only drops `treat_as_transfer` on credits, not debits. An unpaired card payment in that category still nets against the bill; a charge tagged the same way no longer does.
- `forecast_expense_result` (credit-card branch) switched from `counts_as_pnl()` to `counts_toward_bill()`, so a pending transfer-tagged charge counts toward the in-progress cycle instead of waiting to post.
- Kept `Category.is_ignored` excluding on both sides, as you suggested.

Added the three repro cases plus one for `is_ignored` to `test_account_service.py`. Full suite passes locally.

Ready for another look.

tassionoronha:
Saw your BACKLOG note, so I'll ask it here. The credit path leans on `treat_as_transfer` to decide a bill total, which is the same job this PR takes away from that flag on the debit side. Would a nature field (purchase / refund / payment) settle it for good? The debit half looks right to me either way.

tassionoronha:
Also please remove the backlog.md

vhsantos26:
> The credit path leans on `treat_as_transfer` to decide a bill total, which is the same job this PR takes away from that flag on the debit side. Would a nature field (purchase / refund / payment) settle it for good?

That's the real fix, yeah. `treat_as_transfer` deciding the credit side has the same category-as-proxy problem this PR removes from the debit side, just moved over by one field. I'd rather do it as its own PR, it touches the transaction model and probably needs the Pluggy subtype mapping to back it properly. Opened #848 to track it instead of the backlog file. Debit half stands either way.

vhsantos26:
Fair, dropping it from this PR. Moved the entry to #848 so it's not lost.

tassionoronha:
Now it LGTM! Thanks @vhsantos26 !

PR #653, opened 2026-08-21
feat: add dutch translations

thijserven:
## What

Adds dutch translations to the app.

## Why

So native dutch speakers who don't speak english can use the app.

## How to Test

Steps to verify the changes work:

1. Run the development stack
2. Switch the language to: "Nederlands"

## Checklist

- [x] Backend tests pass (`pytest`)
- [x] Frontend lints clean (`npm run lint`)
- [x] Frontend builds (`npm run build`)
- [x] Translations updated (if user-facing strings changed)
- [x] For a large or core change, I discussed it first (issue or Discord) so we could align before the code (see [CONTRIBUTING](../CONTRIBUTING.md#before-large-or-core-changes))

PR #654, opened 2026-08-21
fix: model institutions as first-class rows so SimpleFIN accounts show their own bank

### LINKED ISSUE #345: [Bug] SimpleFIN: accounts from multiple institutions collapse into one connection and share a single institution name/logo
### Summary
When a single SimpleFIN connection returns accounts belonging to **several different
institutions**, Securo files them all under **one** `bank_connections` row and applies
that one connection's `institution_name` + `logo_url` to every account. Because the
account-level data has no institution/logo of its own, the UI shows the **same bank
icon for every account**, even though they are at different institutions.

SimpleFIN provides a distinct `org` (name + domain) **per account** in its `/accounts`
response, so the information needed to differentiate is available but currently discarded.

### Environment
- Securo backend `0.13.5` (image `ghcr.io/securo-finance/securo-backend:latest`, rev `3eae2eb`)
- Self-hosted via docker-compose (prod images)
- Provider: SimpleFIN (production bridge, `https://bridge.simplefin.org`)

### Steps to reproduce
1. In a SimpleFIN Bridge account, connect banks/brokerages across **multiple
   institutions** (e.g. a checking account at one bank + brokerage/401(k) accounts
   at other institutions).
2. Connect that single SimpleFIN setup token to Securo (Accounts → Connect Bank → SimpleFIN).
3. Sync. Observe the imported accounts.

### Actual behavior
- Exactly **one** `bank_connections` row is created (`provider = simplefin`), with a
  single `institution_name` and a single `logo_url`
  (`https://www.google.com/s2/favicons?domain=<one-domain>&sz=128`).
- **All** imported accounts link to that one connection via `accounts.connection_id`.
- In the UI every account shows the **same** institution icon, regardless of which
  institution the account is actually at.

Illustrative (anonymized) data — 13 accounts, all tagged with the *first* institution:

| account name (anonymized) | actual institution | shown logo |
|---|---|---|
| Checking (••••) | Bank A | Bank A |
| Performance Savings (••••) | Bank A | Bank A |
| Employer 401(k) Plan (••••) | Brokerage B | Bank A ❌ |
| Individual Brokerage (••••) | Brokerage C | Bank A ❌ |
| Managed Account / UMA (••••) | Brokerage D | Bank A ❌ |
| … (9 more across several institutions) | various | Bank A ❌ |

### Expected behavior
Accounts should reflect **their own** institution's name and logo. Either:
- **(preferred)** create/maintain one connection per distinct SimpleFIN `org` and
  attach each account to the matching connection, **or**
- store the `org` name + domain (and derived `logo_url`) **per account**, and render
  the account's own institution in the UI.

### Notes
- The `org` (and its `domain`) is present on each account object in the SimpleFIN
  `/accounts` payload, so no extra API calls are required to derive per-account logos.
- Cosmetic only — balances/transactions sync correctly; this is purely about
  institution identity/iconography.

mverrilli:
## Summary

A single SimpleFIN Setup Token can cover several linked institutions (bank + brokerages, etc.), but Securo stored one `institution_name`/`logo_url` per connection and stamped it on every account under it — so accounts at different real banks all showed the same name and logo, and investment holdings all landed in one wallet named after the connection.

This models the hierarchy properly — **a connection has institutions, and each account points at its own**:

- New `institutions` table (`connection_id`, `external_id`, `name`, `logo_url`); `accounts` gains an `institution_id` FK (migration **074**, additive/nullable — existing installs keep working unchanged and institution rows appear on the next sync).
- **Institutions match by the provider's stable org id** (SimpleFIN `conn_id`, stored as `external_id`), so a bank renamed on the provider side updates its row in place. Identity is `(connection_id, external_id)` when the id exists and `(connection_id, name)` otherwise — two partial unique indexes, so two same-named orgs (two logins at one bank) stay distinct rows and two racing syncs can't double-insert. Rows created before a server sent org ids are adopted by the first id-carrying hint with the same name. The Bridge's `connections[]` is preferred, with a fallback to the spec-style per-account `org` object (mirroring `_org_website`) so servers without `connections[]` still work.
- A narrow reap runs each sync for rows referenced by **neither accounts nor wallets**: id-carrying servers never orphan a row, but a name-only server that renames its bank mints a fresh row and abandons the old one, which would otherwise keep a single-bank link presenting as multi-institution forever. The both-refs guard means a wallet's "Synced from" label can never be taken away.
- `serialize_account` resolves institution name and logo **as a pair** from the account's institution. The institution only outranks `connection.display_name` when the link spans **more than one** institution — renaming a single-bank SimpleFIN link relabels its accounts like before. The logo likewise falls back to the connection's on single-institution links (same bank); on multi links a missing favicon still beats another bank's icon.
- **Investment wallets get per-account granularity**: holdings carry their owning account, so each investment account becomes its own wallet ("401(k)" apart from "Rollover IRA"), named after the account and backed by its institution for the "Synced from …" subtitle. Sync re-attributes assets only between its own wallets — a holding the user moved to a custom wallet is never touched — and deletes the sync-owned wallets it empties.
- The connections API exposes the `institutions` list, and the accounts-page card labels multi-institution links by their provider ("SimpleFIN · 9 institutions", localized in all 10 languages with proper plural forms incl. the Cyrillic "one" category) with a generic link icon.
- Also fixes a separate bug in the MCP `list_accounts` tool that always returned `institution: null` due to a dict key mismatch.

`bank_connections.institution_name`/`logo_url` are retained as the link's claim-time label and single-institution fallback; dropping them touches ~50 files across providers, API, frontend, and tests, and is left as a follow-up if wanted.

## Testing

- Backend suite with CI flags on the rebased branch: **3,060 passed, 45 skipped, coverage 92.35%** (4 pre-existing `tests/test_config.py` failures excluded — they also fail on unmodified `main` in this environment; env-sensitive secrets-dir tests, unrelated). `ruff` and `ty` clean.
- New/updated tests: org-id matching incl. rename-in-place, legacy-row adoption, and same-named-orgs-stay-distinct; the full `_institution` precedence matrix (multi-gated rename, single-link logo fallback); per-account `org` fallback parsing; a sync-level integration test covering account `institution_id` stamping/backfill, the orphan reap (and that wallet-referenced rows survive it), and `BankConnectionRead.institutions`; per-account wallet split/re-attribution/cleanup.
- Frontend: `tsc`, eslint, production build, and full vitest suite (132) clean, including locale-parity tests.
- Verified live against a real SimpleFIN Bridge connection spanning 9 institutions: all 9 institution rows carry their MX org ids, all 12 accounts link correctly, every surface (DB, REST API, MCP tool) shows each account's own institution, and holdings sit in per-account wallets (401(k): 4 positions, Rollover IRA: 10) backed by the right institution.

Closes #345


tassionoronha:
Thanks so much for this @mverrilli, and especially for testing it against a real 9-institution link. That kind of verification is really appreciated. The core idea is right: institutions should be their own rows that accounts point at, not one name copied onto the connection. I checked how Actual models this and they landed on the same shape, so the direction is solid.

A few things before we can merge:

Migration collision (blocker). 073_account_institution.py uses revision 073, but main already has 073_import_logs_for_asset_orders.py with the same id (it came in with #651, which your branch merged). Alembic ends up with two heads and upgrade head fails, so nothing migrates. Renumbering to 074 with down_revision = "073" fixes it.

Match institutions by conn_id, not by name. Right now _resolve_institution keys on the institution's name. That means when a bank gets renamed on the provider side, a brand new row is created and the old one is swept away by the cleanup. Since asset_groups.institution_id is ON DELETE SET NULL, the wallet loses its institution and its "Synced from" label falls back to the wrong bank, which is the same problem this PR is fixing.

The nice part is you already read conn_id in _institutions_by_conn_id. If you store it on the institution row and match on it, a rename becomes a simple name update, and the stale-institution cleanup isn't needed at all. That also removes the risk of deleting a wallet's institution.

Two small things in _institution. An account's institution currently always wins over connection.display_name, so renaming a connection quietly stops working even on single-bank SimpleFIN links. And if the institution row has no logo, the account's icon disappears instead of falling back to the connection's. Could we only let the institution override when the connection actually has more than one, and keep the logo fallback?

Small one: _institutions_by_conn_id reads only the top-level connections[], while the logo helper just above also falls back to each account's org object. On a SimpleFIN server that doesn't send connections[] the feature would silently do nothing, so the same fallback would be good here.

Really nice work on this one.

mverrilli:
@tassionoronha Makes sense, thanks for the review.  I should not have merged in the UI, it bites me every time (the migration collision). 

I'll fix these and ping you when ready and tested.

mverrilli:
@tassionoronha (AI response below, however I did also test this manually). 

Ready for another look — all four points addressed in 458cabe (the branch is also rebased onto current `main`, which replaces the accidental UI merge commit):

**Migration collision** — renumbered to `074` with `down_revision = "073"`. Verified the chain applies cleanly from 072 through your 073 into 074, single head.

**Match by conn_id** — done. `institutions.external_id` stores the org id, and `_resolve_institution` matches on it first, so a provider-side rename is now just a name update on the same row (there's a test that does exactly that: rename across two syncs, same row id). Rows created before a server sent ids get adopted by the first id-carrying hint with the same name, so nothing gets orphaned on upgrade. One wrinkle this surfaced: identity became `(connection_id, external_id)` when the id exists and `(connection_id, name)` otherwise — two partial unique indexes — because two logins at the same bank produce two orgs with identical names, and a single name-unique constraint would reject them.

I did keep one much narrower reap, and want to flag it rather than sneak it in: a server that only names its orgs (no ids — the same spec-style servers the `org` fallback exists for) still mints a new row on rename and abandons the old one, and since the multi-institution gate counts the link's institution rows, that abandoned row would permanently flip a single-bank link into multi presentation — disabling exactly the rename/logo behavior you asked to restore. The reap only deletes rows referenced by **neither** accounts **nor** wallets, so the `SET NULL` hazard you called out can't happen anymore — a wallet's institution is never deletable. If you'd still rather drop it entirely I will; it's one self-contained block.

**Precedence** — the institution now only outranks `display_name` when the link spans more than one institution, so renaming a single-bank SimpleFIN link relabels its accounts again. The logo falls back to the connection's on single-institution links. On multi links I left the fallback out deliberately: the connection's logo is derived from the *first* `connections[]` entry's website, so on a multi link it's specifically another bank's favicon — falling back there would reintroduce the wrong-icon pairing this PR fixes. A missing favicon renders as the account-type icon, which felt like the lesser evil. Happy to change if you read it differently.

**`org` object fallback** — `_account_institution_hint` now prefers `connections[]` by `conn_id` and falls back to the per-account `org` object (name/domain, `org.id`/domain as the identity), mirroring `_org_website`. Covered by a parse test with no `connections[]` at all.

Re-verified end to end: full suite green with your new main merged in (3,060 passed, coverage 92.35%, ruff/ty clean), and a live resync of the 9-institution link now shows every institution row carrying its MX org id.

mverrilli:
Fixed some edge cases and indexes.

tassionoronha:
Thanks @mverrilli, the four points from last round are all addressed, I checked a multi-institution link locally and it looks right.

main moved under the branch and broke two things, so I pushed a fix to your branch: #477 landed another migration numbered 074, and #653 added nl.json without the new key. Now 075, CI is green.

One blocker left. On the first sync after this merges, an existing user loses the wallet they already have. _wallet_for keys on f"{connection.external_id}::{account_id}", which never matches the legacy wallet's plain connection.external_id, so sync creates a new one and the cleanup deletes the old as empty. I reproduced it: a wallet renamed with a custom icon, color and position came back as "Rollover IRA" with defaults. Happens even with a single investment account. Could we adopt the legacy row and re-key it instead of minting a new one?

Two smaller ones: _clean_logo_url truncating to 500 chars stores a broken URL (better to drop it), and the institution logo never updates after a provider-side rename while the name does.

Also worth a code comment that re-attribution now moves holdings between sync-owned wallets, not just null ones.

WDYT? :)

mverrilli:
@tassionoronha Thanks! 

All addressed:

- Legacy wallet blocker: _wallet_for now adopts and re-keys the legacy row in place instead of minting, so a renamed/customized wallet survives the upgrade. It extends to stale keys after a reconnect too, with provenance checks so it never adopts the wrong bank's wallet.
- _clean_logo_url: drops URLs over 500 chars now instead of storing a truncated/broken one.
- Logo on rename: a provider-side rename updates the stored logo alongside the name.
- Re-attribution comment: added, noting it now moves holdings between sync-owned wallets, not just null groups.
- Added more tests and edge cases.

tassionoronha:
Amazing work here @mverrilli!
Thanks for seeing this through :)
We're now good to merge.

PR #659, opened 2026-08-22
feat: add Akahu bank connections

cjsewell:
## Summary

- add an opt-in Akahu provider for New Zealand bank connections using Personal App tokens
- sync accounts and posted transactions through the existing token-provider flow
- add setup guidance and Docker Compose configuration

## Verification

- 17 Akahu provider tests pass on Python 3.11
- Ruff passes
- ty check passes
- frontend lint passes with zero errors
- frontend production build passes

Adds an opt-in Akahu provider for New Zealand bank connections. It uses Personal App tokens to sync accounts, balances, and posted transactions.

- Users can connect Akahu with an App ID and User token.
- Users can sync accounts, balances, and posted transactions.
- Transaction ledger dates remain unchanged during synchronization.
- Setup instructions and Docker Compose configuration are included.

Risk: auth, money math, and sync provider.


tassionoronha:
I merged main locally and the frontend checks fail on i18n parity: the Akahu strings only exist in en.json and we now keep 15 locales, all of which need every key. Machine translation is fine for provider copy; pt-BR is the one worth doing by hand. Once that is in I can update the branch and get CI going.

PR #661, opened 2026-08-22
fix: apply rules to generated recurring transactions

### LINKED ISSUE #568: [Bug] Rules are not applied when recurring transactions are generated
## Description

Transactions generated from a recurring transaction do not go through the existing transaction rules.

As a result, rule actions such as setting a payee or appending notes are not applied to generated occurrences, even when an active rule matches the recurring transaction's description and other fields.

## Steps to Reproduce

1. Create an active rule that matches a transaction description, for example `Internet Provider`.
2. Configure the rule to set a payee and/or append notes.
3. Create an auto-generated recurring transaction with the matching description.
4. Allow Securo to generate an occurrence of the recurring transaction.
5. Open the generated transaction.

## Expected Behavior

The generated transaction should be evaluated against the active transaction rules, just like a manually created, imported, or synchronized transaction.

Matching actions should be applied. For example, the generated transaction should contain the payee and notes configured by the rule.

## Actual Behavior

The recurring occurrence is created without the payee or notes configured by the matching rule.

The existing rule is not applied to the generated transaction.

## Environment

- Transaction source: recurring
- Recurring mode: auto-generate enabled
- Rule actions observed: `set_payee`, `append_notes`

seraphblade2010:
## Summary

Generated recurring occurrences currently bypass the transaction rule engine, so matching rules do not apply actions such as setting a payee or appending notes.

This applies the existing rule flow when a new recurring occurrence is generated while preserving reconciliation behavior for connected accounts.

## Changes

- apply active workspace rules to newly generated recurring occurrences;
- keep already-existing real transactions unchanged when they are linked to a recurring definition;
- preserve the pre-rule description as a recurring-match signal when `set_description` normalized a generated occurrence;
- defer automatic `ignore` on pending connected-account placeholders so they remain eligible for reconciliation;
- restore that deferred ignore decision when sync/import promotes the placeholder;
- keep explicit user-ignore match-blocking;
- retain workspace isolation, rule priority/chaining, and nested condition-group behavior.

Deferred ignore is evaluated using the active rule set at promotion time; generation-time rule state is not snapshotted.

## Testing

Regression coverage includes:

- `set_payee` and `append_notes` on generated occurrences;
- inactive-rule exclusion;
- workspace isolation;
- duplicate prevention;
- existing-real-transaction behavior;
- rule-managed description reconciliation;
- deferred ignore through sync and import;
- explicit user-ignore preventing reconciliation.

Validation after rebasing onto current `main`:

- relevant recurring/rule suite: 137 passed;
- full backend suite: 3,051 passed, 45 skipped;
- Ruff: passed;
- ty: passed;
- fork CI: passed.

Closes #568

### DIFF
--- a/backend/app/services/connection_service.py
+++ b/backend/app/services/connection_service.py
@@ -37,7 +37,11 @@ from app.services.account_service import (
 from app.services.asset_group_service import ensure_group_for_connection
 from app.services.credit_card_service import apply_effective_date
 from app.services.rule_engine import merge_notes
-from app.services.rule_service import apply_rules_to_transaction, preview_rules_for_transaction
+from app.services.rule_service import (
+    apply_rules_to_transaction,
+    preview_rules_for_transaction,
+    recurring_placeholder_rule_ignore,
+)
 from app.services.transfer_detection_service import detect_transfer_pairs
 from app.services.fx_rate_service import stamp_primary_amount
 from app.services.payee_service import get_or_create_payee
@@ -1479,6 +1483,11 @@ async def sync_connection(
                 if placeholder:
                     if placeholder.is_ignored:
                         continue
+                    deferred_rule_ignore = (
+                        await recurring_placeholder_rule_ignore(
+                            session, user_id, placeholder
+                        )
+                    )
                     placeholder.external_id = txn_data.external_id
                     placeholder.source = "sync"
                     placeholder.status = txn_data.status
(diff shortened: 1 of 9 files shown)


panga:
Could we scope this lookup by recurring ID and workspace instead of user_id? If another workspace member syncs the account, recurring_placeholder_rule_ignore returns false and the deferred ignore is lost.

PR #663, opened 2026-08-22
fix: reject unsafe regex rule patterns

### LINKED ISSUE #565: [Bug] regex validation / rule-matching
## Description

When creating an expense categorization rule using a regular expression, an invalid/incomplete regex can cause the rule to match **all expenses**.

In my case, I accidentally left a `|` at the end of the regex. Instead of rejecting the regex or only matching the intended pattern, the rule was applied to every record in the database.

This is particularly dangerous because applying the rule resulted in unintended categorization of existing expenses.

## Steps to Reproduce

1. Create a rule to automatically apply a category to expenses using a regex condition.
2. Enter a regex with a trailing `|`, for example:

   ```regex
   foo|
   ```
3. Save the rule.
4. Apply/run the rule against existing transactions.
5. Observe that the rule matches and applies the category to all records.

## Expected Behavior

The application should either:

* Reject the regex as invalid/incomplete and prevent the rule from being saved, or
* Treat the invalid regex as non-matching and apply the rule to no records.

A regex such as `foo|` should **not** match every expense unless that is explicitly intended.

## Actual Behavior

The rule was accepted and matched **all records in the database**, causing the category to be applied to every expense.

This can potentially result in large-scale unintended modifications to existing data from a simple regex typo.

## Environment

* Deployment: Docker
* Browser: Firefox 153
* OS: Windows

## Additional Context

The issue was caused by accidentally leaving the `|` character at the end of the regex.

In regular expressions, a trailing `|` represents an empty alternative. Depending on how the regex is evaluated, this effectively makes the expression match an empty string, which means it can match every record.

It may be worth validating regex rules before allowing them to be saved or executed, and/or showing a warning when a pattern contains constructs that can match an empty string.

Ideally, there should also be some protection against a rule unintentionally modifying all existing records.

seraphblade2010:
## Summary

Regex rule conditions currently accept patterns that can produce zero-length matches.

For example, `foo|` is valid Python regex syntax: the trailing alternative is empty, so `re.search()` succeeds even for unrelated transaction descriptions. When such a rule is applied to existing transactions, that can result in unintended bulk changes.

This also adds save-time validation for malformed regex patterns and protects existing persisted unsafe rules at runtime.

## Changes

- centralize rule-regex compilation using the same normalization and `re.IGNORECASE` semantics as runtime matching;
- reject malformed regex patterns when creating, updating, or importing rules;
- reject regex patterns that can match the empty string;
- apply the same validation to regex leaves inside grouped conditions;
- treat malformed or empty-matching legacy rules as non-matching instead of raising;
- reuse the validated compiled regex for the actual transaction match.

No migration is required. Existing unsafe definitions remain stored but inert until corrected.

## Safety behavior

Patterns capable of matching the empty string are rejected, including examples such as:

- `foo|`
- `a*`
- `.*`
- `^$`

Normal selective regex patterns continue to work, including:

- `foo|bar`
- `^NETFLIX`
- `foo.*`
- `.+`

## Testing

Regression coverage verifies:

- trailing, leading, and grouped empty alternatives;
- nullable quantifiers and zero-width anchors;
- malformed regex patterns;
- legacy persisted unsafe-rule behavior;
- grouped-condition validation;
- rejection before rule persistence or historical application;
- atomic update rejection;
- JSON import partial-skip behavior;
- safe regex compatibility.

Validation:

- targeted rule tests: 168 passed;
- broader rule tests: 65 passed;
- full backend suite: 3,070 passed, 45 skipped;
- Ruff: passed;
- ty: passed;
- fork CI: passed.

Closes #565

### DIFF
--- a/backend/app/services/rule_engine.py
+++ b/backend/app/services/rule_engine.py
@@ -16,6 +16,18 @@ def _strip_accents(text: str) -> str:
     return "".join(c for c in nfkd if not unicodedata.combining(c))
 
 
+def compile_rule_regex(pattern: str) -> re.Pattern[str]:
+    """Compile a regex using Securo's runtime normalization and safety policy."""
+    effective_pattern = _strip_accents(pattern)
+    try:
+        compiled = re.compile(effective_pattern, re.IGNORECASE)
+    except re.error as exc:
+        raise ValueError("Invalid regular expression") from exc
+    if compiled.search("") is not None:
+        raise ValueError("Regular expression must not match an empty string")
+    return compiled
+
+
 def _normalize(text: str) -> str:
     """Normalize text: uppercase and remove diacritics (accents)."""
     return _strip_accents(text.upper())
@@ -80,9 +92,9 @@ def _match_condition(condition: dict, tx: "Transaction") -> bool:
                 # normalized text, but keep its case: uppercasing a regex
                 # inverts escape classes (\s -> \S, \b -> \B, \d -> \D) and
                 # breaks inline flags. Case is already handled by IGNORECASE.
-                pattern = _strip_accents(str(value or ""))
-                return bool(re.search(pattern, tx_str, re.IGNORECASE))
-            except re.error:
+                compiled = compile_rule_regex(str(value or ""))
+                return compiled.search(tx_str) is not None
+            except ValueError:
                 return False
 
     # Numeric operators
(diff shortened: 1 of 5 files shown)


tassionoronha:
LGTM, thanks @seraphblade2010 !! 💪🏼

Issue #642, opened 2026-08-20
[Feature] Add rule preview to show matching transactions before applying


suzarte:
## Problem

When creating a rule, it is currently not possible to preview which transactions match the configured conditions before saving or applying the rule.

This can make it difficult to verify whether the conditions are correctly defined, especially when the rule is going to be applied to existing transactions.

For example, when using a regex condition such as:

```text
^FTD LUMISFERA [1-6]/6$
```

I would like to confirm that the rule matches exactly the six expected installment transactions before applying actions such as changing the category, adding tags, or changing the description.

However, this is not only useful for regex conditions. The same problem exists with any type of rule condition, including `contains`, `starts with`, multiple `AND`/`OR` conditions, and combinations of different conditions.

## Proposed Solution

Add a preview/test option to the rule creation and editing interface that evaluates the current conditions against existing transactions before the rule is saved or applied.

For example, a button such as:

**Preview matches** / **Test rule**

could display:

```text
6 transactions match this rule

01/08/2026  FTD LUMISFERA 1/6  R$ 120.00
08/08/2026  FTD LUMISFERA 2/6  R$ 120.00
15/08/2026  FTD LUMISFERA 3/6  R$ 120.00
22/08/2026  FTD LUMISFERA 4/6  R$ 120.00
29/08/2026  FTD LUMISFERA 5/6  R$ 120.00
05/09/2026  FTD LUMISFERA 6/6  R$ 120.00
```

It would also be useful to show a summary of the actions that would be applied, for example:

```text
Category → Education
Tags → #materialEscolar
Description → FTD - Material Escolar
```

This would allow the user to verify both the matching transactions and the resulting changes before committing them.

The preview should work with all supported condition types, not only regex.

It could also be particularly useful when the option **"Apply matching actions to existing transactions"** is enabled, showing something like:

> **6 existing transactions will be affected.**

Ideally, the user could click the result count or a **View transactions** action to inspect the matching transactions in more detail.

## Alternatives Considered

Currently, the alternative is to save the rule and then inspect the affected transactions afterward.

This works, but it is less convenient and can be risky when the rule modifies existing transactions. If the condition is too broad, the user may unintentionally modify transactions before realizing the mistake.

Another alternative is to manually search the Transactions page using the same criteria, but this becomes difficult for complex conditions, particularly when using regex or multiple `AND`/`OR` conditions.

[...]

PR #669, opened 2026-08-23
feat(rules): preview what a rule would match before saving it

### LINKED ISSUE #609: [Feature] Rule preview

## Problem

When creating new rules it is not always clear the transactions that will be impacted.

## Proposed Solution

Add a collapsible table in the rule creation showing all the transactions that will be affected by the rule.

Image

Image

## Additional Context

I've vibe coded it on my fork and has been really helpful.

paulovitor6:
Writing a rule was a guess: you saved it, ran apply-all, and found out afterwards whether the conditions were too broad. The editor now has a "Preview" panel that lists the existing transactions the draft would match and the category each would end up in.

POST /api/rules/preview is the read-only counterpart of apply_single_rule: same transaction scope, same engine, same overwrite semantics, including the retry against original_description — so what the table shows is what saving the rule would do. It writes nothing, hence the entry in the write permission allowlist.

## What

Adds a preview rule collapsible table to help visualize the impact of a new rule or changes on a existing rule.

## Why

When creating complex rules it is easy to create a rule that is really broad and breaks our transaction mapping. This helps validate the rules before creating them.

## How to Test

1. ...
2. ...
image

image

image

image

## Related Issue

Closes #609 and #642

## Checklist


tassionoronha:
Thanks. Checked the two things I expected to be problems and both are handled: the preview writes nothing (`_rule_preview` is detached, and `payee` is a column not a relationship, so no cascade), and it doesn't fire per keystroke because the `useEffect` collapses the panel on any condition/action edit. Worth keeping that comment. Backend suite and frontend both green, locales complete.

**Blocking: the sample can't be paged.** The backend takes a `limit` (max 100), the frontend declares it in the `preview()` type but never sends it, and there's no offset, so it's always 20 rows. The counts are exact, but on a rule reporting four figures of changes you inspect a fraction of a percent, and a broad rule is exactly the case this feature exists to catch. A preview you can't scroll through doesn't answer the question it's asked. `preview_rule` already evaluates every transaction, so raising the cap or adding an offset costs only serialization.

Non-blocking, but related: `preview_rule` selects every transaction with no limit and evaluates in Python. That mirrors `apply_single_rule`, but apply-all is a deliberate one-off while the preview is opened while editing. Pushing the simple conditions into SQL, or loading only the columns `_rule_preview` copies, would lower the ceiling before paging makes it hotter.

Smaller: `_validate_rule_definition(...)` is called with an empty action list, so a malformed action reaches `apply_rule_actions` unvalidated. The save path validates both.

tassionoronha:
Re-checked `34e1044`. The paging blocker is resolved, and you went further than I asked:

- `offset` on both sides, with `Transaction.id.desc()` added as a tiebreaker so windows stay stable across requests. That was the failure mode I'd have flagged next.
- `load_only(*_PREVIEW_COLUMNS)` keeps `raw_data` out of the scan, which was the cheap half of the unbounded-scan concern.
- `_validate_rule_definition` now receives the actions.

Verified in the browser: the panel shows 50 rows with "load more", clicking it goes to 100 with the counts unchanged and no flicker on the table. Frontend green (163 tests), rules tests pass, branch current with `main`.

No further blockers from me.

PR #687, opened 2026-08-24
fix: clarify dashboard invested assets label

### LINKED ISSUE #686: Clarify the dashboard asset metric as invested assets
The dashboard metric currently labelled as assets or patrimony represents the value of registered assets, not the user's complete net worth. Naming it as invested assets makes the distinction from the account balance and net-worth reports clear, without changing any calculation.

vhsantos26:
Closes #686. Renames the dashboard metric to make clear that it represents registered invested assets rather than the user's complete net worth. This changes no calculations and updates the existing label consistently across all supported locales.

### DIFF
--- a/frontend/src/locales/de.json
+++ b/frontend/src/locales/de.json
@@ -286,7 +286,7 @@
     "categorizeNow": "Jetzt kategorisieren",
     "allCategorized": "Alles erledigt!",
     "allCategorizedDesc": "Alle Transaktionen sind kategorisiert",
-    "assetsValue": "Vermögenswerte",
+    "assetsValue": "Investiertes Vermögen",
     "currencyBreakdown": "Aufteilung nach Währung",
     "convertedFrom": "Umgerechnet von {{currency}}",
     "fxRateUsed": "Wechselkurs: {{rate}}",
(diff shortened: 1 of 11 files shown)


tassionoronha:
Thanks @vhsantos26. This one I'd like to hold, and there's a concrete reason.

The metric isn't invested-only. get_asset_values_at sums every active asset with no filter on type (asset_service.py:940-946), and Asset.type is one of real_estate, vehicle, valuable, investment, other. So someone with a house and a car registered would see both counted under "Invested assets". The label would be less accurate than what's there now, not more.


You did spot something real though: "Patrimônio" in PT-BR does overclaim, since it reads as net worth and this is only registered assets. That's worth fixing. "Invested" just isn't the repair, because it narrows to one of five types.

Same as on #689: a metric label is product vocabulary, it lands in eleven languages and carries into docs and screenshots, so I'd rather settle it in the open. Bring it to the Discord or make the case here and let's pick a term together.

vhsantos26:
You're right, I checked `asset_service.py:940-946` and `get_asset_values_at` really sums every type with no filter, so "Invested" overclaims just as much as "Patrimônio" did. What about **"Registered assets"**? Fixes the net-worth read without pinning it to one type. Happy to push that update, or take it to Discord first if you'd rather get more eyes on it.

tassionoronha:
LGTM! Thanks again @vhsantos26 :)

Issue #688, opened 2026-08-24
Dashboard balance chart is mislabeled and missing a legend


vhsantos26:
## Problem

The dashboard card currently titled "Balance Flow" plots the cumulative account balance for the current month against the previous month. It is not a cash-flow chart (entries vs. exits), so the title misleads users into expecting income/expense flow.

On top of that:

- The two lines (current month, solid; previous month, dashed) have no legend, so it is not obvious which is which without hovering.
- The value shown in the top-right corner is the balance change between day 1 and the latest available day of the current month, but it renders as a bare colored number with no label.

## Proposed fix

- Rename the card to "Balance Evolution".
- Add a small legend under the header identifying the current-month line and the previous-month line.
- Label the corner value as "Period change".

This is a labeling/clarity fix only; the underlying data and calculation are unchanged.

tassionoronha:
Thanks for the report, @vhsantos26! Just merged a fix in #1000, it will be in the next release. 🙌

PR #689, opened 2026-08-24
fix: clarify balance evolution chart on dashboard

vhsantos26:
## Summary

Refs #688.

- Renames the "Balance Flow" card to "Balance Evolution" across all locales.
- Labels the corner value as the period change instead of a bare colored number.
- Adds a legend distinguishing the current month (solid line) from the previous month (dashed line).

No changes to the underlying data or calculation, only labeling and a small legend for clarity.

## Test plan

- [x] `npx vitest run` (156 tests passed, including i18n key/placeholder parity checks across all locales)
- [x] `npm run build` (tsc + vite build, no errors)
- [x] `npm run lint` (0 errors; pre-existing warnings unrelated to this change)
- [x] Validated the same change on a personal fork/deployment before opening this PR


tassionoronha:
Thanks @vhsantos26! Happy to take the legend and the labelled corner value.

The rename I'd like to handle in the open, not inside a fix PR. 

A card title is product vocabulary: eleven languages, plus docs, screenshots and the demo. Bring it to the Discord or make the case here, with the reasoning and other people weighing in. You're right that "Balance Flow" is inaccurate, so there is something to discuss. Drop the rename from this one and it goes in as is.

vhsantos26:
Dropped the rename, kept the legend and the labelled corner value. Reverted `balanceFlow` in the 10 locales that had it; fr.json already used this wording before the PR, so no change there.

I'll bring the rename to Discord separately with the reasoning.

tassionoronha:
Great, now it LGTM. 
I change the PR description since the #688 we will handle in another task.
Good job :)

Issue #678, opened 2026-08-24
[Bug] Importing transactions via Enable Banking fails due to non-unique payees


kodazavr:
## Description

When trying to import bank transactions via Enable Banking integration, the following exception appears and no accounts/transactions are imported (payee name is replaced with `<payee name>`):

```
Sync failed: (sqlalchemy.dialects.postgresql.asyncpg.IntegrityError) <class 'asyncpg.exceptions.UniqueViolationError'>: duplicate key value violates unique constraint "uq_payees_workspace_id_name" DETAIL: Key (workspace_id, name)=(21ebf2f4-38d1-4093-935a-81d3cadef3ff, <payee name>) already exists. [SQL: INSERT INTO payees (id, user_id, workspace_id, name, type, source, is_favorite, notes, email, phone, address, website, created_at) VALUES ($1::UUID, $2::UUID, $3::UUID, $4::VARCHAR, $5::VARCHAR, $6::VARCHAR, $7::BOOLEAN, $8::VARCHAR, $9::VARCHAR, $10::VARCHAR, $11::VARCHAR, $12::VARCHAR, $13::TIMESTAMP WITH TIME ZONE)] [parameters: (UUID('2108ca62-6cc9-4767-b5bc-cb9e9df0ca20'), UUID('18754524-5585-4bcf-a569-cf7360205da2'), UUID('21ebf2f4-38d1-4093-935a-81d3cadef3ff'), '<payee name>', None, 'sync', False, None, None, None, None, None, datetime.datetime(2026, 8, 24, 10, 26, 42, 6311, tzinfo=datetime.timezone.utc))] (Background on this error at: https://sqlalche.me/e/20/gkpj)
```

One of the accounts appears some time after the sync though. I believe this account doesn't have transactions to the same payee, which seems to be part of the issue.

## Steps to Reproduce

1. Go to 'Accounts'
2. Set up an integration via Enable Banking
3. See the error as above
4. Click the button to resync
5. Same error appears, sync button has a warning icon now with the message "Connection error — reconnect to fix"

## Expected Behavior

No errors show up.
All synced accounts appear under the group with the bank name.
Transactions are imported.
Payees are present in the corresponding UI.

## Actual Behavior

Group with the bank name appears, but no accounts are present.
No transactions or payees added.
Error with the message above on sync.
One of the accounts appears in the list later.

## Environment

- Deployment: Kubernetes
- Version: 0.14.4
- Browser: Firefox 154
[...]

PR #694, opened 2026-08-25
ci: label PRs by risk and size, and protect the decision axes

tassionoronha:
Two workflows, both derived from work on the triage labels. Neither creates labels: `risk: *`, `size: *` and `prio: *` already exist on the repo.

## `pr-labels.yml` — applies `risk:` and `size:` to every PR

Both axes come from the **diff**, not from the linked issue. Risk is what the change reaches; size is what the review costs. They are independent: a 20-line migration is `size: S` with `risk: high`.

Deriving from the diff rather than inheriting matters here because **18 of 39 open PRs have no linked issue**, and because the issue estimates risk before the code exists.

**Size thresholds come from this repo's own distribution.** The median open PR is ~280 lines. The common 9/29/99/499 scale would land like this:

| scale | XS | S | M | L | XL |
|---|---|---|---|---|---|
| 9/29/99/499 | 4 | 2 | 2 | 17 | 14 |

Eighty percent in two buckets decides nothing. The 150/700 cut splits the queue 9 / 20 / 10.

**Generated files are excluded from the size tally.** Measured against real PRs:

| PR | raw | reviewable | without cut | with cut |
|---|---|---|---|---|
| #670 | 2464 | 4 | `size: L` | `size: S` |
| #683 | 282 | 282 | `size: M` | `size: M` |
| #661 | 744 | 744 | `size: L` | `size: L` |

#670 is a dependency bump: 2464 lines of lockfile, 4 reviewable. Counting raw would mark the most trivial PR in the repo as the largest. Covers `package-lock.json`, `pnpm-lock.yaml`, `yarn.lock`, `poetry.lock`, `uv.lock`, `Cargo.lock`, `composer.lock`, `*.lock`, `go.sum`, `*.min.js|css`, `node_modules/`, `dist/`, `*.snap`.

**Risk escalates but never downgrades** — a reviewer's manual `risk: high` survives the next push, because they saw something a file path cannot show. **Size always recomputes**, because it is arithmetic and not judgement.

## `label-guard.yml` — reverts changes to `prio:` / `risk:` from outside triage

GitHub has no per-label ACL: anyone with triage can apply any label. This reverts and explains on the issue.

Honest limits, all worth knowing before merging:
- It is **after the fact**, not prevention. The label sits applied for a few seconds.
- Anyone with write can edit this workflow. Pair it with CODEOWNERS on `.github/workflows/`.
- It ignores its own actor, otherwise the revert fires `unlabeled` and loops.

## Needs a decision before merge

- `DONOS = ['tassionoronha']` in `label-guard.yml` — add anyone else who should be able to set the axes.
- Both run on `pull_request_target` to get a write token on fork PRs, which is half the external contributions. Safe here because **neither job checks out PR code** — they only read the file list through the API. Worth confirming you are comfortable with that trigger.

https://claude.ai/code/session_01KXS7f2gtmccQFehkq4Luxw

### DIFF
--- a/.github/workflows/label-guard.yml
+++ b/.github/workflows/label-guard.yml
@@ -1,0 +1,54 @@
+name: label guard
+
+# GitHub has no per-label ACL: anyone with triage can apply any label. This
+# restores the previous state whenever someone outside the list changes one of
+# the decision axes. It is after the fact, not prevention: the label stays
+# applied for a few seconds until the job runs.
+
+on:
+  issues:
+    types: [labeled, unlabeled]
+  # pull_request_target so the token can write on fork PRs too. Safe here
+  # because this job never checks out PR code.
+  pull_request_target:
+    types: [labeled, unlabeled]
+
+permissions:
+  issues: write
+  pull-requests: write
+
+jobs:
+  revert:
+    if: >-
+      startsWith(github.event.label.name, 'prio:') ||
+      startsWith(github.event.label.name, 'risk:') ||
+      startsWith(github.event.label.name, 'roadmap:')
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/github-script@v7
+        with:
+          script: |
+            const OWNERS = ['tassionoronha'];   // who decides the axes
+
+            const actor = context.payload.sender.login;
+            // Without this it is an infinite loop: the revert itself fires
+            // `unlabeled`, which would trigger another revert.
+            if (actor === 'github-actions[bot]' || OWNERS.includes(actor)) return;
+
+            const target = context.payload.issue || context.payload.pull_request;
+            const label = context.payload.label.name;
+            const repo = context.repo;
+            const n = target.number;
+
+            if (context.payload.action === 'labeled') {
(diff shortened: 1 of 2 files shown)


tassionoronha:
Pushed `4ae0b32`: risk now reads the linked issue too.

Rather than a strict fallback, the label is the **higher of the two sources**, because neither one dominates:

- **Paths cannot rule risk out.** A change that avoids `models/` and `alembic/` can still be risky for a reason no file name shows, and triage may have caught that when it read the code.
- **The issue cannot see the implementation.** #678 was triaged `risk: medium`; the PR that fixes it arrives with a migration, which is `high`.

So each source catches what the other misses. This also composes with the existing rule: automation may raise risk, never lower it — a reviewer's manual call still wins over both.

When no issue is linked (18 of 39 open PRs today), the paths decide alone, which is your fallback case.

The comment now names which source decided, and when the diff came out riskier than triage estimated it says so — that gap is a calibration signal on the triage régua, worth reading rather than silently overwriting.

One thing to watch after merge: if `risk: high` ends up on more than ~40% of PRs, the answer is to tighten the path rules, not to drop the max.

PR #696, opened 2026-08-25
fix(frontend): make the nginx proxy work on any container platform

jrvidotti:
## Why

The frontend image hardcoded Docker's embedded DNS (`127.0.0.11`) as the nginx resolver used to
re-resolve `BACKEND_URL`. That address only exists on Docker, so on any other container platform
every `/api/` request 502s. I hit this deploying Securo on Railway, whose internal DNS is
`[fd12::10]`.

The nginx image already knows how to find the right resolver — `15-local-resolvers.envsh` reads
`/etc/resolv.conf` and brackets IPv6 addresses — it is just opt-in and was off. So instead of
adding a knob the operator has to know about, this turns that on.

## What

- **`frontend/`** — enable the image's resolver detection and fold it into `NGINX_RESOLVER` via a
  new `docker-entrypoint.d/19-resolver.envsh`, with the precedence *explicit override > platform
  resolver > Docker's embedded DNS*. Also listen on `[::]:8080`.
- **`frontend/Dockerfile`** — pin `worker_processes`. `auto` counts the *host's* cores, not the
  container's limit, so a small cloud instance started ~48 nginx workers to serve a static SPA.
- **`charts/`** — the chart mounts its own copy of the config over the image's, and it had never
  received the re-resolution fix: `proxy_pass ${BACKEND_URL};` pins the upstream IP at config
  load, so the frontend 502s after the backend pod is recreated. Now identical to
  `frontend/default.conf.template`. Also checksums that ConfigMap so editing it rolls the pods.
- **`docs/deploy-railway.md`** — new deployment guide.

**Compose is unaffected:** on a user-defined network `/etc/resolv.conf` *is* the embedded DNS, so
the rendered config stays `resolver 127.0.0.11 valid=5s;`. `NGINX_RESOLVER` remains as an
escape hatch; no new configuration is required anywhere.

## Verification

Against the actually-built image:

- Compose path renders `resolver 127.0.0.11` — byte-identical to before.
- `NGINX_RESOLVER='[fd12::10] ipv6=on'` renders `resolver [fd12::10] ipv6=on valid=5s;`.
- Frontend + a backend container on one network: `/` → `200`, `/api/` → `404` *from the backend*
  rather than a 502, i.e. DNS resolved and the proxy connected.
- 2 worker processes (was ~48); `nginx -t` passes; no errors in the logs.
- The chart's rendered nginx config is directive-for-directive equal to the image's.
  `helm lint`/`kubeconform` were not run locally — the `helm-checks` CI job covers them.

## Note

Nothing in CI parses the nginx config today, which is how the chart copy drifted. Rendering the
ConfigMap from `frontend/default.conf.template` would remove the drift class entirely, but that
file lives outside the chart directory, so it is a bigger change than belongs here.

Nginx now detects the platform DNS resolver, supports explicit `NGINX_RESOLVER` overrides, and refreshes backend addresses after IP changes. Helm also restarts frontend pods when nginx configuration changes.

- Backend requests continue working after backend pod IP changes.
- Deployments support Docker, Podman, Kubernetes, Railway, and IPv6 resolver addresses.
- Railway deployment guidance covers database, Redis, IPv6, and target port settings.


tassionoronha:
Thanks @jrvidotti !
We're now good to go. 
I removed the doc you have added, bc I will rather add that content in the formal https://docs.usesecuro.com/ instead once this is merged ;) 

Once again, thanks for pushing it. 
I'm sure it will be helpful for a lot of people.

PR #708, opened 2026-08-26
feat(payees): identify a counterparty by its document and original name, not its display name

### LINKED ISSUE #702: [Feature] Identify payees by something other than their display name, so renaming one survives the next sync
> This touches sync and the payee data model, so per CONTRIBUTING I'm opening an issue before
> writing any code.

## Problem

I connected a Brazilian bank (Sicoob) through Pluggy and let it sync. Afterwards my `/payees` list
contained entries whose **name is a CPF**, e.g. `529.982.247-25`, instead of the person's name.

In my workspace this is not an edge case: **64 of 190 payees (34%) are named after a CPF**, covering
253 transactions. Every one of them was created by sync (`source='sync'`).

What I expected: the payee to be named after the counterparty, and — more importantly — that
correcting the name by hand would stick.

It can't. `get_or_create_payee` identifies a counterparty **solely** by its display name:

```python
# backend/app/services/payee_service.py:98
lookup = select(Payee).where(func.lower(Payee.name) == name.lower())
```

So if I rename `529.982.247-25` to `Jane Doe`, the next sync arrives with the raw payee string
`529.982.247-25` again, finds nothing, and inserts a second payee. The correction cannot survive,
and one real person's history splits across two rows. Repro: rename any sync-created payee, then
resync the connection.

The display name is the one attribute a user is explicitly invited to change, and it is being used
as the primary key.

### Why the name is a document in the first place

`PluggyProvider._extract_payee` falls back to the document when the counterparty has no name:

```python
# backend/app/providers/pluggy.py:743
name = receiver.get("name")
if name:
    return name
doc = receiver.get("documentNumber")
if doc and doc.get("value"):
    return doc["value"]
```

For PIX to an **individual**, the institution systematically returns `receiver.name: null` while
still supplying the document. Companies do get a name — which is why in my data the CNPJ
counterparties are named correctly and the CPF ones are not. This is the normal path for
person-to-person transfers, not a rare gap.

A representative payload (redacted):

```json
{
  "description": "TRANSF.REALIZADA PIX SICOOB - FAV.: JANE DOE|@Transferência Pix|@...",
  "paymentData": {
    "receiver": {
      "name": null,
      "documentNumber": { "type": "CPF", "value": "529.982.247-25" }
    }
  }
}
```

### The document is available, and thrown away

It is a stable identifier, and the schema to hold it already exists. Migration
`070_payee_fiscal_data` created `payee_tax_ids` and this index:

```python
# Matching a payment to a counterparty by document is the lookup this
# table exists to make cheap.
op.create_index(
    "ix_payee_tax_ids_workspace_kind_value",
    "payee_tax_ids", ["workspace_id", "kind", "value"],
)
```

Nothing in sync or import ever writes or reads that table — `_apply_tax_ids` is only reachable from
the manual create/update API. In my production database `payee_tax_ids` has **0 rows** and that
index has never been used.

Meanwhile the data is in every payload. Across my 722 synced checking transactions:

| Signal | Present |
|---|---|
| `paymentData.receiver.documentNumber` | 688 |
| `paymentData.payer.documentNumber` | 684 |
| `paymentData.receiver.name` | 328 |

The document is available about twice as often as the name, and `transactions.raw_data` keeps the
full payload, so existing history could be backfilled without re-fetching anything.

## Proposed Solution

Two independent layers. Either one alone fixes the bug above; together they also unify a
counterparty seen through different channels.

### 1. `payees.original_name` — remember what the provider called it

A nullable `String(255)` on `payees`, set once at creation from the provider's raw counterparty
string and never updated, exactly mirroring `transactions.original_description` from migration 071.
`get_or_create_payee` then falls back to `lower(original_name)` when the name lookup misses, so a
renamed payee is still found.

The migration backfills it in one statement:

```sql
UPDATE payees SET original_name = name;
```

That single line makes **every existing payee on every install rename-safe immediately** — no
heuristics, no document parsing, no per-provider work. It also covers Enable Banking, SimpleFIN and
CSV import, none of which expose a document field.

### 2. The document as semantic identity

Add `payee_tax_id_kind` / `payee_tax_id_value` to `TransactionData` and a
`PluggyProvider._extract_tax_id` reading `paymentData.{receiver|payer}.documentNumber` — receiver on
debit, payer on credit, mirroring `_extract_payee`. Extract it *even when a name is present*, so
named company payees gain their CNPJ too. `get_or_create_payee` takes an optional `tax_id`, matches
on `(workspace_id, kind, value)` against the index that already exists, and on create also writes
the `PayeeTaxId` row and sets `Payee.type` from the kind (CPF ⇒ `person`, CNPJ ⇒ `company`) — the
rule `models/payee.py` documents but nothing implements.

### Resolution order

`document` → `lower(name)` → `lower(original_name)` → create. Each lookup is
`ORDER BY created_at, id LIMIT 1` so a duplicate resolves deterministically to the oldest row. On a
hit, fill in whatever is missing — attach the document if absent, set `original_name` if `NULL` — so
existing rows heal themselves without a separate data migration.

### Two things I'd deliberately avoid

- **No unique constraint** on `original_name` or on `(workspace_id, kind, value)`. Existing installs
  may already have two payees sharing either, and a constraint would turn that into a sync-time
  `UniqueViolationError` — the failure shape reported in #678. Oldest-wins is safer.
- **Never letting a bad document break a sync**: validate through the existing
  `normalise_and_validate` and skip on error.

I'm happy to implement this if the approach sounds right.

## Alternatives Considered

- **Parse the counterparty name out of the description.** The name often *is* there —
  `... - FAV.: JANE DOE|@...`. But it is bank- and format-specific (`FAV.:`/`REM.:`/`NOME:`,
  `|@`-delimited fields), and counting the affected rows in my data it would recover a name for only
  about half of them. More importantly it does nothing about identity: rename the payee and the
  duplicate still appears. A cosmetic improvement on top of the real fix, so I left it out.
- **Use the existing `payee_mapping` table as an alias table.** It has no column for a source
  identifier — only `target_id` — so it cannot key anything. It is merge bookkeeping, not aliasing.
- **A rule with `set_payee`.** Works as a stopgap, but it needs one rule per counterparty, it
  reassigns the transaction *after* the duplicate payee has already been created, and it doesn't
  scale to 64 of them.
- **Set `payee_source` to `description` on the connection.** A per-connection, all-or-nothing lever:
  it would replace the CPF with the entire raw descriptor
  (`TRANSF.REALIZADA PIX SICOOB - FAV.: ...|@...|@...`), which is worse.
- **Only the document, without `original_name`.** Simpler in one sense, but it leaves Enable
  Banking, SimpleFIN and CSV import unfixed, and it needs a CPF-detection backfill to help existing
  rows. `original_name` fixes all providers at once and backfills with one UPDATE.

## Additional Context

- Version: 0.14.4, self-hosted, Pluggy connection to Sicoob, PostgreSQL.
- Figures come from querying my own production database; all documents and names in this issue are
  fictitious replacements.
- Note that `handle_oauth_callback` (the initial, largest sync) never passes `payee_source` at all
  and always uses `auto`, and no path ever re-resolves `payee_id` for an already-synced transaction
  — so any fix applies to newly inserted rows, which is why the `original_name` backfill matters.
- Related, and deliberately **not** proposed here: on my credit-card account Pluggy sends
  `merchant: null` *and* `paymentData: null`, so `_extract_payee` returns `None` and **all 2,459
  card transactions have no payee at all**. The merchant descriptor exists only in `description`.
  I'd rather keep this issue to the identity question and open that separately if it's of interest.

jrvidotti:
> Validated on a preview deployment of my own instance, so this is no longer a draft and
> is ready for review. #702 asked for the approach to be agreed before the code, so I'm
> still happy to change direction if you'd rather take a different one.

## What

Stops a payee's identity being its display name, which is the one field a user is
invited to edit.

- `payees.original_name` (migration `076`) records what the provider called the
  counterparty when the row was created, mirroring the existing
  `transactions.original_description`. Set once, never updated. The migration seeds it
  from `name` for every existing row.
- `PluggyProvider._extract_tax_id` reads the counterparty's CPF/CNPJ from
  `paymentData.{receiver,payer}.documentNumber` and carries it on `TransactionData`.
  Both sync paths pass it through to the payee lookup.
- `get_or_create_payee` now resolves **document → `lower(name)` → `lower(original_name)`
  → create**, filling in whatever a matched row is missing, and stamping `type` from the
  document kind (CPF ⇒ `person`, CNPJ ⇒ `company`) — the rule `models/payee.py` already
  documents but nothing implemented.
- Migration `077` backfills the document for payees whose `original_name` *is* a
  document, so they reach the document-first lookup without waiting to be re-synced.

Enable Banking, SimpleFIN and CSV import report no document; they leave it `None` and
behave exactly as before, but still gain `original_name`.

## Why

Full detail in #702. In short: `get_or_create_payee` matched only on `lower(name)`, so
renaming a sync-created payee made the next sync fail to find it and insert a duplicate,
splitting one counterparty's history across two rows.

That is not a rare path. For a Pix to an *individual*, the institution returns
`receiver.name: null` while still sending the document, so `_extract_payee` falls back to
the document and the payee ends up literally **named after a CPF**. Companies are
unaffected, because their name does come through. In my workspace that was 64 of 190
payees, covering 253 transactions — every one of them impossible to rename permanently.

Meanwhile `payee_tax_ids` and its `(workspace_id, kind, value)` index have existed since
migration `070`, with a comment saying document matching is exactly what the table is
for — and nothing had ever written to it. My production database had 0 rows in it.

Two deliberate non-goals:

- **No unique constraint** on `original_name` or on `(workspace_id, kind, value)`.
  Existing installs may already have two payees sharing either, and a constraint would
  turn that into a sync-time `IntegrityError` — the failure shape of #678. Every lookup
  takes the oldest match instead.
- **A malformed document never fails a sync.** On this path the value is whatever an
  institution chose to send, so it is validated and dropped on error. The manual API
  keeps raising, because there a person typed it and wants to be told.

Credit-card payees are **out of scope** here. Pluggy sends `merchant: null` *and*
`paymentData: null` for card purchases, so `_extract_payee` returns `None` and those
transactions get no payee at all. Happy to open that separately.

## How to Test

Automated:

1. `cd backend && uv run pytest` — 3206 pass. The regression itself is
   `tests/test_payee_identity.py::test_rename_survives_the_next_sync`.
2. `uv run ruff check . && uv run ty check .`
3. `uv run alembic heads` — single head, `077`.

Migrations, against a throwaway Postgres:

4. `alembic upgrade head`, then `downgrade -1` twice and `upgrade head` again. `076`
   backfills `original_name` for every pre-existing row; `077` is skip-if-present, so
   re-running produces no duplicates.

By hand, against a real Pluggy connection:

5. Find a payee whose name is a CPF, rename it, and resync. Before this change a second
   payee appears with the original name; after it, the rename holds and the
   transactions stay on the renamed row.

Worth a reviewer's attention: `077` decides *when* a name counts as a document. Check
digits alone are not enough — eleven arbitrary digits pass them by luck about once in a
hundred — so the name must be nothing but a document. That is what stops an MEI trading
name (a CNPJ root followed by a person's name, e.g. `11.222.333 JOHN DOE SILVA`) from
being read as a CNPJ. `tests/test_migration_077_classify.py` pins that down.

## Related Issue

Closes #702.

## Checklist

- [ ] Frontend lints clean (`npm run lint`) — n/a, no frontend files changed
- [ ] Frontend builds (`npm run build`) — n/a, no frontend files changed
- [ ] Translations updated (if user-facing strings changed) — n/a, no user-facing strings
- [x] For a large or core change, I discussed it first (issue or Discord) so we could align before the code (see [CONTRIBUTING](../CONTRIBUTING.md#before-large-or-core-changes)) — #702

### DIFF
--- a/backend/alembic/versions/076_payee_original_name.py
+++ b/backend/alembic/versions/076_payee_original_name.py
@@ -1,0 +1,50 @@
+"""remember what the provider called a payee, so renaming one survives sync
+
+Revision ID: 076
+Revises: 075
+Create Date: 2026-08-25
+
+Sync identified a counterparty by its display name alone, which is the one
+field a person is invited to change. Correcting a payee the bank had named
+after a document therefore made the next sync fail to recognise it and
+insert a duplicate.
+
+The backfill is the point of this migration: setting `original_name` to the
+current `name` for every existing row makes every payee already in the
+database rename-safe immediately, with no heuristics and nothing to detect.
+"""
+
+from typing import Sequence, Union
+
+import sqlalchemy as sa
+from alembic import op
+
+revision: str = "076"
+down_revision: Union[str, None] = "075"
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+
+def upgrade() -> None:
+    op.add_column("payees", sa.Column("original_name", sa.String(255), nullable=True))
+
+    # Every row that exists now was named by whoever created it, and for the
+    # sync-created majority that is exactly the string the provider will send
+    # again on the next run. Seeding it is what lets an existing install
+    # rename a payee without waiting for it to be re-synced first.
+    op.execute("UPDATE payees SET original_name = name")
+
+    # Deliberately not unique. Two payees can legitimately end up sharing an
+    # original_name — after a merge, or alongside a hand-created row — and a
+    # constraint would turn that into a sync-time IntegrityError, which is
+    # the failure this whole change exists to stop causing (issue #678).
+    op.create_index(
+        "ix_payees_workspace_original_name",
+        "payees",
+        ["workspace_id", "original_name"],
+    )
+
+
+def downgrade() -> None:
+    op.drop_index("ix_payees_workspace_original_name", table_name="payees")
+    op.drop_column("payees", "original_name")
--- a/backend/alembic/versions/077_backfill_payee_tax_ids_from_original_name.py
+++ b/backend/alembic/versions/077_backfill_payee_tax_ids_from_original_name.py
@@ -1,0 +1,168 @@
+"""recover the document from payees the bank named after one
+
+Revision ID: 077
+Revises: 076
+Create Date: 2026-08-26
+
+Migration 076 gave every payee an `original_name` and made the lookup fall
+back to it, which is what lets a rename survive. This one goes further and
+recovers the *document* from those rows.
+
+The rows in question are the ones this whole change exists for: a Pix to an
+individual arrives with `receiver.name: null`, so the provider falls back to
+the CPF and the payee ends up literally named `529.982.247-25`. Reading it
+back out of `original_name` gets those payees onto the document-first
+lookup immediately, rather than only once the same counterparty happens to
+be paid again.
+
+Two things it is careful about:
+
+- A name is only treated as a document when it is *nothing but* a document.
+  "52.288.516 OZEIAS AMORIM VIEIRA" is an MEI trading name that opens with a
+  CNPJ root, not a CNPJ, and must not be read as one. Check digits alone
+  would not settle that — an eleven-digit string passes them by luck about
+  once in a hundred — so anything left over after the digits and their
+  punctuation disqualifies the row.
+- It never replaces a document a payee already carries. A value somebody
+  entered by hand is more considered than one inferred from a name here.
+
+CNPJ is handled alongside CPF. The reporter's data has no CNPJ-named payees,
+because institutions withhold an individual's name but not a company's, but
+the two cases are the same mechanism and excluding one would be arbitrary.
+
+Validation goes through `app.fiscal.registry.normalise_and_validate`, the
+same call `payee_service._apply_tax_ids` makes, so a document recovered here
+is stored byte-identically to one entered through the form. The usual reason
+to keep a migration free of application imports — that the code may change
+underneath it — is weak for this particular import: CPF and CNPJ check
+digits are fixed by law, and duplicating the arithmetic would create two
+copies free to drift apart.
+"""
+
+import uuid
+from typing import Optional, Sequence, Union
+
+import sqlalchemy as sa
+from alembic import op
+
+from app.fiscal.registry import TaxIdKind, normalise_and_validate
+
+revision: str = "077"
+down_revision: Union[str, None] = "076"
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+
+#: Punctuation a Brazilian document is allowed to wear. Anything else in the
+#: name means the name is not a bare document.
+_DOCUMENT_PUNCTUATION = set(". -/")
+
+#: The legal nature each document settles, mirroring what the service layer
+#: stamps when it creates a payee from a document.
+_TYPE_BY_KIND = {TaxIdKind.CPF.value: "person", TaxIdKind.CNPJ.value: "company"}
+
+
+def _classify(name: Optional[str]) -> Optional[tuple[str, str]]:
+    """(kind, normalised value) when this name is a document and nothing else."""
+    if not name:
+        return None
+    stripped = name.strip()
+    if not stripped:
+        return None
+    if any(c not in _DOCUMENT_PUNCTUATION and not c.isdigit() for c in stripped):
+        return None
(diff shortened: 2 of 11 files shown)


jrvidotti:
Marking this ready for review: I've now run it on a preview deployment of my own instance, against the real Pluggy-synced dataset the issue describes.

PR #710, opened 2026-08-26
fix(payees): make the list usable at the scale sync creates

### LINKED ISSUE #709: [Bug] Payees list: long names force horizontal scroll, no pagination, and the detail card displaces the table
## Description

Five problems with the Payees page, found using it on a workspace whose payees were created by bank sync — a few hundred rows, many named after a full legal entity.

1. **A long name forces the whole table to scroll sideways.** Every column except the name has a fixed width, and `TableCell` carries `whitespace-nowrap` while the table container is `overflow-x-auto`. One payee named `COMERCIAL DE ALIMENTOS E DISTRIBUIDORA ...` widens the table past the viewport, so the entire list scrolls horizontally to read any row.

2. **Every payee renders at once.** The list has no pagination and no virtualization — it maps the full result straight into the table. Several hundred rows in one scroll makes finding anything a scrolling exercise, and the page has no way to say "show me 20 at a time".

3. **Clicking a payee opens an info card above the table.** The summary panel renders _before_ the list, so opening it pushes the whole table down and creates a second scroll target. There's a comment in the code explaining the placement was chosen precisely because a panel _after_ a hundreds-row table would open below the fold — which is the same problem from the other side. Neither position works; the panel doesn't belong in the page flow at all. It also fights with pagination: the pager sits under a table that has just been shoved off screen.

4. **The info card has no actions.** It shows totals, top category and the last five transactions, but to edit or delete that payee you have to close it and find the row again to use the row's pencil/trash icons. The one place where you've just decided a payee is wrong is the one place you can't act on it.

5. **Adding to favorites gives no feedback until the network answers.** The star toggle is a plain PATCH plus a cache invalidation — no optimistic update, no pending state, and no error toast. On a slow connection the star stays unchanged for a second or more, which reads as a missed click, so people click again. If the request fails, nothing at all is shown.

## Steps to Reproduce

1. Open **Payees** on a workspace with a few hundred synced payees, at least one with a very long name.
2. The table scrolls horizontally; the long name is what widens it.
3. Scroll the list — every payee is on the page, with no pager and no rows-per-page control.
4. Click any row — the info card opens _above_ the table and pushes the list down.
5. In that card, look for a way to edit or delete the payee — there isn't one.
6. Throttle the network to Slow 3G and click a payee's star — nothing changes until the response lands.

## Expected Behavior

1. Long names ellipsize inside the name column, with the full name available on hover. The table never scrolls horizontally.
2. The list is paginated, with a rows-per-page control, matching the Transactions page.
3. The payee detail opens in a modal instead of in the page flow, so the list underneath doesn't move.
4. The detail view carries **Edit** and **Delete**, gated on write permission like the row actions already are.
5. The star flips immediately on click and rolls back with an error message if the server rejects it.

## Actual Behavior

As described above: horizontal scrolling, an unpaginated list, a detail panel that displaces the table, no actions in that panel, and a favorite toggle that appears to do nothing until the request completes.

## Environment

- Deployment: Docker
- Browser: Chrome
- OS: Linux

## Additional Context

Notes for whoever picks this up:

- `GET /api/payees` returns a bare `list[PayeeRead]` with no envelope and no total, and four other places call it unpaginated (`transactions.tsx`, `dashboard.tsx`, `rules.tsx`, `transaction-dialog.tsx`). Paginating **client-side** keeps the change to the one page; server-side pagination would mean a response-shape change rippling through all four callers plus the API tests.
- Pagination interacts with the existing bulk selection: shift-range select and the header "select all" checkbox should be scoped to the visible page, and the deselect-all branch currently calls `setSelectedIds(new Set())`, which would silently discard selections made on other pages.
- There is no `Sheet`, `Pagination`, `Tooltip` or `AlertDialog` primitive in `components/ui/`. The existing conventions are `Dialog` for detail/edit views, a hand-rolled Prev/Next pager (see `transactions.tsx` and `import-review-table.tsx`), and the native `title` attribute for revealing truncated text.
- Everything the fix needs already exists in `common.*` (`previous`, `next`, `rowsPerPage`, `edit`, `delete`), so the only new i18n string is an error message for a failed favorite toggle — which has to land in all 11 locale files, since `locales/i18n.test.ts` asserts key parity in both directions.

Happy to open the PR for this.

## Evidence

Image

jrvidotti:
## What

Five fixes to the Payees list, all frontend:

1. **Long names ellipsize instead of widening the table.** The name column takes the leftover width (`max-w-0 w-full`) and truncates, with the full string on the native `title`. The name also moved from a `<span>` to a `<p>` — `truncate` does nothing on an inline box, which is why the class that was already on the notes line never worked either. The arbitrary `max-w-[300px]` on that line is gone; the cell bounds it now.
2. **The list is paginated**, 20 per page, with a rows-per-page control persisted to `localStorage` and the page mirrored as `?page=`. Reuses the pager shape from `transactions.tsx` and the slicing from `import-review-table.tsx`.
3. **The detail card moved into a modal**, out of the page flow, and into `components/payee-detail-dialog.tsx` along with the two queries only it used.
4. **That modal now carries Edit and Delete**, gated on `canWrite` like the row actions. They stack on top of the detail rather than replacing it, so cancelling either returns to the payee instead of the bare table.
5. **The favorite star responds immediately** — an optimistic cache patch with rollback and an error toast, replacing the PATCH-then-invalidate that left the star unchanged until the refetch landed.

Selection also became page-scoped: shift-range and the header checkbox operate on the visible page, and the deselect-all branch no longer calls `setSelectedIds(new Set())`, which would have discarded selections made on other pages once paging existed.

## Why

A workspace whose payees came from bank sync has hundreds of rows, many named after a full legal entity. The page was built for a handful of hand-entered ones and doesn't hold up at that scale — one long name makes the whole table scroll sideways, every row renders at once, and the detail card shoves the list down when you click it. Full write-up in the issue.

Two decisions worth flagging for review:

- **Pagination is client-side.** `GET /api/payees` returns a bare `list[PayeeRead]` with no total, and four other callers (`transactions.tsx`, `dashboard.tsx`, `rules.tsx`, `transaction-dialog.tsx`) read it as a plain array. Paginating server-side would mean a response-shape change rippling through all four plus the API tests — a much larger diff than the problem warrants. Happy to do it properly in a follow-up if you'd rather the endpoint grew `page`/`limit`.
- **This is the first optimistic mutation in the codebase.** The `['payees']` prefix also matches `['payees', id, 'summary']`, whose data is an object, so the cache patch is filtered to array-shaped caches via a predicate. That has the side benefit of keeping the payee pickers on the other four pages in step.

No new dependencies and no new `components/ui/` primitives. One new i18n key (`payees.favoriteError`), added to all 11 locales; everything else reuses `common.previous`/`next`/`rowsPerPage`/`edit`/`delete`.

## How to Test

1. Open **Payees** on a workspace with more than 20 payees, at least one with a very long name.
2. The table no longer scrolls horizontally at any width; hover the name for the full string. Check 360px, 768px and 1440px — 360px is where the type column drops out and the layout redistributes.
3. Page through the list; change rows-per-page and reload — the size sticks. Paging puts `?page=2` in the URL; open `?q=<term>&page=2` in a new tab and both survive. Typing in the search box returns to page 1.
4. Shift-select a range on page 2 — only page-2 rows are selected. Untick the header checkbox there and page-1 selections stay. Bulk-delete a cross-page selection.
5. Click a row — the detail opens as a modal, the list underneath doesn't move. **Edit** stacks on top; Cancel returns to the detail, Save shows the updated name. **Delete** stacks; Confirm closes both.
6. Throttle to Slow 3G and click a star — it flips immediately. Go offline and click — it flips back and shows an error.
7. On a read-only workspace: no row actions, no modal footer.

## Related Issue

Closes #709

## Checklist

- [x] Backend tests pass (`pytest`) — n/a, no backend changes

`npm test` also passes (161 tests), including the `locales/i18n.test.ts` parity check across all 11 locale files.

https://claude.ai/code/session_01VUKX1tdZiMQ3qcpPhjDV8f

The Payees list now supports pagination, stable table layout, modal details, permission-gated actions, and optimistic favorite updates. Invalid saved page sizes fall back to supported values.

- Long payee names truncate with a hover title.
- Pagination and rows-per-page settings persist across visits.
- Payee details open in a modal.
- Edit and Delete appear when permitted.
- Favorite changes update immediately and roll back with an error message if they fail.
- Selection applies to the current page while preserving selections on other pages.


jrvidotti:
Before:
securo-payees

After:
securo-payees-2

tassionoronha:
Hey @jrvidotti . 
I did some improvs here to adjust with things we have merged, but the PR looks good overall. 
Thanks so much for this :)

PR #714, opened 2026-08-26
feat(payees): edit a counterparty from the transaction you noticed it on

### LINKED ISSUE #713: [Feature] Edit the payee from the transaction editor, and give the field a searchable picker
> Opening this before writing code, per CONTRIBUTING. I have a working plan and I'm happy to
> build it, but it extracts a component out of `pages/payees.tsx` that the payees page also
> uses, so I'd rather agree the shape first.

## Problem

When you edit a transaction, the Payee field is a bare native `<select>`
(`frontend/src/components/transaction-dialog.tsx:1047-1063`):

```tsx
<select value={payeeId} onChange={(e) => setPayeeId(e.target.value)}>
  <option value="">{t('payees.noPayee')}</option>
  {(payeesList ?? []).map((p) => (
    <option key={p.id} value={p.id}>{p.name}</option>
  ))}
</select>
```

No search, no per-item actions, no way to create one. Every other comparable field in the same
form has more than this — Category two rows above uses `CategorySelect`, a Popover + Command
combobox with accent-insensitive search.

That gap hurts most in the case the app produces by default. Sync names a payee after the raw
bank descriptor, so the list fills up with `PADARIA CENTRAL LTDA 0341`-shaped rows. Fixing one
means abandoning the transaction you were editing, navigating to `/payees`, finding the row in a
list that has its own scale problems (#709), renaming it, and coming back. The one moment you
*know* a payee is wrong is the moment you're looking at its transaction, and that's the one place
you can't act on it.

The `<select>` has a second, quieter problem: it renders every payee in the workspace as an
`<option>`, unsearchable. On a sync-heavy workspace that's an unusable control regardless of
naming.

### Two defects found while looking into this

**1. Renaming a payee leaves stale names on screen.** `createMutation` and `updateMutation` in
`frontend/src/pages/payees.tsx` invalidate only `['payees']`:

```ts
onSuccess: () => {
  queryClient.invalidateQueries({ queryKey: ['payees'] })
  ...
}
```

But transaction rows render `tx.payee_name`, which the server computes from `payee_entity.name`
(`backend/app/services/transaction_service.py:517,680,1070,1106`) and which lives in the
`['transactions']` cache. So after a rename, `/transactions`, `/dashboard` and any open drill-down
keep showing the old name until something else happens to refetch. `deleteMutation`,
`mergeMutation` and `bulkDeleteMutation` all call `invalidateFinancialQueries`; create and update
are the two that don't. This is visible today on `/payees`; it becomes glaring if the rename
happens from the transaction editor, where the stale name is right behind the dialog.

**2. A duplicate name reports as a bare "Error".** `update_payee` raises
`ValueError("A payee with this name already exists")`
(`backend/app/services/payee_service.py:185,227`), returned verbatim as a 400 `detail`. The
frontend's only error mapper, `taxIdErrorMessage` (`frontend/src/pages/payees.tsx:64-70`), matches
just the `invalid_tax_id:` prefix and returns `null` for anything else, so the toast falls back to
`t('common.error')` — "Error". Renaming a synced payee onto a name you already use is precisely
the collision this flow invites, and right now it gives no clue what went wrong.

## Proposed Solution

Three frontend changes. **No backend change is needed** — `PATCH /api/payees/{id}` already accepts
`name`.

**1. `PayeeSelect` — a combobox, modelled on `category-select.tsx`.** Popover + Command with
accent-insensitive search over `normalizeText`, a pencil affordance next to the field and on each
row, and a "Create …" row seeded with whatever was typed. Two deliberate divergences from
`CategorySelect`: controlled search with `shouldFilter={false}` and a capped result list, because
cmdk's own filter renders every item and hides non-matches — fine for ~50 categories, not for
thousands of sync-created payees; and a `source` badge on non-manual rows, since `Payee.source` is
already exposed by the API (and documented in `types/index.ts` as being there for exactly this)
but ignored by every picker today.

**2. Extract `PayeeFormDialog` from `pages/payees.tsx`.** The create/edit form only exists inline
on that page (~170 lines of JSX, 8 pieces of form state, the tax-id rows). Extracting it means the
transaction editor gets the real form — including type, contact fields and documents — rather than
a third payee UI to keep in sync. `createMutation`/`updateMutation` move into the component;
`deleteMutation` stays on the page, since its `onSuccess` touches `selectedIds`, `summaryPayee` and
the confirm dialog, all page concerns. The page keeps driving it with the `dialogOpen` /
`editingPayee` state it already has, so `PayeeDetailDialog`'s `onEdit` needs no change.

**3. Fix the two defects above** — `invalidateFinancialQueries` on create/update, and a
duplicate-name branch in the error mapper.

Mounting the dialog inside the transaction dialog follows a pattern already proven in this
codebase: `AddTransactionToRuleDialog` is a Dialog-inside-Dialog containing its own
Popover + Command (`transaction-dialog.tsx:1349-1361`), and `/payees` already stacks the edit
dialog on top of `PayeeDetailDialog`.

## Alternatives Considered

- **Just a pencil next to the existing `<select>`.** Smallest possible change and it does solve
  the stated problem. Rejected because it leaves a searchless `<select>` over a list the sync
  fills with hundreds of entries — the control would still be the thing standing between you and
  the payee you're trying to fix.
- **A rename-only mini dialog** (name field, Save). Cheap, but it's a third payee UI to maintain
  alongside the page form and `PayeeDetailDialog`, and it can't answer the obvious next question
  ("this one's a company, and I have its document").
- **Reusing `PayeeDetailDialog`.** It's a read surface — summary figures and recent transactions —
  and delegates editing back to the page. Wrong shape for this.

## Additional Context

This does **not** fix #702. Renaming a sync-created payee still lets the next sync re-create the
old one, because `get_or_create_payee` (`backend/app/services/payee_service.py:68-113`) matches on
`func.lower(Payee.name)` alone. This change makes renaming easy to *reach*; #702 / #708 makes it
*stick*. They're complementary, and if #708 lands first this feature inherits the fix for free.

Related: #709 (payees list at sync scale) — same underlying pressure, different surface.

Happy to open a PR if the shape sounds right. Also happy to split the two defect fixes into their
own PR first, since they stand on their own and are a handful of lines each.

jrvidotti:
> **Stacks on #710.** This branch is cut from `feature/payees-list-ux`, so until that
> merges the diff here carries its commit too. The commit to review is `e397213`;
> everything in `a3343cb` belongs to #710.

## What

Makes the payee editable from the transaction editor, and gives the field a picker that works at
the scale sync creates.

- **`PayeeSelect`** replaces the bare `<select>` in the transaction dialog. Popover + Command,
  modelled on `category-select.tsx`, with accent-insensitive search, a pencil on each row and one
  next to the field, and a "Create …" row seeded with whatever was typed.
- **`PayeeFormDialog`**, extracted from `pages/payees.tsx`, so the payees page and the transaction
  editor share one form instead of growing a third payee UI. The page drops ~290 lines.
- **Two defect fixes** the new flow would otherwise make glaring (below).

No backend change: `PATCH /api/payees/{id}` already accepts `name`.

## Why

Closes #713, which has the long version.

The short one: sync names a payee after the raw bank descriptor, so the list fills with rows like
`PADARIA CENTRAL LTDA 0341`. The moment you know one is wrong is while you are looking at its
transaction — and that was the one place that could not fix it. The field was also unsearchable,
rendering every payee in the workspace as an `<option>`, two rows below a Category field that has
had a searchable combobox all along.

### The two defects

**Renaming left the old name on screen.** `createMutation`/`updateMutation` in `pages/payees.tsx`
invalidated only `['payees']`, but transaction rows render `payee_name`, which the server computes
from the payee entity and which lives in the `['transactions']` cache. `deleteMutation`,
`mergeMutation` and `bulkDeleteMutation` all call `invalidateFinancialQueries`; these two did not.
Visible today on `/payees`; glaring once the rename happens from the transaction editor, where the
stale name sits directly behind the dialog you just saved.

**A duplicate name reported as "Error".** `update_payee` raises
`ValueError("A payee with this name already exists")` (`payee_service.py:185,227`), returned
verbatim as the 400 detail. The only error mapper matched the `invalid_tax_id:` prefix and returned
`null` for everything else, so the toast fell back to `t('common.error')`. Renaming a synced payee
onto a name you already use is exactly the collision this flow invites.

Matching the English sentinel is brittle to a backend reword — I left a comment pointing at the
raise sites. A machine-readable code (`duplicate_payee_name`) would be the real fix; happy to do it
in a follow-up if you'd rather have it in the API.

## Notes on three decisions

**The payee form is deliberately not a `<form>`, and every button in it is `type="button"`.**
`ui/button.tsx` sets no default type, so `<Button>` is a submit button. The dialog is portalled but
still a React child of the transaction dialog's own `<form>`, and submit events cross React
portals — so a stray submit would save the transaction behind the user's back. This is the hazard
`AddTransactionToRuleDialog` already guards at `transaction-dialog.tsx:1447-1450`. Enter in the
name field is wired explicitly instead.

**`PayeeSelect` filters itself rather than letting cmdk do it.** cmdk's `filter` renders every item
and hides the misses, which is fine for fifty categories and not for the thousands of rows sync
creates, so the search is controlled, `shouldFilter={false}`, and the list is capped at 200 with
favourites lifted. Relevance decides the order and a star only breaks ties.

**`deleteMutation` stayed on the page.** Its success path reaches into `selectedIds`,
`summaryPayee` and the confirm dialog — all page concerns — so the component takes an
`onRequestDelete` callback and the transaction editor simply omits it. Deleting the payee you are
half-way through attaching to a transaction has no confirmation surface there.

## How to Test

Mechanically, on this branch:

```bash
cd frontend && npm ci
npm run lint    # 0 errors
npm run build   # tsc -b + vite build
npm test        # 196 passing, 35 of them new
```

The 35 new tests cover the extracted pure logic — `lib/payee-form-utils.ts`,
`lib/payee-errors.ts`, `lib/payee-select-utils.ts`. The repo has no component testing library, so
the interactive parts need hands:

1. `/transactions` → click a row with a payee. The field is now a combobox; open it and type a
   partial name with the accents dropped (`sao` for `São`) — it should match.
2. Click the pencil **on a row in the list**. The popover should close and the payee dialog open,
   and the row should **not** get selected. (cmdk's `Item` fires `onSelect` from its own click;
   guarded with `stopPropagation` plus `preventDefault` on pointerdown.)
3. Rename, Save. The transaction dialog stays open and shows the new name. Close it: the row's
   Payee cell shows the new name with no reload — this is the invalidation fix. Check `/dashboard`
   too.
4. Type a name that does not exist → "Create …" → the dialog opens in create mode with the name
   pre-filled and **no Delete button**. Save: the new payee is selected on the transaction.
5. Rename a payee onto another's exact name → the toast should name the collision, not say "Error".
6. With the payee dialog open on top of the transaction dialog, press **Esc**: only the top one
   should close, with the transaction's unsaved edits intact.
7. `/payees` regression sweep: add, edit (row pencil and from inside the detail dialog), delete
   from the footer and from the row, merge, bulk delete, star, filters, search, pagination.

**In the interest of not overstating it:** I have run the three automated gates and confirmed the
new modules transform cleanly through Vite, but I have not yet clicked through steps 1-7 in a
browser. Steps 2 and 6 are the ones I would check first.

## Related Issue

Closes #713. Stacks on #710. Complementary to #708 — this makes a rename easy to reach, that makes
it survive the next sync; neither needs the other.

## Checklist

- [x] Backend tests pass (`pytest`) — n/a, the diff touches no backend file
- [x] Translations updated — 9 new keys across all 11 locales
- [x] For a large or core change, I discussed it first — #713


jrvidotti:
securo-payees-edit

- Create and edit inline
- Searchable combobox

PR #735, opened 2026-08-28
fix(sync): let rules override provider categories

### LINKED ISSUE #730: Categorization rules can never override the category a sync provider already assigned
## Description

A rule with a `set_category` action is supposed to categorize a transaction when its conditions match, for example a description containing a specific merchant name. In practice, that action is a no-op on almost every synced transaction, because the provider-supplied category (from Pluggy's category mapping) is already written to the transaction before rules run, and the rule engine treats any pre-existing category as final. A rule can only ever fill in a category the provider left blank. It can never win over one the provider already guessed, no matter how specific the rule's condition is.

## Steps to Reproduce

1. Create a rule with a condition like "description contains X" and an action `set_category` pointing at category A.
2. Sync a bank connection where an incoming transaction's description contains X, and whose provider category (Pluggy `pluggy_category`) maps to a different category B in this workspace.
3. Inspect the resulting transaction's `category_id`.

## Expected Behavior

The rule matches on the transaction's own description, so it is more specific than the provider's generic category guess. It should win, and the transaction should land in category A.

## Actual Behavior

The transaction lands in category B (the provider's mapping), and the rule's `set_category` action is silently skipped.

Reproduced concretely on my own workspace: a recurring "PORTO CONSORCIO" real estate consortium payment, and a recurring "BRADESCO ADMINISTRADORA..." vehicle consortium payment, both had rules created for them pointing at a dedicated "Patrimonio / Consorcios" category. Neither rule ever categorized the transaction; both kept landing in whatever generic category Pluggy's own mapping produced (a transfer-like category for one, "Transporte" for the other), month after month, because that category was already set by the time the rule ran.

## Root Cause (traced in code)

In `connection_service.py`, both in the initial-import branch and the incremental-sync branch, `category_id` comes from `_match_pluggy_category(...)` and is passed straight into the new `Transaction(..., category_id=category_id, ...)` before the transaction is ever evaluated against rules. Only after that does `apply_rules_to_transaction(session, user_id, transaction)` run.

In `rule_engine.py`, `apply_rule_actions()` guards the category action like this:

```python
if op == "set_category" and not category_already_set:
    ...
    tx.category_id = category_id
    category_already_set = True
```

`category_already_set` is derived from `transaction.category_id is not None` at the moment rules start evaluating. Since the provider category was already assigned moments earlier, this is true for the large majority of synced transactions, so the `set_category` branch is dead code in the common case. This also silently affects the recurring-transaction placeholder-merge path (`placeholder.category_id = preview.category_id` only when the placeholder itself had no category yet), which inherits the same problem because the `preview` transaction used to evaluate rules is built from a transaction that already carries the provider category.

## Proposed Fix

Give the rule engine first claim on the category, and use the provider's mapping only as a fallback for whatever no rule covers, instead of the other way around:

1. Build the new `Transaction` with `category_id=None` in both sync branches.
2. Run `apply_rules_to_transaction` (or the rule preview, in the placeholder branch) so a matching rule can claim the category.
3. Only then, if `category_id` is still `None`, assign the provider-derived category as a fallback.

This preserves today's behavior whenever no rule matches (the provider category still wins by default), and only changes behavior when a rule's `set_category` action actually fires, which is exactly the case that was silently broken.

## Testing

Added two regression tests in `test_connection_service.py`:

- `test_sync_rule_category_wins_over_provider_category`, covering the initial-import branch.
- `test_incremental_sync_rule_category_wins_over_provider_category`, covering the incremental-sync branch (account already exists from a prior sync).

Both were confirmed to fail against the pre-fix code (verified by temporarily reverting the fix and re-running) and to pass with it. The rest of the existing suite most relevant to this path (`test_connection_service.py`, `test_connection_service_coverage.py`, `test_recurring_match_sync.py`, `test_recurring_match_import.py`, `test_rule_service.py`, `test_rule_engine.py`, 244 tests before this change) stays green, 246 passing total with the two new tests included.

## Notes

Found and fixed on a fork while cleaning up categorization rules for household budgeting. The fix exists on a local branch of the fork; it has not been committed or opened as a PR against this repository yet, this issue is filed first so the reasoning and test evidence are on record before that happens.

vhsantos26:
## Summary

Rules that match a transaction description now take precedence over the generic category supplied by a provider. The provider category remains the fallback when no rule sets one.

The fix applies to both initial imports and incremental synchronization. It also preserves the provider fallback while an incoming transaction is merged with a recurring placeholder.

## Tests

- `uv run pytest tests/test_connection_service.py -q`
- New regression coverage for initial import and incremental sync

Fixes #730.

Categorization rules now run before provider categories during initial and incremental sync. Provider categories remain the fallback when no rule assigns a category.

- Matching rules can override provider categories.
- Recurring transactions keep rule-selected categories.

Risk: sync provider.


tassionoronha:
Thanks here again @vhsantos26 LGTM!

PR #739, opened 2026-08-29
test: pin dashboard forecast date

ADolkun:
## Summary
- freeze the dashboard and opening-balance clocks inside the forecast split test
- derive transaction dates and requested month from the same mid-month date
- keep production behavior unchanged

## Root cause
The test used today plus three days but queried only the current month. On the last three UTC days of a month, the future transaction moves into the next month and is correctly excluded, making the hard-coded projected balance fail.

## Validation
- UTC regression: passed
- Pacific/Kiritimati regression: passed
- Dashboard API tests: 18 passed, 1 skipped
- Ruff and diff check: passed
- Independent Standards and Spec reviews: passed

The forecast split test now uses one fixed mid-month date for transaction dates and the requested month. This prevents month-boundary failures without changing production behavior.

- Production behavior remains unchanged.

Risk: none.

### DIFF
--- a/backend/tests/test_dashboard_api.py
+++ b/backend/tests/test_dashboard_api.py
@@ -362,10 +362,17 @@ async def test_total_balance_computed_from_transactions(client, auth_headers):
 
 @pytest.mark.asyncio
 async def test_pending_and_future_rows_are_current_vs_projected(
-    client, auth_headers
+    client, auth_headers, monkeypatch
 ):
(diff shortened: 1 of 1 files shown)


tassionoronha:
Thanks for this @ADolkun, but in the end I considered that the #749 had a better response for the same issue.
Closing here since is already resolved. 👍🏼

PR #757, opened 2026-08-30
fix(budgets): offset positive refund transactions in category budget calculation

### LINKED ISSUE #728: [Bug] Category budget calculation ignores positive/refund transactions
## Description

When a category includes a refunded item (marked as a positive value/income transaction), the budget calculation ignores this positive amount. As a result, the net budget expense for the category displays an incorrectly high expense total rather than subtracting the refund.

## Steps to Reproduce

1. Add a refund transaction (positive value) under a specific category (e.g., Medical)
2. Navigate to the Dashboard.
3. View the Budget Categories section for that category.
4. Observe that the total expense ignores the positive refund transaction.

## Expected Behavior

Positive refund transactions within a category should offset expenses when calculating the category's net budget total (i.e., $\text{Net Expense} = \text{Total Expenses} - \text{Refunds}$).

## Actual Behavior

Only negative (expense) transactions are summed up. The positive (refund) transaction is omitted from the category total, displaying an overstated net expenditure.

Example: In the Medical category, there are 3 total transactions, 1 of which is a positive refund. 

Image

On The Dashboard budget total sums only the 2 negative transactions and ignores the positive one.

Image

## Environment

- Deployment: Docker
- Browser: Chrome 120
- OS: Windows 11

## Screenshots

Attached above

## Additional Context

Any other context about the problem.

vjymisal0:
Fixes #728

## Problem

When calculating category budget spending in get_budget_vs_actual (udget_service.py), the query filtered only Transaction.type == debit. As a result, positive refund (credit) transactions within a category were ignored, causing the category's actual net expenditure to appear overstated instead of being offset by refunds.

## Fix

1. Updated get_budget_vs_actual to compute net category expenditure (Debits − Credits) using a SQL case expression:
   `python
   func.sum(
       case(
           (Transaction.type == debit, _primary_amount_expr()),
           (Transaction.type == credit, -_primary_amount_expr()),
           else_=0,
       )
   )
   `
   for both current-month and previous-month actuals.
2. Added unit test 	est_budget_vs_actual_offsets_refunds in ackend/tests/test_budget_service.py verifying that credit/refund transactions reduce the net expense total for a budgeted category.

`get_budget_vs_actual` now subtracts refunds from expenses for current and previous months. This prevents category spending from appearing higher than it is.

- Refunds reduce the expense total for their category.
- Categories with zero or negative net spending are excluded.

Risk: money math.


vjymisal0:
These two CI failures look unrelated to this PR.

This PR only changes `get_budget_vs_actual` in `budget_service.py`, which is reached solely through `app/api/budgets.py`. Neither failing test exercises that path:

- `test_dashboard_api.py::test_pending_and_future_rows_are_current_vs_projected`
- `test_report_service.py::test_cash_flow_recurring_credit_increases_balance`

Both look month-boundary sensitive, and this run landed on Aug 31:

- The report test fails with `ValueError: day 31 must be in range 1..30 for month 9 in year 2026` — a day-of-month rollover when adding a month to Aug 31.
- The dashboard test computes `today = date.today()` (Aug 31) and `future_date = today + timedelta(days=3)` (Sep 3), so the pending row and the queried forecast month fall on opposite sides of the boundary.

The budget-service suite itself passes (`2 failed, 3175 passed`). Happy to open a separate PR hardening those two date rollovers if that would be useful — just let me know.

vjymisal0:
Added the optional regression coverage requested by review in 4cc4c76: previous-month spending now includes a 30 credit refund and asserts the net 45 amount. The two reported CI failures remain unrelated date-boundary failures in dashboard/report tests. Local pytest is blocked by the environment's missing Flask dependency.

vjymisal0:
Addressed the remaining review nit in the latest commit: the previous-month regression now asserts exactly one category result before checking the net amount, so missing comparison output cannot silently pass. `git diff --check` and Python compilation pass; pytest is blocked locally by the environment missing Flask.

vjymisal0:
Good catch. I added a previous-month debit and credit and now assert the category’s net `prev_month_amount` is 55 while retaining the April actual assertion. The targeted Ruff check passes; the test update is in commit `2d7b0da`.

PR #761, opened 2026-08-30
fix(enable-banking): prevent duplicate transactions from pagination loops

### LINKED ISSUE #717: [Bug] Enable Banking pagination loop imports the same transactions 50 times
## Description

When syncing an account through Enable Banking, Securo imports the same transactions exactly 50 times during the initial account setup.

The Enable Banking API itself returns the transactions correctly on the first page. However, for the affected account, requesting the next page using the returned `continuation_key` returns the same page again together with the exact same `continuation_key`.

Securo does not appear to detect this pagination loop and continues requesting the same page until its `TRANSACTION_PAGE_LIMIT = 50` is reached. As a result, every transaction on that page is imported 50 times.

The duplicated transactions in the Securo database have different internal `id` values but the same `account_id` and `external_id`.

## Steps to Reproduce

1. Connect an affected account using the Enable Banking provider.
2. Let Securo perform the initial transaction sync.
3. Open the affected account in Securo.
4. Observe that the same transactions have been imported exactly 50 times.
5. Query the duplicated transactions in the database.
6. Observe that duplicates have different internal `id` values but identical `external_id` values.

The underlying Enable Banking pagination can also be reproduced directly.

The first request returns, for example, 15 transactions and a `continuation_key`:

```text
PAGE 1
Transactions: 15
Current key sent: None
New key received: A
```

Requesting the next page with `continuation_key=A` returns the exact same 15 transactions and the exact same continuation key:

```text
PAGE 2
Transactions: 15
Current key sent: A
New key received: A
```

Therefore, Enable Banking appears to have a pagination issue for this connection.

Securo currently has:

```python
TRANSACTION_PAGE_LIMIT = 50
```

and continues fetching pages until either no continuation key is returned or the 50-page limit is reached. A repeated continuation key does not appear to terminate pagination.

## Expected Behavior

Securo should detect a repeated `continuation_key` and stop pagination instead of requesting the same page repeatedly.

For example, if a continuation key has already been used or seen:

```python
if continuation_key in seen_continuation_keys:
    logger.warning("Enable Banking pagination loop detected")
    break
```

Securo should also avoid inserting multiple transactions with the same `account_id` and `external_id`.

The initial sync should therefore import each transaction only once, even if the upstream provider returns a malformed pagination sequence.

## Actual Behavior

Securo requests the same Enable Banking page repeatedly until the 50-page safety limit is reached.

Consequently, each transaction from the affected page is imported exactly 50 times.

For example, two transactions appear repeatedly in the UI:

```text
04/08/2026   Transaction   -500.00 €
04/08/2026   Transaction   -150.00 €
04/08/2026   Transaction   -500.00 €
04/08/2026   Transaction   -150.00 €
...
```

The corresponding database rows have:

```text
different id
same account_id
same external_id
```

This occurs already during the initial connection setup and does not require multiple manual syncs.

## Environment

- Deployment: Docker / Portainer
- Provider: Enable Banking
- Institution: Not disclosed
- Database: PostgreSQL
- Browser: Safari
- Client OS: macOS

## Additional Context

The upstream Enable Banking behavior appears to be incorrect because it returns the same page and the same `continuation_key` when that continuation key is supplied.

However, Securo should handle this condition defensively. The existing `TRANSACTION_PAGE_LIMIT = 50` prevents an infinite loop, but in this case it causes the same transactions to be accumulated 50 times before stopping.

There also appears to be no database-level uniqueness constraint preventing multiple synced transactions with the same `(account_id, external_id)` from being stored.

The transaction data returned through Enable Banking for the affected account is particularly sparse. For example:

```json
{
  "entry_reference": null,
  "transaction_amount": {
    "currency": "EUR",
    "amount": "-500.000000"
  },
  "creditor": null,
  "debtor": null,
  "credit_debit_indicator": "DBIT",
  "status": "BOOK",
  "booking_date": "2026-08-04",
  "value_date": null,
  "transaction_id": null,
  "remittance_information": null
}
```

Because `entry_reference` is unavailable, Securo generates a deterministic fallback `external_id`. The duplicate rows in the database consequently have the same generated `external_id`, which makes the duplication directly detectable.

Two safeguards would therefore be useful:

1. Stop Enable Banking pagination when a `continuation_key` repeats.
2. Prevent duplicate synced transactions with the same `(account_id, external_id)` from being inserted.

jfrcom:
## What
- Detect repeated Enable Banking `continuation_key` values and stop pagination with a warning.
- Deduplicate transactions by `external_id` across fetched pages.
- Add a second safeguard during initial account setup to prevent duplicate `(account_id, external_id)` rows.
- Add regression tests for repeated pagination cursors and duplicate initial transactions.
## Why
Enable Banking can return the same transaction page together with the same `continuation_key`. Securo previously continued requesting that page until reaching the 50-page safety limit, causing every transaction on the page to be imported up to 50 times.
The new safeguards stop the pagination loop and ensure repeated provider data cannot create duplicate transactions during initial account setup.
## How to Test
1. Deploy this branch in a test or staging environment with Enable Banking configured.
2. Connect an Enable Banking account affected by the repeated `continuation_key` behavior.
3. Allow the initial account and transaction sync to complete.
4. Confirm that each transaction appears only once in Securo instead of 50 times.
5. Check the backend logs and confirm that the repeated cursor produces an `Enable Banking pagination loop detected` warning.
6. Verify that the database contains no duplicate external transaction IDs for the connected account:
   ```sql
   SELECT account_id, external_id, COUNT(*)
   FROM transactions
   WHERE external_id IS NOT NULL
   GROUP BY account_id, external_id
   HAVING COUNT(*) > 1;
   ```
7. Trigger another account sync and confirm that no additional duplicate transactions are inserted.
8. Connect or sync an Enable Banking account with normal pagination and confirm that all transaction pages are still imported.
Automated verification: the two new regression tests and both affected backend test files pass (124 tests total). The complete backend suite was interrupted before completion, so the backend-test checklist item remains unchecked.
## Related Issue
Closes #717
## Checklist

Enable Banking pagination now stops when a continuation key repeats and removes duplicate transactions. Initial account setup imports each transaction only once.

- Repeated provider pages no longer create duplicate transactions.
- Initial sync skips transactions with the same account and external ID.

Risk: sync provider and database writes. No database migration, auth, or money math.


tassionoronha:
Thanks for this one @jfrcom ! LGTM :)

PR #784, opened 2026-08-31
fix(security): prevent path traversal prefix collision and add tomllib fallback

vjymisal0:
### Summary
1. **Path Traversal Containment Check**: In LocalStorageProvider._full_path, string prefix check str(full).startswith(str(base)) was vulnerable to prefix collisions with sibling directories (e.g. /data/attachments_evil). Replaced with canonical ull.relative_to(base). Added unit test 	est_path_traversal_prefix_collision_blocked.
2. **Python Compatibility**: Added fallback import for 	omllib / 	omli in pp/fiscal/registry.py to prevent startup import crashes on Python <3.11.

### Testing
- pytest backend/tests/test_local_storage.py (13/13 passed)
- Full backend suite: 3,184 passed, 49 skipped, 0 failures.

Storage path validation now rejects sibling paths with matching prefixes. The fiscal registry also supports Python versions before 3.11.

- Rejects uploads that target paths outside local storage.
- Supports `tomli` when `tomllib` is unavailable.


vjymisal0:
Removed the unreachable tomli fallback in e978c71. The backend explicitly requires Python >=3.11, where stdlib tomllib is guaranteed, so the fallback was inconsistent with package metadata and introduced an undeclared import. Python compilation, Ruff, and diff checks pass.

tassionoronha:
Thanks for that and welcome here @vjymisal0 ! 
LGTM!

PR #798, opened 2026-09-01
fix(frontend): honour the number-format setting in transaction amount…

### LINKED ISSUE #767: [Bug] , is not recognised as decimal point
## Description

For some languages we use a comma as the decimal point, and I would like to use a comma instead of a full stop. As I have selected “dot for thousands, comma for decimal” in the settings, I would presume it would use that setting.

## Steps to Reproduce

1. Use a comma instead of a full stop.

## Expected Behaviour

It should honour the setting from the settings.

## Actual Behaviour

It forces the use of a full stop.

## Environment

- Deployment: LXC Proxmox
- Browser: Firefox 154.0.1
- OS: Windows 11

## Screenshots

Image

Image

## Additional Context

Any other relevant context about the problem.

shaynhornik:
… inputs

Amount fields were type=number, so the browser rejected comma decimals outright, and parseFloat read locale-formatted input wrong (1.234,56 became 1.234). Parse and render the transaction dialog's money fields through the display locale instead, including seeds when editing, so stored values round-trip under every separator convention.

Closes #767

## What

Transaction dialog amount fields (amount, converted amount, FX rate) now
honour the admin number-format setting. The inputs change from
`type="number"` to `type="text" inputMode="decimal"`, and parsing/rendering
goes through two new locale-aware helpers in `lib/format.ts`
(`parseAmountInput` / `formatAmountInput`), driven by the same
`useDisplayLocale()` that already controls how amounts are displayed.
Values seeded into the fields when editing an existing transaction are
formatted the same way, so stored amounts round-trip correctly under every
separator convention. Invalid input on save is blocked with the standard
error toast instead of being persisted.

Scope note: this covers the transaction dialog, the surface reported in the
issue. The other money forms (transfer, budgets, splits, accounts) still use
`parseFloat` and can migrate to the same helpers in follow-up PRs.

## Why

With "dot for thousands, comma for decimal" selected, typing `1,50` in the
amount field was rejected by the browser (`type="number"` only accepts a
dot on an English-locale page), and `parseFloat` reads locale-formatted
input wrong - `1.234,56` becomes `1.234`. A finance app silently storing
the wrong number is the worst version of this bug. Display already honours
the setting; input now does too.

## How to Test

1. Settings → set number format to "dot for thousands, comma for decimal".
2. Open Add Transaction and type `1.234,56` in Amount - it is accepted.
3. Save, then check the transaction list shows 1.234,56 (stored 1234.56).
4. Reopen the transaction - the field shows `1234,56`, and saving again
   does not change the value.
5. Switch the format back to "comma for thousands, dot for decimal" and
   verify `1,234.56` still parses as before.
6. `cd frontend && npx vitest run src/lib/format.tests.ts' - 12 new tests
   cover all three separator conventions, NBSP grouping, signs, garbage
   rejection, and round-tripping.

## Related Issue

Closes #767

## Checklist

- [x] Backend tests pass (`pytest`)  - not affected, frontend only change
- [x] Frontend lints clean (`npm run lint`) — 0 errors on changed files
- [x] Frontend builds (`npm run build')
- [x] Translations updated (if user-facing strings changed) — no new strings; the save guard reuses the existing 'common.error' key
- [ ] For a large or core change, I discussed it first — not applicable,
      scoped bug fix on an open issue

Transaction dialog amount fields now honor the selected number format. Users can enter and edit localized amounts, including comma decimals.

- Amount, converted amount, and FX rate fields accept locale-specific separators.
- Existing values display using the selected locale.
- Invalid conversion values are rejected when saved.
- Zero is accepted where valid. FX divisors must remain non-zero.
- Frontend tests cover localized parsing and formatting.

Risk: money math.

### DIFF
--- a/frontend/src/components/transaction-dialog.tsx
+++ b/frontend/src/components/transaction-dialog.tsx
@@ -2,7 +2,7 @@ import { useState, useEffect, useCallback, useMemo, useRef } from 'react'
 import { getAccountLabel, getAccountName, sortAccountsByDisplayName } from '@/lib/account-utils'
 import { useTranslation } from 'react-i18next'
 import { useDateLocale, useDisplayLocale } from '@/hooks/use-display-locale'
-import { formatCurrency } from '@/lib/format'
+import { formatAmountInput, formatCurrency, parseAmountInput } from '@/lib/format'
 import { useMutation, useQuery, useQueryClient } from '@tanstack/react-query'
 import { useAuth } from '@/contexts/auth-context'
 import { currencies as currenciesApi, transactions as transactionsApi, settings as settingsApi, payees as payeesApi, rules as rulesApi, categories as categoriesApi, categoryGroups as categoryGroupsApi } from '@/lib/api'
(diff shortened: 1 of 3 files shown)


tassionoronha:
Thanks for that @shaynhornik ! LGTM :)

PR #806, opened 2026-09-01
fix(assets): show profit for manual and sold holdings

arturschaefer:
## What

- show the profit amount in each collapsed holding row
- calculate return for manually valued assets from `purchase_price`, even when they have no ticker or ledger
- show cumulative `realized_gain` for fully sold ledger holdings
- derive profit from `sell_price - purchase_price` for manually sold assets

## Why

The collapsed row only calculated return when both ledger-only fields (`average_price` and `total_invested`) existed. Manually valued assets already expose `gain_loss` and `purchase_price`, but the UI ignored them. Fully sold ledger holdings similarly expose cumulative `realized_gain`, but rendered a dash.

This is a display fix; imports and backend calculations already provide the required values.

## How to test

- Add a manually valued asset with a purchase price and a newer value; its collapsed row shows profit and percentage.
- Fully sell a market-priced holding; its collapsed sold row shows cumulative realized profit.
- Sell a manually valued asset with purchase and sale prices; its collapsed sold row shows profit and percentage.

## Browser verification

Checked the patched holdings screen in the browser at desktop width with synthetic BRL data:

- a manually valued asset without a ticker rendered `+R$ 446,45` and `+12.1%` in the collapsed Return cell
- a fully sold ledger asset rendered its cumulative realized profit, `+R$ 249,48`, in the collapsed Return cell
- both values remained aligned with the existing table columns, badges, balances, and row actions

Only synthetic data was used for this check.

## Checks

- `vitest run src/lib/asset-profit.test.ts` (4 passed)
- `vitest run src/pages/pages.contract.test.ts` (30 passed)
- `npm run build`
- `npm run lint` (no errors; existing warnings only)

The full frontend suite reached 655/656 tests; one unrelated page-module import timed out at the five-second limit under parallel load, then passed in isolation.

Updates asset holding rows to show profit amounts and percentages for manually valued, manually sold, and fully sold ledger holdings.

- Shows profit in collapsed holding rows.
- Calculates profit from purchase and sale prices for manual holdings.
- Shows realized profit for fully sold ledger holdings.
- Supports assets without a ticker or ledger.

Risk: money math.


tassionoronha:
Thanks for spotting this one! Before merge: a manual asset with no valuation gets gain_loss 0, not null, so it renders +0.0% instead of —. Gating on value_count > 0 should cover it.

arturschaefer:
> Thanks for spotting this one! Before merge: a manual asset with no valuation gets gain_loss 0, not null, so it renders +0.0% instead of —. Gating on value_count > 0 should cover it.

Thanks for pointing it out. That's good catch! Let me fix it 🚧

arturschaefer:
Fixed in 59b8bc6. Unsold profit now requires `value_count > 0`, while sold assets still use realized or sale-price profit. Added a regression test for a purchase-price-only manual asset (`gain_loss: 0`, `value_count: 0`) rendering no profit.

tassionoronha:
LGTM now! Thanks for that @arturschaefer !

PR #807, opened 2026-09-01
fix(frontend): derive display currency from active workspace

Santhosh17:
## Problem
Balances and totals across the app (accounts list, dashboard, transactions, budgets, goals, assets, reports, invoices, etc.) were formatted using the signed-in user's global `currency_display` preference, not the active workspace's `default_currency`.

Because `currency_display` is stored on the user (not the workspace), switching between two workspaces with different currencies (e.g. EUR vs. INR) didn't change what was displayed — the old currency stuck around until the user manually edited a workspace's currency (which happens to sync the preference as a side effect server-side).

## Fix
`userCurrency` (and equivalent local variables) now prefer the active workspace's `default_currency`, falling back to the user's `currency_display` preference only when no workspace is loaded yet:

```ts
const userCurrency = current?.default_currency ?? user?.preferences?.currency_display ?? 'USD'
```

Applied consistently across all pages/components that read `user.preferences.currency_display` for display/formatting purposes. Per-record currency pickers for creating new data (new account, new invoice, new goal, etc.) were left as-is since those are user-editable form defaults, not the reported display bug.

## Testing
- Verified no other files still read `user.preferences.currency_display` for display formatting without falling back through the workspace's `default_currency`.
- No backend changes needed; `Workspace.default_currency` was already available via `useWorkspace()`.

Display currency now follows the active workspace’s `default_currency`. If no workspace is loaded, the app uses the signed-in user’s preference, then `USD`.

- Balances, totals, accounts, dashboards, transactions, budgets, goals, assets, reports, invoices, and related views use the workspace currency.
- Creation forms keep their existing user-editable currency defaults.
- The change affects frontend display only.

### DIFF
--- a/frontend/src/components/app-layout.tsx
+++ b/frontend/src/components/app-layout.tsx
@@ -88,7 +88,8 @@ export function AppLayout() {
   const { t } = useTranslation()
   const { user, logout, updateUser } = useAuth()
   const { activeAccountIds } = useCollectionFilter()
-  const userCurrency = user?.preferences?.currency_display ?? 'USD'
+  const { current: currentWorkspace } = useWorkspace()
+  const userCurrency = currentWorkspace?.default_currency ?? user?.preferences?.currency_display ?? 'USD'
   const locale = useDisplayLocale()
   const { theme, setTheme, resolvedTheme } = useTheme()
   const location = useLocation()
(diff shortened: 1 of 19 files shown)


tassionoronha:
Thanks for taking this on! Backend FX always converts to the user's primary currency, not the workspace default, so this would mislabel amounts. Closing — primary_currency in the API is the way in.

Issue #624, opened 2026-08-19
mBank Credit Card: Available Credit Displayed as Used Credit


szddawe:
## Description

When a credit card is imported from mBank, the credit limit and available credit appear to be interpreted in reverse.

For example, if the credit card has a total credit limit of €5,000 and €100 has been spent, the correct situation is:

- Credit limit: €5,000
- Used credit: €100
- Available credit: €4,900

However, Securo displays the values as if €4,900 had been spent and only €100 were available.

The credit card information is therefore effectively displayed in the opposite direction.

## Steps to Reproduce

1. Connect an mBank account containing a credit card to Securo.
2. Let Securo import the credit card data.
3. Use a credit card with a €5,000 credit limit.
4. Spend €100 using the card.
5. Open the imported credit card in Securo.
6. Compare the displayed used and available credit with the values shown by mBank.

## Expected Behavior

For a credit card with a €5,000 credit limit and €100 spent, Securo should show:

- Credit limit: €5,000
- Used credit: €100
- Available credit: €4,900

The amount already spent should reduce the available credit.

## Actual Behavior

Securo appears to interpret the imported values in reverse.

With a €5,000 credit limit and €100 spent, it displays the situation as if:

- €4,900 had been spent
- only €100 were available

This results in an incorrect representation of the credit card's financial status.

## Environment

- Deployment: Docker
- Browser: Firefox 153
- OS: Windows 11

## Screenshots

<img width="968" height="280" alt="Image" src="https://github.com/user-attachments/assets/b91d2692-b8ff-46f8-a759-7d459b3330d8" />

## Additional Context

The issue occurs with credit card data imported from mBank.

The problem appears to be related to how the credit card balance or available-credit information returned by the mBank connection is interpreted by Securo.

A possible cause is that the value provided by mBank represents available credit while Securo interprets it as used credit, or vice versa. The mapping of the credit-card-related fields from the mBank/Enable Banking response should therefore be checked.

joaopgmaria:
I can confirm this also happens for Portuguese bank Santander through Enable Banking

PR #808, opened 2026-09-01
fix(enable-banking): credit-card balance handling

### LINKED ISSUE #800: [Bug] Enable Banking credit-card balances, available credit, and opening reconciliation use inconsistent semantics
## Description

Credit card balances imported through Enable Banking are interpreted inconsistently across provider parsing, account serialization, available-credit calculation, and opening-balance reconciliation.

Enable Banking can return several balances for the same card account, each with different semantics:

- `CLBD` (`ClosingBooked`): the balance of booked transactions;
- `XPCD` (`Expected`): the expected balance including pending transactions;
- `ITAV` or `CLAV` (`InterimAvailable` / `ClosingAvailable`): the amount currently available to spend.

Securo currently selects a single generic balance using this priority:

```text
CLBD → ITAV → CLAV → XPCD → OPBD → OPAV
```

The selected value is stored as `Account.balance`, while the semantic distinction between booked balance, expected balance, and available credit is discarded. Credit-card code later applies sign conversion independently in several places.

This produces incorrect results for Enable Banking because its card-balance sign follows the cash-account perspective: a positive booked balance can mean that the cardholder has credit, while a negative balance means debt. Securo's existing internal convention for connected credit cards is the opposite:

```text
positive Account.balance = card debt
negative Account.balance = card credit / overpayment
```

The provider response must therefore be normalized once at the provider boundary, and the different balance types must remain available for their intended downstream purposes.

## Steps to Reproduce

1. Connect a credit card through Enable Banking.
2. Use a card with a synthetic credit limit of `3000.00 EUR`.
3. Have Enable Banking return the following sanitized balance payload:

   ```json
   {
     "balances": [
       {
         "name": "Balance of type \"ClosingBooked\"",
         "balance_amount": {
           "currency": "EUR",
           "amount": "25.00"
         },
         "balance_type": "CLBD",
         "reference_date": "2025-01-15"
       },
       {
         "name": "Balance of type \"Expected\"",
         "balance_amount": {
           "currency": "EUR",
           "amount": "-27.00"
         },
         "balance_type": "XPCD",
         "reference_date": "2025-01-15"
       },
       {
         "name": "Balance of type \"InterimAvailable\"",
         "balance_amount": {
           "currency": "EUR",
           "amount": "2973.00"
         },
         "balance_type": "ITAV",
         "reference_date": "2025-01-15"
       }
     ]
   }
   ```

4. Have the provider return these booked transactions:

   ```text
   +100.00 credit
    -30.00 debit
    -25.00 debit
    -20.00 debit
   ```

   Their net booked amount is:

   ```text
   +25.00 EUR
   ```

5. Have the provider additionally return these pending transactions:

   ```text
   -40.00 debit
   -12.00 debit
   ```

6. Trigger a connection sync.
7. Open the credit-card account and inspect its balance, available credit, utilization, and generated opening-balance transaction.

## Expected Behavior

Securo should preserve and use each Enable Banking balance according to its meaning:

```text
Booked balance (CLBD):       +25.00 EUR credit at the provider
Expected balance (XPCD):    -27.00 EUR including pending
Available credit (ITAV):   2973.00 EUR
```

At the provider boundary, booked and expected card balances should be normalized to Securo's internal positive-for-debt convention:

```text
Provider CLBD +25.00 → internal booked balance -25.00
Provider XPCD -27.00 → internal expected balance +27.00
Provider ITAV 2973.00 → available credit 2973.00, unchanged
```

Opening-balance reconciliation should use only the normalized booked balance and posted transactions. Pending transactions must not be included because `CLBD` excludes them.

For this example:

```text
reconciliation target = -internal booked balance
                      = +25.00

booked transaction sum = +25.00

opening balance = target - booked transaction sum
                = 0.00
```

No synthetic opening-balance transaction should be created.

Available credit should prefer the explicit provider value:

```text
2973.00 EUR
```

It must not be reconstructed from the booked balance when `ITAV` or `CLAV` is available, because those values may already reflect pending transactions.

When no provider-reported available-credit value exists, the fallback should use Securo's stored card-balance convention explicitly:

```text
available credit = credit limit - internal booked balance
```

Examples for a `3000.00` limit:

```text
150.00 debt        → 2850.00 available
0.00 balance       → 3000.00 available
-25.00 card credit → 3025.00 available
```

Existing generic balance selection for checking and savings accounts should remain unchanged. Pluggy and SimpleFIN behavior must not regress.

## Actual Behavior

Securo selects `CLBD = +25.00` as the generic account balance and stores it without converting Enable Banking's credit-card sign convention.

Connected-card serialization later negates the stored value, while opening-balance reconciliation independently targets the negated account balance:

```text
stored Account.balance = +25.00
reconciliation target  = -25.00
booked transaction sum = +25.00

synthetic opening balance = -25.00 - 25.00
                          = -50.00
```

Securo therefore creates an incorrect synthetic opening-balance transaction of `-50.00 EUR`, even though the imported booked transactions already reconcile exactly to the provider's booked balance.

The explicit Enable Banking available-credit value is discarded. Securo instead reconstructs available credit from an already-negated display balance:

```text
credit limit    = 3000.00
display balance = -25.00
available       = 2975.00
```

The correct value reported by Enable Banking is:

```text
2973.00
```

The difference exists because the provider's available balance already includes the pending card transactions represented by `XPCD`.

## Environment

- Deployment: Docker
- Browser: Any
- OS: Any

## Additional Context

The core problem is that Securo collapses different provider balance semantics into one generic field and then applies credit-card sign conversion in multiple downstream components.

A provider-neutral account snapshot model should preserve at least:

```text
booked balance
expected balance
available credit
```

The exact internal invariant should be documented and consistently enforced:

```text
For credit cards:
positive stored balance = debt
negative stored balance = credit / overpayment
```

Provider-reported available credit is an absolute spendable amount and must not be negated. Expected balance may include pending transactions and must not be combined with an additional manual subtraction of pending transactions.

Existing connected Enable Banking credit-card rows also require a one-time migration so their stored booked balances follow the corrected internal sign convention immediately, rather than remaining incorrect until the next successful provider sync.

Regression coverage should include:

1. Enable Banking card debt: `CLBD = -150.00` becomes `150.00` internal debt.
2. Zero card balance: `CLBD = 0.00` remains zero.
3. Card credit: `CLBD = +25.00` becomes `-25.00` internally.
4. `XPCD` and `ITAV` remain distinct from `CLBD`.
5. Provider-reported available credit takes precedence over fallback calculation.
6. Available-credit fallback handles debt, zero, and card credit correctly.
7. Booked transactions that exactly match `CLBD` produce no opening balance.
8. Incomplete transaction history produces the correct synthetic opening balance.
9. Pending transactions are not included when reconciling against `CLBD`.
10. Checking and savings account balance selection remains unchanged.

jfrcom:
## What
Fix Enable Banking credit-card balance handling by preserving the different ISO 20022 balance semantics instead of collapsing them into a single generic value.
- Treat `CLBD`/`OPBD` as the booked card balance.
- Preserve `XPCD` separately as the expected balance, which may include pending transactions.
- Preserve `ITAV`/`CLAV`/`OPAV` separately as provider-reported available credit.
- Normalize Enable Banking card balances at the provider boundary to Securo's internal convention: positive values represent debt and negative values represent card credit or overpayment.
- Persist `expected_balance` and `available_credit` on connected credit-card accounts and expose the expected balance through the account API.
- Prefer provider-reported available credit and only fall back to `credit_limit - booked_balance` when the provider does not supply it.
- Reconcile connected credit-card opening balances against booked transactions only, excluding pending transactions from a booked-balance target.
- Add migration `085` to create the snapshot columns and normalize existing Enable Banking credit-card balances once.
- Keep the existing generic balance selection for non-credit-card accounts and avoid changing Pluggy and SimpleFIN balance behavior.
## Why
Enable Banking may return booked, expected, and available balances for the same credit-card account, but these values have different meanings. Previously Securo discarded that distinction and applied credit-card sign conversion in multiple downstream locations.
This could invert the displayed balance incorrectly, generate a synthetic opening-balance transaction even when the imported booked transactions already reconciled, and calculate available credit from the wrong balance. Keeping the snapshots separate and normalizing the sign once gives account summaries, utilization, available credit, and opening reconciliation a consistent source of truth.
## How to Test
1. Start from a database at migration `084` and run `alembic upgrade head`; verify that `expected_balance` and `available_credit` are added to `accounts` and Alembic reports `085` as the single head.
2. Connect or mock an Enable Banking credit card that returns `CLBD`, `XPCD`, and `ITAV` balances.
3. Verify that a provider booked balance of `-150.00` is stored as `150.00` debt, `0.00` remains zero, and `25.00` provider credit is stored as `-25.00` card credit.
4. Verify that `XPCD` is stored separately as the expected balance and that `ITAV`/`CLAV` is preserved as an absolute available-credit value without sign inversion.
5. Verify that the account API displays the normalized credit-card balance and expected balance using the existing connected-card display convention.
6. Set a credit limit and confirm that provider-reported available credit takes precedence over the calculated fallback.
7. Remove the provider-reported available-credit snapshot and verify the fallback for debt, zero balance, and card credit using `credit_limit - booked_balance`.
8. Import booked transactions whose signed total exactly matches `CLBD`; verify that no synthetic opening-balance transaction is created.
9. Add pending transactions alongside the booked transactions and verify that they do not affect reconciliation against `CLBD`, while `XPCD` can still represent the expected balance including those pending entries.
10. Verify that incomplete booked history creates only the required opening-balance adjustment.
11. Sync checking and savings accounts through Enable Banking and verify that their existing balance-selection behavior is unchanged.
12. Run the targeted backend tests:
    ```bash
    cd backend
    pytest -q tests/test_migration_chain.py tests/test_providers_enable_banking.py tests/test_account_service.py
    ```
13. Run the frontend checks:
    ```bash
    cd frontend
    npm run lint
    npm run build
    ```
## Related Issue
Closes #800 ([Enable Banking credit-card balances, available credit, and opening reconciliation use inconsistent semantics](https://github.com/securo-finance/securo/issues/800)).
## Checklist

### DIFF
--- a/backend/alembic/versions/078_credit_card_balance_snapshots.py
+++ b/backend/alembic/versions/078_credit_card_balance_snapshots.py
@@ -1,0 +1,59 @@
+"""preserve credit card balance snapshots
+
+Revision ID: 078
+Revises: 077
+Create Date: 2026-08-31
+"""
+
+from typing import Sequence, Union
+
+import sqlalchemy as sa
+from alembic import op
+
+revision: str = "078"
+down_revision: Union[str, None] = "077"
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+
+def upgrade() -> None:
+    op.add_column(
+        "accounts",
+        sa.Column("expected_balance", sa.Numeric(precision=15, scale=2), nullable=True),
+    )
+    op.add_column(
+        "accounts",
+        sa.Column("available_credit", sa.Numeric(precision=15, scale=2), nullable=True),
+    )
+    # Enable Banking cash-account signs are the inverse of Securo's card
+    # invariant. Existing rows were stored verbatim, so normalize them once;
+    # later syncs perform this conversion at the provider boundary.
+    op.execute(
+        sa.text(
+            """
+            UPDATE accounts
+            SET balance = -accounts.balance
+            FROM bank_connections
+            WHERE accounts.connection_id = bank_connections.id
+              AND accounts.type = 'credit_card'
+              AND bank_connections.provider = 'enable_banking'
+            """
+        )
+    )
+
+
+def downgrade() -> None:
+    op.execute(
+        sa.text(
+            """
+            UPDATE accounts
+            SET balance = -accounts.balance
+            FROM bank_connections
+            WHERE accounts.connection_id = bank_connections.id
+              AND accounts.type = 'credit_card'
+              AND bank_connections.provider = 'enable_banking'
+            """
+        )
+    )
+    op.drop_column("accounts", "available_credit")
+    op.drop_column("accounts", "expected_balance")
--- a/backend/alembic/versions/085_credit_card_balance_snapshots.py
+++ b/backend/alembic/versions/085_credit_card_balance_snapshots.py
@@ -1,7 +1,7 @@
 """preserve credit card balance snapshots
 
-Revision ID: 078
-Revises: 077
+Revision ID: 085
+Revises: 084
 Create Date: 2026-08-31
 """
 
@@ -10,8 +10,8 @@ from typing import Sequence, Union
 import sqlalchemy as sa
 from alembic import op
 
-revision: str = "078"
-down_revision: Union[str, None] = "077"
+revision: str = "085"
+down_revision: Union[str, None] = "084"
 branch_labels: Union[str, Sequence[str], None] = None
 depends_on: Union[str, Sequence[str], None] = None
 
--- a/backend/app/models/account.py
+++ b/backend/app/models/account.py
@@ -23,7 +23,9 @@ class Account(Base):
     workspace_id: Mapped[uuid.UUID] = mapped_column(
         UUID(as_uuid=True), ForeignKey("workspaces.id", ondelete="CASCADE"), index=True
     )
-    connection_id: Mapped[Optional[uuid.UUID]] = mapped_column(UUID(as_uuid=True), ForeignKey("bank_connections.id"), nullable=True)
+    connection_id: Mapped[Optional[uuid.UUID]] = mapped_column(
+        UUID(as_uuid=True), ForeignKey("bank_connections.id"), nullable=True
+    )
     external_id: Mapped[Optional[str]] = mapped_column(String(255), nullable=True)
     name: Mapped[str] = mapped_column(String(255))
     display_name: Mapped[Optional[str]] = mapped_column(String(255), nullable=True)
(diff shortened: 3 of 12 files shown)


rhuanbarreto:
Two more Norwegian data points for this, both via Enable Banking on `0.14.5`, in case they help with the design discussion.

**Bank Norwegian (NO), credit card** — matches #800 exactly.

- In the bank app the card has a little over half of its limit in use.
- Securo stored the balance as a negative number and, because the account is `credit_card`, displayed the same amount as **positive** — the debt showed up as an asset in net worth.
- The synthetic opening-balance row came out as a **credit** of roughly 1.5× the real debt.
- Switching the account type to `checking` made the displayed balance negative, and on the next sync the opening row was rebuilt as a **debit**. After that, the sum of the account's transactions matched the balance to the cent.

So this bank reports card debt with the bank-account sign (negative = owed), and the blanket negation in `account_service.py:168` / the `-account.balance` target at `:485` inverts it.

**Nordea (NO), credit card** — the sign looks right, but the amount doesn't.

- The card is paid in full every month, with small invoices.
- Securo nevertheless shows a debt of about 30× a typical monthly invoice, carried entirely by a **debit** opening-balance row created on the first sync.

That size looks like available credit rather than used credit, which would fit `_pick_balance` falling through to `ITAV`/`CLAV` when there's no `CLBD` (same symptom as #624). I haven't captured the raw `/accounts/{uid}/balances` response for this card, so I can't confirm which balance type it sent — I can pull it if that's useful for testing #808.

Workaround for anyone hitting the Bank Norwegian case before this lands: set the account type to `checking` (sync keeps the type you set) and trigger a sync.

PR #809, opened 2026-09-01
chore(frontend): add react-is dependency

jfrcom:
## Summary

Adds `react-is` as a direct frontend dependency.

Recharts imports `react-is`, but it was missing from `frontend/package.json`, causing Vite/Rolldown builds to fail with:

```text
Rolldown failed to resolve import "react-is"
```

## Changes
- Added react-is matching the current React 19 version
- Updated package-lock.json

## Testing
- npm run build now completes successfully

Adds `react-is` version `^19.2.0` as a direct frontend dependency. This fixes Recharts imports that caused Vite/Rolldown build failures.

- Frontend builds without the missing `react-is` dependency error.

Risk: none


tassionoronha:
Thanks! Happy to take this. The description mentions the lock file, but only package.json changed — could you run npm install and commit package-lock.json too? Then it's good to go.

tassionoronha:
LGTM now,  thanks @jfrcom !

PR #812, opened 2026-09-01
fix(import): preselect CSV categories in preview

joaolucenalima:
## What

- Preselect an existing category in the import preview when the CSV category name matches exactly.
- Keep transaction rules authoritative when both the CSV and a matching rule provide a category.
- Preserve the existing exact-name matching behavior used by the final import.

## Why

CSV category values were already parsed and resolved by exact name during the final import, but matching categories were not exposed as suggestions in the preview. As a result, the frontend could not show them preselected before confirmation.

This PR only aligns the preview with the existing import behavior. It intentionally does not add case-insensitive, whitespace-collapsing, or Unicode-normalized category matching.

## How to Test

1. Import a CSV with a `categoria` value that exactly matches an existing category and confirm it is preselected in the preview.
2. Add a matching transaction rule with a different category and confirm the rule category is preselected instead.
3. Confirm that a category name with different casing remains unmatched, preserving the existing exact-name semantics.

## Validation

- `pytest tests/test_import_api.py -q`: 18 passed.

## Related Issue

No related issue.

## Checklist

- [x] Affected backend tests pass.
- [x] Existing exact-name matching semantics are preserved.
- [x] No user-facing strings changed.
- [x] No frontend or migration changes.

CSV category suggestions now follow the final import precedence: explicit user choice, matching rule, then CSV category. Exact category names remain unchanged, and uncategorized selection is preserved.

- Preview shows matching rule categories before CSV categories.
- Preview shows CSV categories when no rule matches.
- Explicit category changes override suggestions.
- Selecting uncategorized keeps the transaction uncategorized.
- Failed CSV rows appear in the preview with error details.


tassionoronha:
Thanks for digging into this! One blocker: preview sets category_set=False so rules beat the CSV, while import derives it from category_id so the CSV wins. The two land differently.

joaolucenalima:
Thanks for catching this — I fixed the mismatch between preview and the final import flow.

During preview, a matching category rule can override the CSV category; if no rule matches, the CSV category is shown as the default. The UI now treats that preview value as a suggestion only, and only sends `category_id` when the user explicitly changes the category.

On import, the precedence is now: explicit user choice -> matching rule -> CSV fallback. Choosing uncategorized still sends `force_uncategorized=true`, so automatic category assignment is skipped.

I also added coverage for rule-over-CSV, CSV fallback, and explicit user overrides.

jorishaenseler15:
Hi! Thanks for working on this. 

Just a quick heads-up that this was already implemented and merged into `main` via #576 (closing #566). You can pull the latest `main` to check it out!

tassionoronha:
All good for me now :) 
Thanks @joaolucenalima and @jorishaenseler15 for the review! 🙌🏼

PR #814, opened 2026-09-02
feat: currency formatting with narrow currency symbol rather than ISO currency text code

turkelali:
## What

All contributed currencies have their currency symbol in codes, but they are not been utilized in interface. I noticed only USD, EUR and GBP are showing their currency symbols, all other currency are just showing ISO Currency code rather than symbol. For to get professional look I suggest to utilize currency symbol in Securo. This will affect all currencies like below. Below screenshots shows changes affecting home page only:

**Before Changes**
image
image
image

**After Changes**
image
image
image

## Why

Utilizing explicit narrow currency symbols provides a cleaner, more professional, and highly cohesive user interface. The symbols are already present within the code — this small change simply enables them.

## How to Test

1. Navigate to your workspace settings
2. Change the default workspace currency to a regional currency
3. Refresh the dashboard and verify that the layout displays the specific currency symbol (e.g., `$`, `₼`, `₺`, `¥`) rather than the raw ISO text string.

## Checklist

Currency formatting now uses narrow symbols where available, with locale-aware amount input parsing and formatting.

- Shows symbols such as `$`, `₺`, `₼`, and `¥`.
- Formats balances, income, expenses, and net amounts.
- Reads and displays typed amounts using the selected separators.


tassionoronha:
Thanks for raising this! Trouble is narrowSymbol maps USD, CAD, AUD, MXN and ARS all to $, so multi-currency users can't tell them apart. Would need to be scoped to the primary currency.

tassionoronha:
I tried this in the browser and it does look cleaner for a single-currency workspace. My worry is multi-currency, which is common here: with the narrow symbol, USD, CAD, AUD, NZD, MXN, ARS and a few others all become a bare $, so a CAD account and a USD account look identical side by side. One idea that keeps the win: use the narrow symbol only for the workspace's own display currency and the full symbol for everything else. If you want to go that way I'm glad to review it.

PR #822, opened 2026-09-03
fix(accounting): correct inverted cash/accrual naming for credit card mode

vhsantos26:
Fixes #821

## Summary

The credit-card accounting mode setting had the cash/accrual terminology backwards relative to standard accounting usage:

- `accrual` bucketed purchases by the bill's due date — that's cash-basis behavior (money actually leaving).
- `cash` bucketed purchases by the purchase date — that's accrual-basis behavior (expense recognized when incurred).

The invoicing module (`Invoice`/`competence_date`) already uses these terms correctly; this inversion was isolated to the credit-card accounting-mode feature.

## Change

Renamed the internal setting values to be self-describing (`purchase_date` / `invoice_due_date`) instead of reusing the ambiguous cash/accrual jargon, and assigned the correct accounting label to each in the UI, across all 12 locales. Existing deployments keep their current behavior unchanged — a data migration (090) remaps the single global `credit_card_accounting_mode` setting row so only the label changes, not what it does.

## Test plan

- [x] `ruff check .`, `ty check .` — clean
- [x] Backend: full suite (`pytest -n auto`) — 3439 passed, 92.55% coverage
- [x] New unit test for the migration (`test_accounting_mode_migration.py`) covering upgrade/downgrade
- [x] Alembic revision chain check — single line, head 090
- [x] Frontend: lint, build (type check), test — all clean
- [x] Verified in own fork first: vhsantos26/securo#48 (merged, CI green)

Corrects credit-card accounting labels without changing their existing date behavior. Migration 090 preserves current settings while renaming the internal values.

- Purchase-date mode is labeled accrual basis.
- Invoice-due-date mode is labeled cash basis.
- Labels and descriptions are updated in all 12 locales.
- Adds invoice tracking, linked-card filters, investment categories, and related UI updates.
- Adds VPS deployment and branch-sync workflows.

Risk: database migration, money math, sync provider.

PR #824, opened 2026-09-03
fix(accounting): correct inverted cash/accrual naming for credit card mode

### LINKED ISSUE #821: Credit card accounting mode: 'cash'/'accrual' labels are inverted relative to standard accounting terms
## Summary

The credit-card accounting mode setting (Admin → Contabilidade) offers two options that group card purchases either by purchase date or by the invoice's due date. The internal values and labels currently have the accounting terminology backwards.

- **"Regime de caixa"** (`credit_card_accounting_mode=cash`, default) buckets a purchase by `Transaction.date` — the date it happened.
- **"Regime de competência"** (`credit_card_accounting_mode=accrual`) buckets a purchase by the bill's due date (`Transaction.effective_date`).

Standard accounting usage is the opposite:
- **Regime de competência (accrual)** recognizes an expense when it's *incurred* — the purchase date, independent of payment.
- **Regime de caixa (cash)** recognizes it when the money actually *leaves* — the bill's due/payment date.

So today's "accrual" mode is actually cash-basis behavior, and vice versa. The code's own comments say this out loud without noticing the contradiction, e.g. `dashboard_service.py`: *"Reporting mode is a global app setting... When 'accrual', aggregation queries bucket credit card transactions by the bill's due date... gives a true cash-flow view."* Calling a due-date view "accrual" while describing it as "cash-flow" is the tell.

## Impact

Not a data bug — reports/dashboard/budgets compute correctly for whichever mode is selected. But the labels mislead anyone reasoning about the setting from accounting first principles: a purchase made in August whose bill is due in September appears in the September filter under "Regime de competência," which is the opposite of what a user familiar with the term would expect.

Confirmed the invoicing module (`Invoice`/`competence_date` in `invoice.py`/`invoice_service.py`) already uses "accrual"/"competência" correctly — this inversion is isolated to the credit-card accounting-mode feature.

## Proposed fix

Rename the internal setting values to be self-describing instead of reusing the accounting jargon (`purchase_date` / `invoice_due_date`), and assign the correct accounting label + description to each in the UI:

- `purchase_date` (new default, same behavior as today's `cash`) → labeled "Regime de competência"
- `invoice_due_date` (same behavior as today's `accrual`) → labeled "Regime de caixa"

Requires a small data migration remapping the single existing `credit_card_accounting_mode` app setting row (`cash` → `purchase_date`, `accrual` → `invoice_due_date`) so existing deployments keep their current behavior — only the label changes.

vhsantos26:
Fixes #821

(Supersedes #822, which was opened from the wrong base and showed an unrelated ~30k-line diff.)

## Summary

The credit-card accounting mode setting had the cash/accrual terminology backwards relative to standard accounting usage:

- `accrual` bucketed purchases by the bill's due date — that's cash-basis behavior (money actually leaving).
- `cash` bucketed purchases by the purchase date — that's accrual-basis behavior (expense recognized when incurred).

The invoicing module (`Invoice`/`competence_date`) already uses these terms correctly; this inversion was isolated to the credit-card accounting-mode feature.

## Change

Renamed the internal setting values to be self-describing (`purchase_date` / `invoice_due_date`) instead of reusing the ambiguous cash/accrual jargon, and assigned the correct accounting label to each in the UI, across all locales (including `el.json`). Existing deployments keep their current behavior unchanged — a data migration (086) remaps the single global `credit_card_accounting_mode` setting row so only the label changes, not what it does.

## Rollback plan

`alembic downgrade -1` reverts the row (`purchase_date` → `cash`, `invoice_due_date` → `accrual`). The previous backend/frontend release already understands those values, so a rollback is safe.

**Known rollout risk:** the migration runs as a Helm `post-upgrade` hook (weight 5), which fires once the new Deployment is applied — not after old pods finish draining. A pod still running the previous release's code can read the freshly-migrated `invoice_due_date` value, fail to recognize it, and fall back to `cash` behavior for the rest of its lifetime (typically under a minute during a rolling update). New pods already handle the reverse case — `admin_service.get_credit_card_accounting_mode` normalizes legacy `cash`/`accrual` values read before migration 086 lands.

Impact is limited to transient misclassification of credit-card reports/dashboards for whichever old pod is still alive; no data is written incorrectly, and it self-resolves once the old pod terminates. Accepting this risk rather than splitting the rename into a two-release rollout, given the narrow window and that the setting only affects read-time computation of a single global row.

## Test plan

- [x] `ruff check .`, `ty check .` — clean
- [x] Backend: full suite (`pytest -n auto`) — 3421 passed, 92.57% coverage
- [x] New unit test for the migration (`test_accounting_mode_migration.py`) covering upgrade/downgrade
- [x] Alembic revision chain check — single line, head 086
- [x] Frontend: lint, build (type check), test — all clean (one unrelated flaky timeout in `pages.contract.test.ts` on the full run, passes in isolation)
- [x] Verified in own fork first: vhsantos26/securo#48 (merged, CI green)

Renames credit-card accounting modes to match their actual behavior. A migration preserves existing behavior while updating stored values and labels.

- Purchase-date grouping is labeled accrual basis.
- Invoice-due-date grouping is labeled cash basis.
- Reports and notices use the corrected labels.
- Legacy settings are converted automatically.

Risk: database migration and money math.


tassionoronha:
Merged main into the branch. Two locales (hi, ja) were added after you opened this and still had the old key names, so I renamed and translated them to match the others. I also replaced the em dash in the cash-basis description with a plain clause in every language, since we keep that character out of UI text. Nothing else changed.

tassionoronha:
One more thing: the migration was numbered 086, which main already has, so I renumbered it to 090. Two other open PRs also carry a 090, so the second one to land will need a bump.

PR #833, opened 2026-09-04
feat(accounts): add loan account type via a centralized account-type registry

### LINKED ISSUE #823: [Feature] Add support for Loan account type

## Problem

Currently, Securo only supports `checking`, `savings`, `investment`, `credit_card`, and `wallet` account types. There is no dedicated `loan` account type to manage debt such as personal loans, mortgages, or auto loans.

Right now, the only way to track a loan is to create a `checking` account with a negative balance. This causes several issues:
- **Distorted Balances:** It incorrectly reduces overall checking/liquid balances and distorts financial reporting.
- **Unintuitive Tracking:** Managing pending loan balances as a negative checking balance is confusing and does not reflect actual loan liability tracking.

## Proposed Solution

Add a new `loan` account type to Securo:
1. **First-Class Loan Account Type:** Include `loan` as an official account type choice when creating or editing accounts.
2. **Proper Liability Handling:** Treat loan accounts as liability balances so pending debt is tracked accurately without distorting checking or liquid balances.
3. **Future Extensibility:** Set up the foundation for future loan features such as repayment schedules, interest rate tracking, and payoff progress.

## Alternatives Considered

- **Using a Checking Account with negative value:** Current workaround, but distorts checking balance totals and causes confusing UI representations.
- **Using a Credit Card Account:** Does not fit fixed term loans, mortgages, or structured loan repayments.

## Additional Context

- This will require updating backend schemas/models (`AccountType` enum) and frontend components (account creation modal, icons, and account grouping).
- This feature will greatly help users who want to track personal loans or debt alongside their liquid assets and investments.

Sainimayank1:
## What

Adds `loan` as a first-class account type, and introduces the account-type registry it needed in order to exist.

The registry is the substance of the change. `credit_card` was previously spelled as a literal at every site that cared about debt-shaped behaviour — nine separate sites for the sign convention alone — so adding a second liability type meant finding all nine, and missing one would have produced a quietly wrong number rather than an exception. The behaviours are now named as traits in one place (`backend/app/core/account_types.py`, `frontend/src/lib/account-types.ts`), and call sites ask about the trait instead of the type.

Three traits, deliberately independent, because `loan` is the first type to split them apart:

- **`is_liability`** — money owed. Drives the sign convention and the assets-vs-liabilities split. A loan sets this, like a card.
- **`counts_pending_in_balance`** — is an authorised-but-uncleared row already real? True for a card, whose bill will include it. False for a loan, whose balance moves only when a payment settles.
- **`has_billing_cycle`** — does it bill in statement cycles? Drives `effective_date` bucketing, bill-vs-P&L totals, and the cycle-date fields. A loan has a due date but no cycle, so none of that machinery applies.

`metadata_fields` names the columns each type owns, so changing an account's type clears the columns the new type does not own and an account is never left half credit-card. Adding a type is now a row; adding a behaviour is a field plus the sites that honour it.

Also included: `AccountCreate`/`AccountUpdate` reject unknown types on the way in, so a typo is a 422 rather than an account that silently renders as checking. `AccountRead` deliberately stays a plain `str` — the column is free-form and predates the registry, so a database written by an older build may hold a value this build has never heard of, and refusing to serialise those would break the whole accounts list.

Enable Banking's `LOAN` now maps to `loan` instead of the `checking` placeholder it used before.

## Why

Closes #823.

Without a loan type the only way to track debt was a `checking` account with a negative balance, which reduced liquid balances and distorted reporting. A loan now carries its debt on the liability side: it stays out of the dashboard's "Available balance" headline (checking/savings only) and lands in the liabilities half of net worth, with the sign convention handled by the trait rather than by a type check.

The registry is also the extensibility half of the issue's ask. Loan terms — principal, rate, maturity — are the obvious next columns, and they attach as a `metadata_fields` entry on the existing row without touching the type-change path.

## How to Test

1. Create an account of type **Loan** with an outstanding balance entered as a positive number, e.g. 12,000.
2. Confirm it displays as negative, is grouped under liabilities, and is **not** counted in the dashboard's "Available balance" tile.
3. Confirm it is counted against Net Worth.
4. Switch an existing credit card to **Loan** and confirm the card-only fields (credit limit, statement close day, payment due day, minimum payment, card brand/level) are cleared, and that the cycle-date UI disappears.
5. Switch it back and confirm no stale card metadata is resurrected.
6. Post a transaction to the loan and confirm its effective date is the transaction date, with no statement bucketing.
7. `backend/tests/test_account_types.py` covers the above at the unit level, including the case that motivated the split: a loan must behave like a card on the liability axis and unlike it on the pending and billing-cycle axes.

## Related Issue

Closes #823.

## Checklist

Two new keys (`accounts.typeLoan`, `accounts.balanceLoanHint`) plus `accounts.typeOther`, which the registry surfaces as a real choice, are translated across all 14 locales.

## Notes for review

- `other` is now an offered type rather than only a silent fallback. It was already advertised as a filterable value on the agent-facing MCP contract before this change, so the registry makes the two agree.
- Unknown types still resolve to a plain asset-shaped fallback rather than raising, so one stale row cannot 500 the accounts list.
- Rebased on `main` at d4b698b. The dashboard hero card added there filters `credit_card` directly for its open-card-bills line, which stays correct for loans: a loan has no bill, and its already-negated balance reconciles inside the net-worth breakdown. Those two call sites were left alone rather than folded into the registry, to keep this PR to the type change.

Adds `loan` as a supported account type and treats loans as liabilities. Centralized account rules also update balances, pending transactions, billing cycles, metadata, reporting, and provider mappings.

- Create and edit loan accounts.
- Track loan balances as liabilities without reducing liquid balances.
- Import loans from Enable Banking.
- Clear invalid account details when the account type changes.
- Use loan labels, icons, colors, and balance guidance.
- Reject unknown account types for new or updated accounts.

Risk: money math and sync provider. No database migration or auth changes.


tassionoronha:
Hey @Sainimayank1, I appreciate your effort but since this is a core feature present in our official roadmap this will be better implemented by our core contributors which has more context of our application. 

I suggest you start collaborating with minor improvements and not highly related to the financial engine in the first moment. 

I hope you understand. 

Cheers!

PR #838, opened 2026-09-04
fix(transactions): persist external IDs for manually created transactions

### LINKED ISSUE #837: [Bug] Manual transaction API drops accepted external_id
## Description

POST /api/transactions accepts external_id, but the manual creation service does not copy it onto Transaction. The database stores NULL, so the normal POST response, GET-by-ID response, and list response expose NULL.

## Steps to Reproduce

1. Create a manual transaction through POST /api/transactions with    "external_id": "client-tx-001".
2. Inspect the 201 response.
3. Fetch the returned ID and list the account's transactions.

## Expected Behavior

The supplied opaque ID is persisted and returned by each response. Omitting it continues to return null.

## Actual Behavior

All responses return external_id: null.

## Proposed Scope

Pass data.external_id into the existing Transaction(...) constructor and cover POST, single-read, and list-read behavior. No migration, uniqueness constraint, UI change, or installment-series behavior is proposed.

z0diak:
## What

Persist the optional external_id supplied to POST /api/transactions when creating a single manual transaction.

## Why

The API schema already accepts and returns external_id, but the manual creation service dropped the input before saving the transaction. This caused the create response and later reads to return null, preventing API clients from reconciling the row with their stable external reference.

This is backward compatible: transactions created without an external ID, and historical manual rows whose input was discarded, remain null. Existing stored external IDs are unchanged. No migration or UI change is needed.

## How to Test

1. Run the focused service and API regression tests.
2. POST a manual transaction with external_id: "client-tx-api-001" and verify the 201 response preserves it.
3. GET the returned transaction and list its account's transactions; verify the same value appears in both responses.
4. Run the full backend lint, type-check, test/coverage, lock, and migration checks plus frontend lint, test, and build.

## Related Issue

Closes https://github.com/securo-finance/securo/issues/837

## Checklist

- [x] Backend tests pass (pytest)
- [x] Frontend lints clean (npm run lint)
- [x] Frontend builds (npm run build)
- [x] Translations updated (not applicable: no user-facing strings changed)

Manual transaction creation now persists the supplied `external_id`. The value remains available in creation, single-transaction, and account-list responses.

- Supplied `external_id` remains available after creation.
- Omitted `external_id` remains `null`.


z0diak:
@tassionoronha thanks for the branch actualisation. Let me know if there are any questions or concerns about this fix 🙏

tassionoronha:
LGTM @z0diak , thanks for that :)

PR #839, opened 2026-09-04
fix(rate-limit): derive client IP from a configured trusted-proxy hop

### LINKED ISSUE #801: Rate limiter buckets all clients together behind a reverse proxy (request.client.host is the proxy)
## Summary

`RateLimiter` in `backend/app/core/rate_limit.py` builds its Redis key from `request.client.host`. In the shipped deployment the backend sits behind the bundled nginx frontend (and usually behind another reverse proxy on top of that), and uvicorn is started without a `forwarded_allow_ips` that trusts it. The result is that `request.client.host` is always the proxy's container IP, so **every client on the internet shares a single rate-limit bucket** for each rate-limited endpoint.

## Reproduce

On a stock deployment from `docker-compose.prod.yml`, send one failed login through the frontend, then look at the key that was created:

```
$ docker exec <redis> redis-cli --scan --pattern "rate_limit:*"
rate_limit:/api/auth/login:172.26.0.3
```

`172.26.0.3` is the frontend container, not the client. Every request that reaches `/api/auth/login` from anywhere lands on that same key.

## Why it happens

- `backend/app/core/rate_limit.py` uses `request.client.host` directly.
- The backend entrypoint is `alembic upgrade head && uvicorn app.main:app --host 0.0.0.0 --port 8000`. uvicorn enables proxy headers by default but only trusts `127.0.0.1` unless `FORWARDED_ALLOW_IPS` is set, and the nginx container connects from a Docker bridge address, so the `X-Forwarded-For` it sets is discarded.
- `frontend/nginx.conf` does set `X-Real-IP` and `X-Forwarded-For` correctly, so the information is there. It just is not trusted downstream.

## Impact

The limiters affected are `login_rate_limit` (5 req / 60 s, applied to the custom auth router and to `/api/auth/2fa/verify`), `register_rate_limit` and `password_reset_rate_limit`.

The interesting consequence is not weakened brute-force protection: 5 attempts per minute against a TOTP code is still useless to an attacker. It is the opposite. **Anyone who can reach the login endpoint can lock the legitimate user out indefinitely** with a loop that spends 5 requests per minute. The `zadd` in `RateLimiter.__call__` runs before the count is checked, so rejected requests are added to the window too, which means the bucket never drains while the loop runs.

For an instance exposed to the internet, ordinary background scanning is enough to trigger this without anyone intending to. Bots routinely probe `/login`, `/register` and `/api/auth/signin`.

## Suggested fix

`FORWARDED_ALLOW_IPS=*` is not a good fix on its own: `X-Forwarded-For` is client-writable, and a proxy that appends rather than replaces leaves an attacker in control of part of the header, which turns a global limiter into a trivially bypassable one.

Options that seem sound:

1. Make the trusted proxy set explicit and configurable (a `TRUSTED_PROXIES` setting feeding uvicorn's `forwarded_allow_ips`), and document that operators must set it to their proxy's address. Take the rightmost untrusted hop rather than the leftmost value.
2. Add a documented setting for how many proxy hops to strip, and derive the client IP from that.
3. At minimum, detect the case at startup: if the limiter is producing a single distinct key, that is almost always a misconfiguration worth a warning in the logs.

Happy to send a PR for whichever direction you prefer.

## Environment

- Securo 0.14.5, images from ghcr, `docker-compose.prod.yml` topology (nginx frontend proxying `/api/` to the backend).
- Reverse proxy in front of the frontend container.

AmirF194:
`RateLimiter` keyed its Redis bucket by `request.client.host`, which in the shipped `docker-compose.prod.yml` topology is always nginx's container address, so every client shared one bucket per rate-limited endpoint (matches your repro exactly).

Added `TRUSTED_PROXY_HOPS` (your option 2), 0 by default so nothing changes unless an operator opts in. When set, `resolve_client_ip` takes the Nth-from-right entry of `X-Forwarded-For` instead of the raw socket peer, since each trusted proxy appends its own observed peer to the right end regardless of what a client puts on the left (a client-forged leading entry can't override the trusted hop). Set it to 1 in `docker-compose.prod.yml`'s defaults, since that topology always has exactly one hop, nginx, in front of the backend.

Didn't touch the `zadd`-before-check ordering you flagged as compounding the impact. That's a separate fix (making the window check atomic) and this PR is just the IP derivation.

Tests: added unit tests for `resolve_client_ip` covering the hop math directly, including the forged-prefix and short-chain cases. The two existing rate-limit integration tests are untouched. Ran `ruff`, `ty`, and the full backend suite on Python 3.14, all green.

Fixes #801

Rate-limit client IPs now use the configured trusted proxy hop from `X-Forwarded-For`. This prevents clients behind the bundled nginx proxy from sharing one rate-limit bucket.

- Added `TRUSTED_PROXY_HOPS`, defaulting to `0`.
- Production Compose sets it to `1`.
- Added coverage for proxy-hop and short-header cases.


tassionoronha:
Important piece you added here @AmirF194 !
Thanks for that 🙏🏽

AmirF194:
Appreciated. That shared-bucket collapse behind nginx was easy to miss until someone actually hit it with concurrent traffic.

PR #407, opened 2026-07-09
Speed up backend PR checks

RomuloGatto:
## Summary
- run backend pytest with `pytest-xdist` (`-n auto --dist loadfile`) so the 2.2k-test backend suite uses GitHub runner cores
- keep the required `Backend Tests` check fast on PRs that only touch frontend/docs files by detecting backend-relevant changes before Python setup
- keep full backend checks on pushes to `main` and on PRs touching `backend/` or `.github/workflows/ci.yml`

## Diagnosis
Recent CI runs show the backend bottleneck is almost entirely `Run tests with coverage`:

- PR #401 latest run: Backend Tests 15m31s total; tests-with-coverage 14m51s; dependency install ~29s
- PR #406 frontend-only run: Backend Tests 15m21s total despite no backend code changes; tests-with-coverage 14m43s
- Frontend Checks finish in ~1m10s

So dependency caching is not the issue; the backend suite is serial and runs unnecessarily on frontend-only PRs.

## Results from this PR
- Backend Tests: 6m12s total on GitHub Actions
- `Run tests with coverage`: 5m29s on GitHub Actions
- Frontend Checks: 1m05s

That cuts backend PR wait time from about 15m30s to about 6m12s for backend-touching PRs. Frontend/docs-only PRs should skip Python setup and the backend suite entirely while keeping the required `Backend Tests` check green.

## Verification
- `uv run --extra dev pytest -n auto --dist loadfile --cov=app --cov-report=xml --cov-report=term-missing --cov-fail-under=60 -q`
  - 2270 passed, 7 skipped
  - Total coverage: 92.89%
  - Local wall time: 247.49s / 4m07s
- `uv run --extra dev ruff check .` — passed
- `python -c "import yaml; yaml.safe_load(open('.github/workflows/ci.yml'))"` — parsed
- backend path detector smoke test:
  - frontend/docs-only paths -> skip
  - backend paths -> run
  - CI workflow path -> run
- PR CI: Backend Tests and Frontend Checks passed

PR #870, opened 2026-09-07
ci(workflows): use native uv and speed up test suites

### LINKED ISSUE #820: [Bug] Test suite: a uuid4 can be stored as REAL in SQLite, breaking reads with "'float' object has no attribute 'replace'"
## Description

The backend test suite fails intermittently with:

```
AttributeError: 'float' object has no attribute 'replace'
```

raised from SQLAlchemy's UUID result processor, at fixture setup, on any test that reads back a row with a UUID column. I hit it on `tests/test_transfer_api.py::test_get_transfer_pair_returns_counterpart`, inside the `test_account` fixture's `session.refresh(account)`, but the affected test is effectively random.

No test writes a float. The value comes from `uuid.uuid4()` itself.

## What I Found

`tests/conftest.py` carries this comment:

> SQLite doesn't support PostgreSQL UUID type natively — SQLAlchemy handles the mapping automatically when we create tables via Base.metadata (it converts PostgreSQL UUID to CHAR(32)).

SQLAlchemy does not do that conversion. The DDL emitted for `postgresql.UUID(as_uuid=True)` on SQLite is the literal string `UUID`:

```sql
CREATE TABLE probe (id UUID NOT NULL, PRIMARY KEY (id))
```

SQLite decides [column affinity](https://www.sqlite.org/datatype3.html#determination_of_column_affinity) by substring of the declared type name, in order: `INT`, then `CHAR`/`CLOB`/`TEXT`, then `BLOB`, then `REAL`/`FLOA`/`DOUB`, otherwise **NUMERIC**. The string `UUID` contains none of them, so the column gets NUMERIC affinity — which coerces any text that *looks* numeric into a number.

A UUID hex is 32 characters from `0-9a-f`. In that alphabet `e` is a valid hex digit, while `.`, `+` and `-` are not. So there are exactly two ways a UUID hex can also be a valid SQLite numeric literal:

| Pattern | Stored as | Probability per UUID |
| --- | --- | --- |
| `^\d+$` | REAL ≈ 1.1e31 (overflows int64) | 2.94e-7 |
| `^\d+e\d+$` | REAL, and `inf` when the exponent is large | 8.82e-7 |
| | **total** | **1 in 850,700** |

The suite generates on the order of 140k UUIDs per run, so roughly 0.16 expected failures per run — rare enough to look like a flake, frequent enough to keep coming back. In my CI it hit once in six full runs.

## Steps to Reproduce

Deterministic, no test suite needed:

```python
import sqlite3, uuid
from sqlalchemy import Column, MetaData, Table, create_engine, select
from sqlalchemy.dialects.postgresql import UUID as PG_UUID

md = MetaData()
t = Table("probe", md, Column("id", PG_UUID(as_uuid=True), primary_key=True))
poison = uuid.UUID(hex="1e999999999999999999999999999999")  # 32 valid hex chars

engine = create_engine("sqlite://")
with engine.begin() as c:
    md.create_all(c)
    c.execute(t.insert().values(id=poison))
    print(c.exec_driver_sql("SELECT typeof(id), id FROM probe").one())
    print(c.execute(select(t.c.id)).scalar())
```

Output:

```
('real', inf)
AttributeError: 'float' object has no attribute 'replace'
```

The all-digits variant (`uuid.UUID(hex="1" * 32)`) fails the same way, with a finite float instead of `inf`.

## Expected Behavior

A UUID written to the test database reads back as the same UUID, regardless of which value `uuid4()` happened to produce.

## Actual Behavior

Roughly 1 in 850,700 UUIDs is silently stored as a REAL and breaks every subsequent read of that row.

## Scope

**Production is not affected.** `database_url` defaults to `postgresql+asyncpg`, and PostgreSQL has a native `uuid` type with no affinity coercion. This is specific to the SQLite-backed test harness.

## Suggested Fix

Render the type as `CHAR(32)` on SQLite, which has TEXT affinity — i.e. make the existing comment true:

```python
from sqlalchemy.dialects.postgresql import UUID as PG_UUID
from sqlalchemy.ext.compiler import compiles

@compiles(PG_UUID, "sqlite")
def _uuid_as_char32(element, compiler, **kw):
    return "CHAR(32)"
```

I verified this on the full suite: 4131 passed, coverage unchanged at 92.71%. It affects 168 UUID columns across 57 tables, and it does not change the type emitted for the PostgreSQL dialect.

If it's useful, a regression test can assert the property over `Base.metadata` rather than a fixed table list, so a future model with a UUID key is covered automatically. Happy to open a PR if you'd like.

## Environment

- Backend test suite, `sqlite+aiosqlite:///:memory:` with `StaticPool`
- SQLAlchemy 2.0.52, Python 3.14.7, pytest 9.1.1 with `-n auto --dist loadfile`
- Also reproduced on Python 3.14.5 / macOS and Debian 13

## Additional Context

Found while running the suite on a private fork; the reproduction above is independent of any fork change and runs against a bare SQLAlchemy install.

ADolkun:
## What

Use native locked uv projects, reduce backend fixture overhead, and select complete CI suites safely from PR changes. Run the seven previously skipped PostgreSQL report scenarios against disposable PostgreSQL 16 in the existing backend test job.

Rebased onto upstream `2869ff40cbf56da7d0d07aba602c52617a71f137`, preserving #859's dependency update and #873's warning cleanup. Regenerated `backend/uv.lock` with the pinned uv 0.12.10; all 122 non-project package entries, including artifact hashes, exactly match this upstream base. The PostgreSQL comment marker requested in review is removed.

## Why

Repeated default-cost password hashing and per-table database cleanup add avoidable test overhead. The existing changed-file detector also skipped backend checks for shared Compose edits and backend-to-docs renames. The shared `.github/ci` selector verifies the PR merge and uses NUL-delimited paths with both rename endpoints; unknown, shared, or unverifiable changes run all suites, and main pushes always run all suites.

Historical hosted measurements from the earlier version of this PR showed lower elapsed time with every suite running:

| Hosted measurement | Baseline `4f4da94c` | Earlier PR head `ba6262c6` | Earlier PR head `e8250520` |
| --- | ---: | ---: | ---: |
| Total elapsed | 6m23s | 2m42s | 3m24s |
| Backend job | 380s | 158s | 201s |
| Backend dependency installation | 38s | 3s | <1s |
| Backend tests with coverage | 327s | 112s | 168s |
| Frontend job | 103s | 86s | 89s |
| Frontend tests | 55s | 36s | 37s |

Runs: [baseline](https://github.com/securo-finance/securo/actions/runs/34109071026), [earlier PR head `ba6262c6`](https://github.com/securo-finance/securo/actions/runs/34166728893), [earlier PR head `e8250520`](https://github.com/securo-finance/securo/actions/runs/34167469622). These are individual historical observations, not medians or measurements of the newly rebased head. The earlier successful runs used four pytest workers; `e8250520` changed only a documentation comment. The first uv cache was cold and the later install reused it. Both runs paid about 22s to start PostgreSQL.

## Behavior and scope

- Keep existing job names, triggers, the unconditional migration-chain check, and the canonical coverage-badge guard. Retain upstream's `-ra -W error` backend gate and `npm run lint -- --max-warnings=0` frontend gate.
- Keep dependency resolution locked across CI, Docker, release preparation, and local development. Native packaging includes the API and MCP modules, with the container environment outside `/app` so source mounts cannot hide dependencies.
- Keep real password hash, verify, and upgrade operations in tests, while lowering only implicit test costs. Explicit hash parameters and production defaults remain unchanged. Batch test cleanup atomically and preserve SQLite UUID values as text.
- Keep all discovered frontend test identities and isolation, splitting pure library tests into Node and retaining jsdom for browser-dependent tests. Preserve upstream's PDF.js test alias.

PostgreSQL still starts before a skipped backend job reaches path detection; the existing-job design retains that startup cost. The real PostgreSQL report tests cover their SQL paths and per-test schema isolation, not additional vector-extension or migration integration coverage. Production application source and runtime dependency declarations are unchanged relative to the latest upstream base.

## Current local validation

Validated the rebased candidate with the comment change now committed as `311518cbc50906ea96b07b1c027be396880acea9`:

- Backend: `uv sync --locked --group dev`, lock freshness, Ruff, and ty pass. The complete workflow pytest command, including warnings as errors and coverage, passes against disposable PostgreSQL 16: **3,835 passed, no skips, 92.64% coverage**. All seven former PostgreSQL skips execute. Four workers were used locally.
- Frontend, Node 24: `npm ci`, zero-warning lint, typecheck/build, and `npm test` pass: **718 tests across 76 files**. A discovery comparison with the upstream configuration preserves the exact multiset of test identities: 297 Node and 421 DOM tests.
- CI: all seven selector regression tests, the single 89-migration chain, workflow contract checks, and actionlint pass. ShellCheck integration was unavailable locally. Helm/chart source is unchanged by this PR; Helm and kubeconform were not rerun locally.
- Packaging: both Linux ARM64 and AMD64 backend images build and pass API/MCP/CLI/Celery/Alembic import smokes with read-only source mounts and networking disabled. A separate locked Python 3.11 runtime installation and import smoke also pass.
- Negative checks: a stale manifest fails lock verification, missing `POSTGRES_TEST_URL` fails under CI, and missing PostgreSQL configuration skips explicitly in local-only runs.

Hosted CI also passed at `311518cbc50906ea96b07b1c027be396880acea9`: [CI run](https://github.com/securo-finance/securo/actions/runs/34498991828). Two independent post-push agent reviews inspected the full PR at this exact head and found no actionable defects, including separate fixture, PostgreSQL isolation, dependency-lock, selector, and frontend-discovery probes. These are independent agent review results; formal maintainer approval and merge gates remain separate.

## Related issue

Closes #820.

Builds on already-merged #407 (parallel pytest), #667 (fork-safe badge), #777 (month-end test fix), #858 (uv 0.12.10), #859 (dependency update), and #873 (warning cleanup).

## Checklist

- [x] Local backend tests, lint, type checks, and coverage pass
- [x] Local frontend lint, tests, and build pass
- [x] No user-facing strings or application business logic changed
- [x] Independent pre-push review of the rebase and comment update completed
- [x] Exact new-head hosted CI passes
- [x] Two independent post-push reviews of the updated PR completed

CI now uses locked `uv` environments and selects complete suites safely from changed files. SQLite UUID handling and PostgreSQL report coverage were improved.

- PostgreSQL report scenarios now run against PostgreSQL 16.
- SQLite UUID values keep their original value when read.
- CI runs the required backend, frontend, Helm, and Docker checks for each change.
- Node.js 24 is used for frontend checks.

Risk: money math.

### DIFF
--- a/.github/ci/changes.py
+++ b/.github/ci/changes.py
@@ -1,0 +1,63 @@
+"""Select complete CI suites only when a verified PR merge proves relevance."""
+
+import json
+import os
+import subprocess
+from pathlib import Path
+
+PROSE = {
+    b"README.md", b"CONTRIBUTING.md",
+    b"CODE_OF_CONDUCT.md", b"SECURITY.md", b"LICENSE",
+}
+
+
+def git(*args):
+    return subprocess.check_output(["git", *args], stderr=subprocess.PIPE, timeout=30)
+
+
+def changed_jobs():
+    # Every push validates the integrated codebase, including documentation pushes.
+    if os.environ.get("GITHUB_EVENT_NAME") != "pull_request":
+        return True, True, True
+
+    try:
+        event = json.loads(Path(os.environ["GITHUB_EVENT_PATH"]).read_bytes())
+        pr = event["pull_request"]
+        if os.environ["GITHUB_REF"] != f"refs/pull/{event['number']}/merge":
+            raise ValueError("checkout is not the PR merge ref")
+        parents = git("rev-list", "--parents", "-n", "1", "HEAD").decode("ascii").split()
+        if parents != [os.environ["GITHUB_SHA"], pr["base"]["sha"], pr["head"]["sha"]]:
+            raise ValueError("checkout does not match the event's merge and parents")
+
+        # NULs preserve odd filenames; disabling renames retains both sides of a move.
+        diff = git("diff", "--no-ext-diff", "--no-renames", "--name-only", "-z",
+                   "HEAD^1", "HEAD", "--")
+        if not diff.endswith(b"\0"):
+            raise ValueError("empty or incomplete changed-file list")
+        backend = frontend = helm = False
+        for path in diff[:-1].split(b"\0"):
+            if path.startswith(b"backend/"):
+                backend = True
+            elif path.startswith(b"frontend/"):
+                frontend = True
+            elif path.startswith(b"charts/securo/"):
+                helm = True
+            elif path in PROSE or (path.startswith(b"docs/") and path.endswith(b".md")):
+                continue
+            else:
+                # Unknown/shared paths (including workflows and this script) run everything.
+                return True, True, True
+        return backend, frontend, helm
+    except (OSError, ValueError, KeyError, TypeError, subprocess.SubprocessError) as error:
+        print(f"Cannot determine PR relevance ({type(error).__name__}); running all suites.")
+        return True, True, True
+
+
+if __name__ == "__main__":
+    backend, frontend, helm = changed_jobs()
+    outputs = (f"run_backend={str(backend).lower()}\n"
+               f"run_frontend={str(frontend).lower()}\nrun_helm={str(helm).lower()}\n")
+    # Failure to publish is a failed step, never a successful skip.
+    with open(os.environ["GITHUB_OUTPUT"], "a") as output:
+        output.write(outputs)
+    print(outputs, end="")
--- a/.github/ci/test_changes.py
+++ b/.github/ci/test_changes.py
@@ -1,0 +1,184 @@
+"""Run with python3 .github/ci/test_changes.py; no application dependencies needed."""
+
+import json
+import os
+import subprocess
+import sys
+import tempfile
+import unittest
+from pathlib import Path
+
+SELECTOR = Path(__file__).with_name("changes.py").resolve()
+
+
+class CIChangesTest(unittest.TestCase):
+    def setUp(self):
+        self.temp = tempfile.TemporaryDirectory()
+        self.addCleanup(self.temp.cleanup)
+        self.repo = Path(self.temp.name) / "repo"
+        self.repo.mkdir()
+        self.env = dict(os.environ, GIT_CONFIG_NOSYSTEM="1", GIT_CONFIG_GLOBAL=os.devnull)
+        self.git("init", "-q", "-b", "base")
+        self.git("config", "user.name", "CI Test")
+        self.git("config", "user.email", "ci@example.invalid")
+        for name in ("README.md", "docs/guide.md", "backend/old.py", "frontend/old.ts", "charts/securo/old.yaml"):
+            self.write(name, "original\n")
+        self.git("add", ".")
+        self.git("commit", "-qm", "base")
+        self.base = self.git("rev-parse", "HEAD")
+
+    def git(self, *args):
+        return subprocess.check_output(
+            ["git", "-c", "commit.gpgSign=false", "-c", f"core.hooksPath={os.devnull}", *args],
+            cwd=self.repo, env=self.env, stderr=subprocess.PIPE,
+        ).decode().strip()
+
+    def write(self, name, text):
+        path = self.repo / name
+        if text is None:
+            path.unlink()
+        else:
+            path.parent.mkdir(parents=True, exist_ok=True)
+            path.write_text(text)
+
+    def merge(self, files):
+        self.git("switch", "-C", "feature", self.base)
+        for name, text in files.items():
+            self.write(name, text)
+        self.git("add", "-A")
+        self.git("commit", "--allow-empty", "-qm", "feature")
+        head = self.git("rev-parse", "HEAD")
+        self.git("switch", "--detach", self.base)
+        self.git("merge", "--no-ff", "--no-edit", "feature")
+        self.event = {"number": 123, "pull_request": {
+            "base": {"sha": self.base}, "head": {"sha": head},
+        }}
+        self.env.update(GITHUB_EVENT_NAME="pull_request", GITHUB_REF="refs/pull/123/merge",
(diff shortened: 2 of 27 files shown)


tassionoronha:
One thing before this goes in: `ci.yml` line 18 still carries a `ponytail:` prefix on the PostgreSQL comment. The repo is public, so worth dropping the marker.

Everything else reads well from here. CI on this branch runs in 2 to 3 minutes against 6 to 8 on main, and scoping the Postgres service to the report tests instead of the whole suite is the right call.

tassionoronha:
Update: #859 just landed on main and it touches the same two files this branch reworks, so we now have a conflict in `backend/pyproject.toml` and `backend/uv.lock`. Resolving a lock file by hand tends to leave it inconsistent, so this one is better rebased on your side than by us.

ADolkun:
Dropped the marker and rebased onto latest main (`2869ff40`), preserving the #859 dependency update and #873 warning gates. Regenerated `uv.lock` with the pinned uv 0.12.10; all non-project lock entries match main. Local validation passed: 3,835 backend tests with disposable PostgreSQL and warnings as errors, 718 frontend tests, lint/type/build checks, selector and migration checks, and both backend image architectures. Updated head: `311518cbc50906ea96b07b1c027be396880acea9`.

tassionoronha:
Thanks for your patience, @ADolkun! I rebased this onto main and narrowed it to CI and tests: the Dockerfiles, Node version and backend packaging are back to what main ships, since many people build and run these images and we don't want any change on their side from this PR. CI now installs with `uv sync --locked --all-extras` on the existing layout, and the selector, PostgreSQL report tests and fixture speedups are all kept.

tassionoronha:
One more note, @ADolkun: we trimmed the changes down to what the PR title covers (CI and tests). For future PRs, please keep each one focused on its main goal and leave unrelated changes (Docker images, runtime versions, packaging, docs) for separate PRs. Smaller, focused PRs are much easier and faster for us to review and merge. Thanks again for the work here!

PR #872, opened 2026-09-07
feat: configure providers and backend timezone in app

### LINKED ISSUE #209: Feature Request: Configurable Backend Timezone via Settings UI
Summary

Add a configurable timezone setting in the Securo admin/settings interface that defines the backend application timezone used by automations, AI agents, reminders, scheduled tasks, and date-sensitive calculations.

Problem

Currently the backend/container uses the server/container timezone (commonly UTC). This causes inconsistencies for users running automations and AI tasks based on local dates and times.

Example:

Local timezone: America/Sao_Paulo (-03)
Local date/time: May 18 22:00
Backend/container timezone: UTC
Backend considers current date as May 19

This affects:

AI agent date interpretation
scheduled automations
reminders

In my case, AI automations were interpreting Monday as already being Tuesday because the backend/container was running in UTC instead of the local timezone.

Suggested Solution

Add a timezone setting accessible from the Securo interface.

Example:

Settings → System → Timezone
Dropdown with IANA timezones:
America/Sao_Paulo
America/New_York
Europe/London
etc.

### LINKED ISSUE #856: Configure providers in the app instead of environment variables
Provider credentials live only in environment variables today, so changing one means editing `.env` and restarting the containers. That is a lot of friction for something a person may want to set up after the stack is already running, and it is easy to miss: a rate provider that was never configured fails quietly rather than saying so.

We already have in-app settings backed by the `app_settings` table and the admin screen, covering things like number format, date format and registration. The idea is to extend that to provider configuration instead of adding more environment variables.

Good candidates to start with:

- **Exchange rates.** `OPENEXCHANGERATES_APP_ID`
- **Bank sync.** Pluggy, Enable Banking and SimpleFIN credentials

Worth keeping in mind:

- Environment variables should keep working, so existing deployments are untouched and anyone who prefers configuring by file can. In-app value wins when both are set.
- These are secrets, so they need to be encrypted at rest and never sent back to the browser once saved.
- Providers register themselves when their credentials are present, so registration has to become dynamic rather than settled at startup.

Related: #209, which asks for the same thing for the backend timezone.

domgjntoki:
## What

- Add write-only, Fernet-encrypted in-app credentials for Open Exchange Rates, Pluggy, and Enable Banking, plus the SimpleFIN instance toggle.
- Resolve provider configuration for each operation so saves and removals take effect without restarting API or worker processes, while preserving environment-variable fallback.
- Add an admin-selectable IANA backend timezone for automations, AI context, reminders, scheduled work, and date-sensitive calculations.
- Apply the timezone to HTTP requests, MCP calls, Celery jobs, and the recurring-generation CLI while retaining UTC timestamps and provider-specific UTC protocol boundaries.
- Add migration `086` to widen `app_settings.value` for encrypted PEM data, localized admin controls, and regression coverage.

## Why

Provider credentials currently require editing `.env` and restarting the stack, and missing exchange-rate configuration fails without a clear setup path. Backend calendar behavior also follows the container clock, which can place local evening activity on the following day.

These settings make both behaviors configurable after deployment. Saved provider values take priority over environment values; removing a saved value restores an available environment fallback.

## How to Test

1. Open **Admin Settings**, save provider credentials, and verify the provider becomes available without restarting the stack.
2. Replace or remove a saved credential and verify the confirmation explains whether an environment fallback exists. Confirm that saved values are never returned by the API.
3. Change **Backend timezone** across a UTC date boundary and verify recurring generation, reports, AI context, and other calendar calculations use the selected local date.
4. Verify environments that only use existing provider variables and `TZ` continue to work without saving in-app overrides.

Automated validation performed:

- Backend focused provider/timezone regression suite: 297 passed.
- Full backend suite on Windows: 3,484 passed and 47 skipped; the sole failure is the existing Unix-only `signal.SIGALRM` PDF test.
- Frontend suite: 708 passed; focused provider/timezone components: 6 passed.
- Ruff and ty passed.
- Frontend lint completed with 0 errors and 40 existing warnings; typecheck and production build passed.
- Alembic reports one head at `086`, the migration-chain script passes, and PostgreSQL offline SQL generation passes.
- Headless Edge verified the real admin/API flow, accessibility tree, console/network behavior, and layouts from 320px through 1440px.

## Deployment and rollback

Run migration `086` before saving provider credentials. A unique `SECRET_KEY` of at least 32 characters is required for in-app secret storage and must remain stable so existing ciphertext can be decrypted.

For an application rollback, revert these commits and leave the widened `TEXT` column in place; existing settings remain compatible. If the schema must also be downgraded, remove provider overrides longer than the former 2,000-character limit before downgrading `086`.

## Related Issue

Closes #856
Closes #209

## Checklist

Adds admin settings for encrypted provider credentials and the backend IANA timezone. Provider changes apply without restarts, while environment fallbacks remain available.

- Configure Open Exchange Rates, Pluggy, Enable Banking, and SimpleFIN in the admin settings.
- Saved credentials stay encrypted and write-only.
- Select the backend timezone for schedules, reminders, AI context, syncs, and date-based calculations.
- Use the selected timezone for local dates while keeping UTC timestamps and provider protocols unchanged.

Risk: database migration, auth, money math, and sync provider.

# Screenshots
image
image
image

Adds admin settings for encrypted provider credentials and the backend IANA timezone. Provider values resolve at runtime, with environment variables as fallbacks, and date-sensitive operations use the selected timezone.

- Configure Open Exchange Rates, Pluggy, Enable Banking, and SimpleFIN in the admin settings.
- Save credentials without exposing them after storage.
- Change the backend timezone from the admin settings.
- Use the selected timezone for reminders, automations, scheduled work, AI context, reports, and date-based calculations.
- Keep UTC timestamps and provider protocol boundaries unchanged.

Risk: database migration, auth, money math, sync provider.


domgjntoki:
Adding app settings for dates seemed like an already big change to be done to the app, with multiple files being changed at once.
Will split this into 2 merge requests, one for the bank setup, and one just for app dates to be consistent with user timezones.

PR #873, opened 2026-09-08
ci: eliminate backend and frontend warnings

ADolkun:
CI currently emits noisy backend warnings and 40 frontend lint warnings, while the permission audit skips its 39 allowlisted routes. This fixes the warning sources and makes those exemptions executable assertions. CI now treats pytest warnings and ESLint warnings as failures.

- Use synthetic test signing credentials, correct Redis pipeline mocks, update deprecated APIs, remove misplaced asyncio markers, and make date-sensitive assertions deterministic. Keep the two known ofxparse compatibility warnings scoped to that parser call, with tests proving unrelated warnings and parser errors still propagate.
- Preserve form drafts/reset behavior while fixing hook dependencies and Fast Refresh exports. Give real charts deterministic jsdom dimensions and use PDF.js's Node entry point only in tests.
- Load optional translations on demand through i18next, retaining English fallback and waiting for the saved language before rendering. The i18n chunk drops from 1,576 kB to 216 kB under the existing build budget.

This is independent of #870. Its native-uv, test-harness performance changes, and restoration of the seven PostgreSQL-only scenarios remain in that PR. This PR keeps the current dependency locks and CI coverage configuration.

Validation: full backend suite passed with 3,494 tests, zero warnings, and only the seven existing PostgreSQL-only skips; all 710 frontend tests passed without warnings. Frontend lint/typecheck/build, backend Ruff/ty, and actionlint passed. Two independent pre-PR reviews and two pinned-head post-PR reviews found no functional defects. The two formerly timezone-sensitive invoice assertions also pass under the local non-UTC timezone.

[GitHub Actions passed](https://github.com/securo-finance/securo/actions/runs/34178926795) at `a48abcd30a74ed4fddcfc99a7ddb27358218a08b`: backend, frontend, migration chain, and Helm checks all succeeded.


tassionoronha:
Thanks for that @ADolkun ! LGTM

PR #875, opened 2026-09-09
fix(payees): serialize legacy payee types as null

### LINKED ISSUE #830: [Bug] Payees list returns 500 when legacy type values (transfer, employer) exist
## Description

`GET /api/payees` returns **500 Internal Server Error** when any payee in the workspace has a legacy `type` value that is not `"person"` or `"company"` (e.g. `"transfer"`, `"employer"`, or the retired `"merchant"`).

The payees are **not deleted from the database**. They still exist, but the list endpoint cannot serialize them because `PayeeRead` uses `PayeeType = Literal["person", "company"]`, which rejects any other string value.

## Root Cause

PR #602 (`feat: enrich payees with contact and fiscal data`, merged 18 Aug 2026) changed `PayeeType` from a free-form string to `Literal["person", "company"]`. Migration 070 removed the `merchant` server default and added the `source` column, but **did not clean up existing rows** that had `type = "transfer"`, `"employer"`, or `"merchant"`.

When the list endpoint serializes these rows, Pydantic raises a `ResponseValidationError` and FastAPI returns 500.

## Backend Error Log

```
fastapi.exceptions.ResponseValidationError: 6 validation errors:
  Input should be person or company, input: transfer
  Input should be person or company, input: employer
  (6 payees affected)
```

## Steps to Reproduce

1. Have a workspace with payees that have legacy `type` values (`transfer`, `employer`, `merchant`)
2. Call `GET /api/payees`
3. Returns 500 instead of the payee list

Note: `POST /api/payees` (create) and `GET /api/payees/{id}` (get single) work fine — only the list endpoint is affected because it serializes all payees at once.

## Expected Behavior

The list endpoint should gracefully handle legacy `type` values — either:
- Coerce unknown values to `null` via a `@field_validator` on `PayeeRead.type`
- Or accept any string as a valid type

## Affected Code

- `backend/app/schemas/payee.py` — `PayeeType = Literal["person", "company"]` (too strict for existing data)
- `backend/app/api/payees.py` line 47 — `list_payees` endpoint

## Suggested Fix

Add a `@field_validator` to `PayeeRead.type` that coerces unknown values to `null`:

```python
from pydantic import field_validator

@field_validator("type", mode="before")
@classmethod
def coerce_legacy_type(cls, v):
    if v is not None and v not in ("person", "company"):
        return None
    return v
```

This preserves the API contract (clients still get `"person"`, `"company"`, or `null`) while handling legacy data gracefully. Optionally, a data migration can clean up the DB values afterward.

## Environment

- Securo latest (`d4b698b`)
- PostgreSQL
- Self-hosted on Kubernetes

wolfgang-aura:
Closes #830.

Payees created by older data imports can retain `transfer`, `employer`, or `merchant` as their type. The payees list endpoint serialized those rows through a response schema that accepted only `person` and `company`, so the endpoint returned 500.

The read-side response validator now maps unknown persisted type values to `null`. Create and update validation remain strict, so new writes still accept only `person` and `company`.

### How this was tested

- Payee API tests: 20 passed.
- Focused payee type and source tests: 13 passed.
- `ruff check` and `ty check` reported no findings.
- The exact reproducer passed with `transfer`, `employer`, and `merchant` serializing as `null`.

The broad backend suite remains environment-limited by existing setup and temporary-database failures. The target repository's CI should provide the full-project check.

### Alternative considered

Accepting every persisted string in the response schema would avoid the 500 but weaken the API contract. Mapping only unknown read values to `null` preserves known values and keeps write validation strict.

### AI disclosure

This change was drafted and reviewed with AI assistance through Codex (`gpt-5.6-luna`) under Mailman. The verification results above were recorded by Mailman's test harness.

`GET /api/payees` now returns legacy payees without a 500 error. Unknown types map to `null`, while new payee types remain limited to `person` and `company`.

- Legacy `transfer`, `employer`, and `merchant` types appear as `null`.
- `person` and `company` remain valid payee types.
- New payees cannot use unknown types.

Risk: none of these.

### DIFF
--- a/backend/app/schemas/payee.py
+++ b/backend/app/schemas/payee.py
@@ -3,7 +3,7 @@ from datetime import date as _Date, datetime
 from decimal import Decimal
 from typing import Literal, Optional
 
-from pydantic import BaseModel, ConfigDict, Field
+from pydantic import BaseModel, ConfigDict, Field, field_validator
 
 from app.fiscal.registry import TaxIdKind
 from app.schemas.category import CategoryRead
@@ -71,6 +71,14 @@ class PayeeRead(BaseModel):
     user_id: uuid.UUID
     name: str
     type: Optional[PayeeType] = None
+
+    @field_validator("type", mode="before")
+    @classmethod
+    def coerce_legacy_type(cls, value: Optional[str]) -> Optional[str]:
+        if value is not None and value not in ("person", "company"):
+            return None
+        return value
+
     # Read-only. Exposed so a client picker can tell the handful of
     # counterparties somebody entered on purpose from the hundreds sync
     # created from card descriptors.
--- a/backend/tests/test_payees_api.py
+++ b/backend/tests/test_payees_api.py
@@ -7,6 +7,7 @@ from httpx import AsyncClient
 from sqlalchemy.ext.asyncio import AsyncSession
 
 from app.models.account import Account
+from app.models.payee import Payee
 from app.models.transaction import Transaction
 from app.models.user import User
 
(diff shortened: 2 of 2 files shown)


tassionoronha:
Thanks for this @wolfgang-aura. LGTM!

PR #889, opened 2026-09-10
fix(agents): preserve Gemini tool-call thought signatures

### LINKED ISSUE #884: [Bug] Gemini via OpenAI-compatible connection fails with missing `thought_signature` in function calls
Description

Using a Gemini connection configured as openai_compatible (endpoint https://generativelanguage.googleapis.com/v1beta/openai), any agent tool/function call fails with a thought_signature error from the Gemini API. This appears to be caused by the OpenAI-compatible conversion layer dropping the thought_signature field that Gemini's thinking-enabled models require on functionCall parts across turns.

Steps to Reproduce
Go to Settings → Connections → New connection.
Set provider type to "OpenAI-compatible", endpoint https://generativelanguage.googleapis.com/v1beta/openai, and default model gemini-3.6-flash.
Create a new agent using this connection, with access to a tool (e.g. list_recurring_transactions).
Ask the agent a question that triggers that tool.
See error.

Image

Image

Expected Behavior

The agent should call the tool and return a normal response, the same way it does with OpenAI/Anthropic connections.

Actual Behavior

The request fails with:

OpenAI 400: [{ "error": { "code": 400, "message": "Function call is missing a thought_signature in functionCall parts. This is required for tools to work correctly, and missing thought_signature may lead to degraded model performance. Additional data, function call `default_api:securo__list_recurring_transactions`, position 2.", "status": "INVALID_ARGUMENT" } }]

Likely cause: Gemini models with thinking enabled require each functionCall returned by the model to carry a thought_signature, so the model can preserve internal reasoning state across turns. The OpenAI function-calling format has no field for this, so it's likely dropped when Securo converts conversation history to/from the OpenAI format before/after calling Gemini. See Google's docs: https://ai.google.dev/gemini-api/docs/thought-signatures

Possible fixes: preserve/replay thought_signature in functionCall parts through the OpenAI-compatible conversion, or support Gemini as a native provider using generateContent/streamGenerateContent directly.

Environment
Deployment: Docker
Browser: N/A (backend/API error)
OS: -
Screenshots

N/A

Additional Context
Connection: gemini API (openai_compatible)
Model: gemini-3.6-flash
Securo version: v0.15.1

11VitorAlves11:
## What

Preserve Gemini tool-call thought signatures through OpenAI-compatible streaming, message persistence, and conversation replay. Also recognize versioned beta/alpha API paths so the Gemini `/v1beta/openai` endpoint is not given an extra `/v1` suffix.

## Why

The OpenAI-compatible provider inherits the OpenAI streaming parser, which extracted only tool-call IDs, names, and arguments. Gemini's `tool_calls[].extra_content.google.thought_signature` was discarded, and neither the internal tool-call model nor the persisted history carried it. The next request therefore replayed unsigned calls and Gemini rejected the tool results.

An optional per-call signature now survives streaming (including late signature-only deltas), the shared response collector, and the executor's JSON history. The OpenAI serializer restores it to the original `extra_content.google.thought_signature` location. Calls without signatures retain their existing wire and stored JSON shape; tool execution still receives only arguments. No database migration is needed.

Reference: https://ai.google.dev/gemini-api/docs/generate-content/thought-signatures

## How to Test

- Focused provider/streaming/executor tests: **60 passed**.
- Backend `ruff check .` and `ty check .`: **passed**.
- Full backend suite, `pytest -n 4 --dist loadfile -q -W error`: **3826 passed, 7 skipped, 1 failed**. `test_update_transaction_rejects_category_from_other_workspace` returned 404 instead of 400. It passed on an isolated rerun; the original branch also passed that test and its complete 38-test file. The full-suite failure has not been conclusively diagnosed.
- Regression coverage includes sequential and parallel tool calls, first/late/absent signatures, split arguments, exact outbound signature placement, persisted history reload, and Gemini endpoint normalization.
- Manual validation: the issue reporter tested this branch in Docker with the Gemini OpenAI-compatible connection and confirmed that the recurring-expenses query and a follow-up tool-calling query in the same conversation both succeeded.

## Related Issue

Closes #884

## Checklist

- [x] Backend tests pass in full (one failure described above)
- [x] Focused regression tests pass
- [x] Backend lint and type checks pass
- [x] Live Gemini validation by the issue reporter
- Frontend and translations: not applicable; no frontend or user-facing text changes.

Preserves Gemini tool-call signatures across streaming, saved messages, and conversation replay. It also handles versioned Gemini OpenAI-compatible endpoints without adding an extra `/v1` path.

- Gemini tool calls continue to work across conversation turns.
- Tool calls without signatures keep their existing format.
- Saved conversations restore signatures when needed.
- Tool execution receives only the tool arguments.

### DIFF
--- a/backend/app/agents/models/conversation.py
+++ b/backend/app/agents/models/conversation.py
@@ -50,7 +50,7 @@ class Message(Base):
     role: Mapped[str] = mapped_column(String(20))
     content: Mapped[Optional[str]] = mapped_column(Text, nullable=True)
 
-    # When role="assistant" with tool calls, list of {id,name,arguments}.
+    # When role="assistant", list of {id,name,arguments,thought_signature?}.
     tool_calls: Mapped[Optional[list]] = mapped_column(JSON, nullable=True)
     # When role="tool", payload is {tool_call_id, server, name, result, error?}.
     tool_result: Mapped[Optional[dict]] = mapped_column(JSON, nullable=True)
--- a/backend/app/agents/providers/base.py
+++ b/backend/app/agents/providers/base.py
@@ -51,6 +51,8 @@ class ToolCall:
     id: str
     name: str
     arguments: dict[str, Any]
+    # Opaque Gemini signature; replay unchanged, never pass to the tool itself.
+    thought_signature: Optional[str] = None
 
 
 @dataclass
@@ -87,6 +89,8 @@ class ChatChunk:
     args_delta: Optional[str] = None  # incremental JSON args while streaming
     finish_reason: Optional[str] = None
     usage: Optional[Usage] = None
+    # Delivered on tool_call_end, after all streaming deltas have arrived.
+    thought_signature: Optional[str] = None
 
 
 @dataclass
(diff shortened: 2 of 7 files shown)


11VitorAlves11:
bug_fix

tassionoronha:
Thanks for this @11VitorAlves11. LGMT! :)

PR #891, opened 2026-09-11
fix(2fa): report the real reason a 2FA disable failed

ADolkun:
## What

When disabling 2FA fails, the dialog reported "invalid 2FA code" for every
failure — including a wrong password and an unreachable server.

The backend already distinguishes these: `/2fa/disable` returns
`detail: "Invalid password"` and `detail: "Invalid 2FA code"` as separate
cases. The frontend threw that away with a bare `catch`.

Also adds `autoComplete="current-password"` / `"one-time-code"` to the two
inputs, and `role="alert"` on the error so screen readers announce it.

## Why

A user typing their password wrong is told their authenticator is wrong. They
go regenerate a code that was never the problem, fail again, and have no way to
find out what is actually wrong — there are no recovery codes, so "I can't
disable 2FA" is a bad place to get stuck.

## How to Test

1. Enable 2FA on an account.
2. Open the 2FA dialog and click Disable.
3. Enter the **wrong password** with a **valid** code → "current password is
   incorrect" (previously: "invalid 2FA code").
4. Enter the **right password** with a **wrong** code → "invalid 2FA code".
5. Stop the backend and retry → the connection error, not a credential error.

`npm test -- two-factor-setup` covers steps 3 and 5.

## Related Issue

None — small bug fix, no prior discussion per CONTRIBUTING.

## Checklist

- [x] Tests added
- [x] `npm run lint -- --max-warnings=0`, `npm test`, `npm run build` all pass
- [x] No new locale keys — `auth.currentPasswordWrong`, `auth.serverError`,
      `auth.invalid2faCode` and `common.error` already exist in all 15 locales
- [x] Not a large or core change, so no prior discussion needed

The 2FA disable flow now shows the actual failure reason instead of always showing an invalid-code error. It also improves autofill and screen reader support.

- Shows separate errors for an incorrect password, invalid 2FA code, unreachable server, and unknown failures.
- Adds autofill hints for the password and one-time-code fields.
- Announces error messages to screen readers.

Risk: auth.

### DIFF
--- a/frontend/src/components/two-factor-setup.test.tsx
+++ b/frontend/src/components/two-factor-setup.test.tsx
@@ -1,0 +1,52 @@
+import { beforeEach, describe, expect, it, vi } from 'vitest'
+import { screen } from '@testing-library/react'
+import type { AxiosError } from 'axios'
+
+import { TwoFactorSetup } from '@/components/two-factor-setup'
+import { renderWithProviders, t } from '@/test/utils'
+
+const authContext = vi.hoisted(() => ({
+  user: { is_2fa_enabled: true },
+  updateUser: vi.fn(),
+}))
+vi.mock('@/contexts/auth-context', () => ({ useAuth: () => authContext }))
+
+const api = vi.hoisted(() => ({ auth: { disable2fa: vi.fn() } }))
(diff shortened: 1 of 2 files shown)


tassionoronha:
Thanks for that @ADolkun !

PR #892, opened 2026-09-11
fix(login): don't spend a passkey challenge after an aborted ceremony

ADolkun:
## What

Three small fixes to the login page, all independent of each other:

1. **Don't spend a passkey challenge after an aborted ceremony.** The
   conditional-UI effect checks `abortController.signal.aborted` before starting
   the ceremony and again after verification returns, but not in between. If the
   user navigates away while the browser prompt is open, the credential promise
   still settles and we still `POST` it to
   `/auth/passkeys/authenticate/verify`.
2. **No passkey-button flash on first paint.** `passkeySupported` started `false`
   and was set in an effect, so the button was missing for one frame on every
   load. Moved to a lazy `useState` initializer.
3. **Accessibility.** The login error is now `role="alert"` with an `id`, the
   email input points at it via `aria-describedby`, and editing the email clears
   a stale error.

## Why

(1) is the substantive one. Challenges are one-shot server-side — stored in
Redis and consumed with `getdel`. Answering a ceremony the user walked away from
burns the challenge and logs a verification attempt nobody made.

## How to Test

1. On a device with a platform authenticator, load `/login` so conditional UI
   arms.
2. Focus the email field to trigger the passkey prompt.
3. Navigate away while the prompt is still open.
4. No request to `/api/auth/passkeys/authenticate/verify` in the network tab
   (previously: one fired after the prompt resolved).

`npm test -- login` covers this: the test holds the ceremony pending, unmounts,
then resolves the credential and asserts verification was never called. It fails
without the guard.

## Related Issue

None — small bug fix, no prior discussion per CONTRIBUTING.

## Checklist

- [x] Tests added
- [x] `npm run lint -- --max-warnings=0`, `npm test`, `npm run build` all pass
- [x] No new locale keys, no user-facing string changes
- [x] Not a large or core change, so no prior discussion needed

One note on the test setup: the existing `@/lib/webauthn` mock was a static
object with `isPasskeySupported: () => false`, so the conditional-UI effect had
never executed in any test. I converted it to hoisted `vi.fn()`s with the old
defaults restored in `beforeEach`, so all 11 pre-existing tests behave exactly
as before.

The login flow now avoids verifying passkeys after an aborted ceremony and initializes passkey support before the first paint. Login errors are also easier to identify and clear when the email changes.

- The passkey button no longer flashes after page load.
- Abandoned passkey sign-ins do not consume the server challenge.
- Login errors are announced and linked to the email field.
- Editing the email clears the current login error.

Risk: auth

### DIFF
--- a/frontend/src/pages/login.test.tsx
+++ b/frontend/src/pages/login.test.tsx
@@ -21,7 +21,11 @@ vi.mock('@/contexts/auth-context', () => ({ useAuth: () => authContext }))
 
 const api = vi.hoisted(() => ({
   setup: { status: vi.fn() },
-  auth: { oidcConfig: vi.fn() },
+  auth: {
+    oidcConfig: vi.fn(),
+    passkeyAuthenticationOptions: vi.fn(),
+    verifyPasskeyAuthentication: vi.fn(),
+  },
   admin: { registrationStatus: vi.fn(), defaultColors: vi.fn() },
 }))
 vi.mock('@/lib/api', () => ({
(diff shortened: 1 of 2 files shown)


tassionoronha:
Thanks @ADolkun, LGTM!

PR #894, opened 2026-09-11
fix(sync): inherit categories across card installments

panga:
## What

- Reuse the category from the latest earlier transaction in the same synced credit-card installment series.
- Keep transaction rules as the highest-priority category source.
- Fall back to the provider category only when no installment category exists and provider categories are enabled.
- Apply the same precedence in initial connection imports and incremental syncs.

## Why

When provider categories are disabled, each newly synced installment currently arrives uncategorized even after the user categorized earlier parcels from the same purchase. This makes the user repeat the same correction every month.

The provider already supplies a stable installment fingerprint: card account, original purchase date, installment count, total purchase amount, and transaction type. Using that fingerprint lets the next parcel inherit the user's most recent category without matching unrelated purchases.

## How to Test

1. Disable provider/integration categories.
2. Categorize an existing credit-card installment transaction.
3. Sync the following installment and verify it inherits that category.
4. Verify an unrelated installment with a different purchase date and a regular provider-categorized charge remain uncategorized.
5. Run `cd backend && ./.venv/bin/pytest tests/test_connection_service.py -q`.

## Related Issue

N/A

## Checklist

- [ ] Frontend lints clean (`npm run lint`) — not applicable; no frontend changes
- [ ] Frontend builds (`npm run build`) — not applicable; no frontend changes
- [x] Translations updated (if user-facing strings changed) — not applicable; no user-facing strings changed
- [x] For a large or core change, I discussed it first (issue or Discord) so we could align before the code (see [CONTRIBUTING](../CONTRIBUTING.md#before-large-or-core-changes)) — not applicable; narrow sync fix

New installments inherit the category from the latest earlier installment in the same purchase series. Transaction rules remain first priority, with provider categories used only as a fallback.

- Keeps installment purchases grouped under the same category.
- Applies during both initial imports and later syncs.
- Leaves regular charges and unrelated installment series unchanged.

Risk: sync provider.

### DIFF
--- a/backend/app/services/connection_service.py
+++ b/backend/app/services/connection_service.py
@@ -842,6 +842,48 @@ async def _match_pluggy_category(
     return result.scalars().first()
 
 
+async def _find_installment_category(
+    session: AsyncSession,
+    account_id: uuid.UUID,
+    txn_data,
+) -> Optional[uuid.UUID]:
+    """Reuse the latest user-selected category from an earlier installment.
+
+    Pluggy gives every parcel the original purchase date, installment count,
+    and total purchase amount. Together with the card account and transaction
+    type, those fields identify the purchase series without relying on the
+    description (whose ``03/12`` suffix changes every month).
+
+    Only earlier synced parcels qualify. This keeps manual installment series
+    separate and lets a user's most recent category correction become the
+    source of truth for the next imported parcel.
+    """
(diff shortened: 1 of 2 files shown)


tassionoronha:
Thanks for this @panga ! LGTM :)

PR #895, opened 2026-09-11
fix(dashboard): improve account balance summary

panga:
## What

- Order the dashboard's available-balance account chips by absolute balance, matching the sidebar while continuing to show every eligible checking/savings account.
- Share the non-mutating balance-order helper between the dashboard and sidebar.
- Reduce the net-worth value font size on narrow screens so it does not overlap the savings-rate column.
- Add focused coverage for positive/negative balance ordering and input immutability.

## Why

The dashboard chips previously followed API order instead of surfacing the largest balances first. On narrow screens, large net-worth values could also overlap the adjacent savings-rate value.

## How to Test

1. Run `npm run lint` from `frontend`.
2. Run `npm run build` from `frontend`.
3. Run `npx vitest run src/lib/account-utils.test.ts` from `frontend`.
4. Open the dashboard with several checking/savings accounts and verify every chip remains visible, ordered by absolute balance.
5. At a narrow viewport, verify the net-worth value does not overlap the savings-rate column.

## Related Issue

N/A

## Checklist

- [ ] Backend tests pass (`pytest`) — backend unchanged
- [x] Translations updated (not needed; no user-facing strings changed)
- [x] For a large or core change, I discussed it first (not applicable; narrow dashboard presentation fix)

The dashboard now orders account balances by absolute value and shares this ordering with the sidebar. The net worth value uses a smaller font on narrow screens to prevent overlap.

- Shows checking and savings accounts in balance order.
- Keeps accounts with negative balances in the correct position.
- Prevents narrow-screen overlap with the savings-rate column.

### DIFF
--- a/frontend/src/components/app-layout.tsx
+++ b/frontend/src/components/app-layout.tsx
@@ -1,5 +1,5 @@
 import { useState, useCallback, useEffect, useMemo } from 'react'
-import { getAccountName } from '@/lib/account-utils'
+import { getAccountName, sortAccountsByAbsoluteBalance } from '@/lib/account-utils'
 import { Link, Outlet, useLocation, useNavigate } from 'react-router-dom'
 import { useTranslation } from 'react-i18next'
 import { useDisplayLocale } from '@/hooks/use-display-locale'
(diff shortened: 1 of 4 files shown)


tassionoronha:
Thanks for extracting the sort into a helper, that part is nice. One catch when I ran it on a workspace with several currencies: the helper sorts by each account's native balance, but the chips show the converted value. So R$ 1,000 sorts ahead of € 500 even though the chips read about $185 and $580, and the row looks out of order. For the dashboard the sort should use the same converted figure the chip shows. Happy to re-review once that is in.

panga:
@tassionoronha change pushed. Thanks for the review.

tassionoronha:
Hey @panga, not sure if your changes were pushed 😅

panga:
@tassionoronha ops, should be pushed now.

PR #898, opened 2026-09-11
fix(auth): let OIDC-only users remove existing 2FA and passkeys

ADolkun:
## What

In OIDC-only mode (`local_auth_enabled = false`) the frontend hides the
Two-Factor Authentication and Passkeys menu entries entirely. A user who
enrolled TOTP or registered passkeys *before* the server switched to OIDC-only
can therefore never remove them — the UI to do so no longer exists.

This makes those entries reachable in cleanup mode: the 2FA dialog renders
disable-only, and the passkey dialog lists existing keys with their delete
buttons but no registration form.

## Why

The backend already allows exactly this, which is what makes it a frontend-only
gap rather than a policy question:

- `backend/app/api/passkeys.py` — `list_passkeys` and `delete_passkey` carry no
  `require_local_auth_enabled`, unlike the register/authenticate routes beside
  them.
- `backend/app/api/two_factor.py` — `POST /2fa/disable` carries none, unlike
  `/2fa/setup` and `/2fa/enable`.

So the server is already saying "you may clean up, you may not enroll." The UI
just wasn't saying it. And because the product has no recovery codes, a user
stranded with an enrolled factor they cannot remove has no way out.

## How to Test

1. Enable 2FA and register a passkey on an account.
2. Restart the server with local auth disabled (OIDC-only).
3. Open the user menu: **Change password** is gone (unchanged), **Disable 2FA**
   and **Passkeys** are present.
4. Disable 2FA — still requires both the password and a valid code.
5. Open Passkeys — existing keys are listed and deletable; no "add passkey"
   form.
6. Re-enable local auth: everything returns to the previous behavior.

## Related Issue

None — no existing issue covers this.

## Checklist

- [x] Backend tests added (`backend/tests/test_local_auth_cleanup.py`): 2FA
      disable under OIDC-only still requires password **and** TOTP; passkey
      list/delete stay owner-scoped (401 unauthenticated, 404 for another
      user's key, 204 for your own)
- [x] Frontend test added (`auth-controls.test.tsx`): drives the real app-layout
      desktop and mobile menus under `local_auth_enabled: false`
- [x] Both confirmed to fail without the fix
- [x] `npm run lint -- --max-warnings=0`, `npm test` (723), `npm run build` pass
- [x] `ruff check`, `ty check`, `pytest tests/test_local_auth_cleanup.py` pass
- [x] New locale key `auth.passkeysCleanupDescription` added to all 15 locales
      (EN + PT-BR translated, not placeholders)

## Two things worth a reviewer's attention

**The Passkeys menu entry is now unconditional.** Unlike 2FA, where
`user.is_2fa_enabled` tells us whether the user has anything to clean up, there
is no equivalent flag for passkeys — the count is only known after the dialog
fetches. So an OIDC-only user who never registered one now sees a Passkeys entry
that opens to an empty list. If you'd rather avoid that, the clean fix is a
`has_passkeys` boolean on the user read schema and I'll happily add it; I didn't
want to change an API schema uninvited.

**`frontend/tsconfig.test.json` is in the diff.** The new test is the first to
render the app shell, which reaches `lib/build-info` transitively and exposed
that this tsconfig's `include` omits `src/vite-env.d.ts`, where `__APP_VERSION__`
is declared — so `npm run build` failed with TS2552. The runtime side was
already handled (`vitest.config.ts` defines the global); only the type
declaration was missing. One-line `include` addition.

Minor: this touches `two-factor-setup.tsx`, which #891 also touches, but in
different regions — they should merge cleanly in either order.

OIDC-only users can remove existing 2FA and passkeys enrolled before local authentication was disabled. New enrollment and password changes remain unavailable.

- Existing 2FA can be disabled.
- Existing passkeys can be listed and deleted.
- New passkeys and 2FA enrollment remain hidden.
- Password changes remain unavailable.
- Cleanup guidance is translated across supported locales.

Risk: auth


tassionoronha:
LGTM! Thanks @ADolkun

PR #901, opened 2026-09-12
Fix MCP UUID parsing for invalid account/category ids

MPKharche:
## Summary
- `parse_uuid` returns `None` for empty/invalid input instead of raising `ValueError` from `uuid.UUID`
- `parse_uuid_list` tolerates a bare non-UUID string (e.g. a name) without crashing; filters Nones as before
- If a non-empty id list has *no* valid UUIDs, raise `ValueError` with a clear message so agents get feedback instead of silent ignore

This stops Hermes `list_transactions` from crashing when the model passes account/category names instead of UUIDs.

## Test plan
- [ ] Unit-check: `parse_uuid("Groceries") is None`, valid UUID still parses
- [ ] `parse_uuid_list("Groceries")` raises ValueError with UUID message
- [ ] `parse_uuid_list([valid_uuid, "Groceries"])` returns only the valid UUID
- [ ] Call MCP `list_transactions` with a bad `account_ids` / `category_ids` name — no traceback; clear error if all invalid
- [ ] Redeploy `mcp-server` only; money services stay up

Adds loan tracking, repayment schedules, prepayment tools, expanded MCP workspace tools, token revocation, and Frankfurter FX fallback support. It also fixes tolerant UUID parsing for MCP transaction filters and updates the frontend for loans and investments.

- Manage loans, EMI schedules, prepayments, analytics, exports, and payment links.
- Revoke web and MCP tokens.
- Attach, rename, delete, and list transaction attachments.
- Use Frankfurter when Open Exchange Rates is unavailable.
- View loans and expanded investment types in the app.
- Use safer MCP transaction filters with invalid account or category IDs.

Risk: database migration, auth, money math, sync provider.


MPKharche:
Opened against upstream by mistake; tracking on fork MPKharche/secureFinanceApp instead.

PR #906, opened 2026-09-12
feat(rules): reuse full rule editor when adding a transaction to an existing rule

### LINKED ISSUE #905: Adding a transaction to an existing rule uses a bare-bones form instead of the full rule editor
## Problem

When categorizing a transaction, you can create a new rule from it or add it to an existing rule. Creating a new rule opens the full rule editor: multiple conditions with AND/OR groups, multiple actions, priority, active toggle, apply-to-existing/overwrite-categories, and a preview.

Adding to an existing rule opens a different, much simpler dialog: pick a rule, pick a match type (contains/starts with), type the match text. It can only append a single description condition and gives no visibility into the rule's other conditions or actions before saving.

## Expected

Adding a transaction to an existing rule should use the same rich editor as creating a new rule, pre-filled with that rule's current conditions and actions plus the new match condition, so the user can review and adjust everything (groups, actions, priority, apply-to-existing, preview) before saving instead of blindly appending a condition.

## Screenshots

- Current "add to existing rule" dialog: bare match type + text.
- Full rule editor (used only for creating a new rule): conditions with AND/OR, actions, priority, active, apply-to-existing/overwrite-categories, preview.

vhsantos26:
## Summary
- "Add to existing rule" only showed a bare description match field (match type + text), unlike the full condition/action editor used when creating a rule from scratch.
- It now opens the same `RuleDialog` used for creating/editing rules, pre-filled with the picked rule's conditions and actions plus the new description-match condition, so users can edit groups, actions, priority, active state, apply-to-existing/overwrite-categories, and preview before saving.
- Dropped the two now-unused `matchOperator`/`matchText` translation keys from all 15 locale files.

Fixes #905

## Test plan
- [x] `npm run typecheck`
- [x] `eslint src/components/transaction-dialog.tsx` (only pre-existing, unrelated warnings)
- [x] `vitest run src/locales/i18n.test.ts` (key parity across locales)
- [ ] Manual verification in the browser (not done here, no local backend/test data in this environment)

“Add to existing rule” now opens the full rule editor with the selected rule’s settings and a new description condition. Users can review and change the rule before saving.

- Edit conditions, actions, priority, and active state.
- Change apply-to-existing and overwrite-categories settings.
- Preview the rule before saving.


tassionoronha:
Great PR @vhsantos26 ! LGTM :)

Issue #816, opened 2026-09-02
[Feature] Custom range date for analysis (Report and Assets)


aqlmsmdrluna:
## Problem

The analysis part of securo (Report and Assets) have only pre-defined periods of time (for Report) and no period selection for Assets, lacking a custom range option.
<img width="2319" height="263" alt="Image" src="https://github.com/user-attachments/assets/11608476-8bcf-4f25-986d-c42153ca4272" />

<img width="1895" height="1052" alt="Image" src="https://github.com/user-attachments/assets/2983f043-fe16-4a08-bb1a-cfbb918b5686" />

## Proposed Solution

Custom range is available in the transactions, it could be expanded to the analysis

<img width="1169" height="1029" alt="Image" src="https://github.com/user-attachments/assets/f9cd5c06-de3a-4155-b068-a768a2de89a8" />

yonicsurny:
I too am interested in this feature, I have imported data older than 2Y and cannot see a report for them at the moment.

PR #907, opened 2026-09-12
feat(reports): add custom date range for report analysis (#816)

yonicsurny:
## What

Adds a **Custom range** option to the Reports segment control (Net Worth, Income vs
Expenses, Money Map — Cash Flow keeps its forecast-only presets, out of scope here):

- Custom ranges are **historical and actual-only** — no projected/recurring/invoice
  forecast amounts leak into a custom window, even one ending today.
- The picker only commits on **Apply**. Opening it, editing dates, Cancel, Escape, and
  clicking outside never change the active report.
- **Reset** clears the draft; Reset → Apply clears a saved custom range and falls back
  to the tab's normal preset (`1Y` for Net Worth / Income vs Expenses, `3M` for Money
  Map).
- Client-side validation blocks a future end date or a span over 10 years before
  submitting (mirroring the backend's own cap); the backend independently rejects the
  same cases.
- A failed report request (invalid range, network error) now shows a visible error
  with **Retry** instead of silently leaving the previous result on screen.

Screenshot 2026-09-12 at 20 33 57

Screenshot 2026-09-12 at 20 34 19

Screenshot 2026-09-12 at 20 35 02

## Why

[#816](https://github.com/securo-finance/securo/issues/816) asked for a custom date
range in the analysis views (Report and Assets) — the fixed presets don't cover a
user's specific window into their own transaction history. This PR covers the Reports
half; Assets has a different report shape and is left as separate follow-up work.

This is a larger diff than the feature itself because a full review of the original
implementation, before merge, turned up four correctness/UX bugs that needed fixing
first, each with its own test coverage:

1. **Unbounded custom `end_date` reused as the forecast cutoff** — could surface
   fabricated future net-worth figures, or spurious recurring/invoice projections
   layered onto an already-closed historical range.
2. **Opening the picker auto-applied a range** before the user picked anything or even
   saw the calendar, and Cancel couldn't undo it.
3. **Reset → Apply could never clear a saved range** — Apply was disabled whenever both
   draft dates were empty, exactly the state Reset produces.
4. **A rejected or failed report request was swallowed** — the UI just kept showing the
   last successful numbers with no indication the new request had failed.

A nonblocking cleanup pass followed (deduped date-range formatting and per-tab range
config that had drifted into three places; removed a redundant per-point account query
in the Net Worth trend), then a final verification pass re-audited every changed file
against these four issues plus a related backend precedence bug, and ran the full test
suite end to end.

## How to Test

1. `docker compose up --build`, sign in, go to **Reports**.
2. On **Net Worth**, **Income vs Expenses** or **Money Map**, click **Custom range**.
3. Pick a start and end date (years are freely navigable in both calendars) and
   **Apply** — the segment shows a compact, year-less label for the active range.
4. Reopen it, click **Reset**, then **Apply** — the range clears and the tab falls back
   to its normal preset.
5. Try a future end date, or a span over 10 years — Apply is blocked with a visible
   message.
6. Confirm **Cash Flow** has no Custom range option (forecast-only presets).

Automated coverage:

- Backend (`cd backend`): `pytest` — 3848 passed, 7 pre-existing skips (4 are
  PostgreSQL-only, run in CI); `ruff check .`; `ty check .`.
- Frontend (`cd frontend`): `npm test` — 759 passed; `npm run lint -- --max-warnings=0`;
  `npm run build`.

## Related Issue

Addresses [#816](https://github.com/securo-finance/securo/issues/816) — the Reports
half of that request. Not using "Closes" since Assets custom-range is out of scope
here and the issue should stay open for that.

## Checklist

- [x] Translations updated (all 15 locales)
- [x] Discussed first — building on the already-scoped #816

## A note on AI

This implementation — the feature, the four fixes above, their tests, and the final
verification pass — was built with Claude Code. I directed the approach, reviewed
every change, ran the tests and the app myself, and I'm accountable for this PR per
the repo's [AI contribution guidelines](../CONTRIBUTING.md#using-ai).

Reports now support custom historical date ranges for Net Worth, Income vs Expenses, and Money Map. Custom ranges use actual data only and enforce a maximum of 10 calendar years.

- Apply commits the selected dates.
- Reset then Apply restores the default range.
- Future dates and ranges over 10 years are rejected.
- Failed reports show an error with Retry.
- Cash Flow keeps its forecast presets.


tassionoronha:
Tried it in the browser, the custom range works nicely and it reuses our own Button, Calendar and Popover, which is what I was hoping for. Two things to line up before merging: the calendar shows up in English when the UI is in Portuguese, because it picks the date locale instead of the UI language like our other date picker does; and the range summary uses an em dash, which we keep out of UI copy. Both are small, and with them done I'm happy to approve.

yonicsurny:
Hi @tassionoronha, thank you for the review.

I've applied the 2 changes requested.

tassionoronha:
LGTM now @yonicsurny . Thanks for that!

Issue #871, opened 2026-09-07
Feature requests: Gulf currencies, removable default categories, available balance


AhmadShiekhO:
Hi — I've been running Securo self-hosted for a few weeks and it's been really good.
Three things I keep running into that I'd love to see:

**1. Please add Gulf currencies**

There's no Saudi Riyal in the currency list, and no AED, KWD, BHD, QAR, OMR, JOD or
EGP either. All my accounts are in SAR, so on a fresh setup there's no correct
currency to choose.

**2. Let me delete the default categories**

When you first set up, the app creates its own set of categories and groups. I already
had my own, so I ended up with duplicates — including two groups with exactly the same
name, one of them empty.

I can hide the individual default categories, which helps. But the default *groups*
can't be hidden or deleted at all, so they stay in my list permanently with nothing in
them. Could the defaults be removable, at least once nothing is using them?

**3. Available balance shouldn't count savings or investments**

The "Available balance" on the dashboard adds up my checking and savings accounts. My
savings is money I've deliberately set aside and am not planning to spend, so counting
it makes the number look bigger than what I can actually use. An investment account
would have the same problem.

Meanwhile my cash account isn't included at all — and cash is the most available money
I have.

Could savings and investments be left out and cash included? Or better, a per-account
toggle for "include in available balance", since people probably disagree about savings.

Thanks for the app.

izayoood:
I need this as well, lacking some of the above currencies stop me from being able to migrate to the app, as I do not know how it would handle changing the currency later on once available.

asankahh:
**+1** — same need here. Running Securo self-hosted with accounts in AED and USD

If this issue is scoped narrower than the full GCC set, it'd be great to
cover the rest of the Gulf currencies too while it's being worked on:
SAR, AED, KWD, BHD, QAR, OMR, JOD. All seven are USD-pegged, so exchange
rate handling should be trivial once support is added — Open Exchange
Rates already covers all of them.

shaynhornik:
Picking up the remaining 2-decimal Gulf currencies from this list, SAR and QAR, in one PR following the pattern used for AED in #945. KWD, BHD, OMR and JOD use three decimal places, and the money columns are Numeric(15, 2), so I'm leaving those out until #531 (currency-specific decimal precision) is addressed rather than store rounded amounts. Will link the PR here.

shaynhornik:
[...]

PR #945, opened 2026-09-15
feat(currency): add UAE Dirham (AED) support

asankahh:
## What

Adds UAE Dirham (AED) as a supported currency: currency metadata (symbol,
decimals, flag), the setup/register currency picker, number formatting,
and the corresponding translation key.

## Why

Securo's default currency list doesn't include AED, so there's no correct
currency to select for accounts held in UAE Dirhams on a fresh setup —
see #871, which covers the broader set of missing Gulf currencies. This
PR is scoped to AED only; the rest of that issue is still open for
anyone to pick up.

## How to Test

1. Run the app and go to Setup (or Register) → Currency, confirm "UAE
   Dirham (AED)" appears in the picker.
2. Select AED as the workspace/account currency and create an account.
3. Add a transaction and confirm amounts render with 2 decimal places
   and the د.إ symbol in the correct position for the locale.
4. `cd backend && uv run pytest tests/test_currencies_api.py -v` — new
   AED assertions pass alongside the existing currency tests.

## Related Issue

Addresses #871 (partial — AED only, not closing since other Gulf
currencies in that issue are still outstanding).

## Checklist

- [x] Backend tests pass (`pytest`) — 3,820 passed, 0 failed, 7 skipped;
      `app/api/currencies.py` at 100% coverage
- [x] Translations updated (`frontend/src/locales/en.json`)
- [ ] For a large or core change... — not applicable, single-currency
      addition following the established pattern (EGP, CNY, PKR); already
      surfaced via a comment on #871 before starting

Adds AED support to currency metadata, selection, formatting, and API coverage. Users can select AED for workspaces and accounts and see amounts with the `د.إ` symbol.

- AED shows the UAE flag.
- AED uses two decimal places.
- Automatic formatting uses the `ar-AE` locale.
- The currencies API returns AED metadata.


tassionoronha:
Thanks for the AED addition, it looks right. I merged main into the branch to clear the conflict with the MDL PR that landed first, and in the same pass dropped two things that were unrelated to the currency:

- `frontend/package-lock.json`: the `dev` to `devOptional` flips were a newer npm rewriting the file, not a dependency change.
- the `currencies.AED` key in the 15 locale files: nothing reads `currencies.*` today, so they were dead keys.

The PR is now the same five-file shape as the other currency additions. Four currency PRs are queued behind each other on the same lines, so it may need one more small merge depending on the order they land.

PR #952, opened 2026-09-16
fix(helm): prevent celery beat startup failure

### LINKED ISSUE #855: [Bug] Celery beat unable to start on kubernetes by default
## Description

When deploying the helm chart with default values, we have these values set :  

```yaml
...
config
...
  agentsEnabled: "false"
persistence:
...
  agentKnowledge:
    enabled: true
```

And on beat templates, the condition use the volumeclaim `agent-knowledge` is :  

```yaml
{{- if .Values.persistence.agentKnowledge.enabled }}
```

So it tries to mount the use the claim.
But the condition for the claim to be created is :  

```yaml
{{- if and (eq (.Values.config.agentsEnabled | toString) "true") .Values.persistence.agentKnowledge.enabled }}
```

So it is not created and thus the beat pod cannot be created.

## Actual Behavior

The beat pod is not created because `agent-knowledge` does not exists

## Environment

- Deployment: helm

## Additional Context

As all others pods mounting this pvc beat deployment template should check the same condition used by the pvc to be created : 

```yaml
{{- if and (eq (.Values.config.agentsEnabled | toString) "true") .Values.persistence.agentKnowledge.enabled }}
```

But I am no expert of this application and do not know if there are any other implications behind this

Also as it seems that all agent things are deactivated by default, it may be good to set this as default value :  
```yaml
persitence:
  agentKnowledge:
    enabled: false
```

And the same occurs with the value `agentEmbeddingModels`

Which prevent beat to startup correctly (lack of a check) and could be disabled by default

jorishaenseler15:
## What
Fixed a bug where the `celery-beat` pod fails to start when deploying the securo Helm chart with default values.
Key changes include:
*   **Beat Deployment Alignment:** Updated `charts/securo/templates/beat/deployment.yaml` to ensure the agent PVC volume mounts and volumes definition use the exact same conditional logic as the PVC creation (`and (eq (.Values.config.agentsEnabled | toString) "true") .Values.persistence.agentKnowledge.enabled`).
*   **Default Persistence Values:** Set `persistence.agentKnowledge.enabled: false` and `persistence.agentEmbeddingModels.enabled: false` by default in `charts/securo/values.yaml` to match the default disabled state of the agents.

## Why
The issue occurred because the `celery-beat` deployment attempted to mount the `agent-knowledge` and `agent-embedding-models` PersistentVolumeClaims (PVCs) whenever their respective `persistence.<name>.enabled` flags were true. However, the PVCs themselves are only created if both `config.agentsEnabled` is true *and* the persistence flags are true. Since agents are disabled by default (`config.agentsEnabled: "false"`), the PVCs were never created, causing the `celery-beat` pod to get stuck waiting for missing volumes.

## How to Test
You can test this fix using a local [Kind](https://kind.sigs.k8s.io/) (Kubernetes in Docker) cluster.

### 1. Start Kind Cluster
```bash
kind create cluster --name securo-test
```

### 2. Deploy the chart with default values
```bash
helm upgrade --install securo ./charts/securo -n securo --create-namespace
```

### 3. Verify the pods are running correctly
```bash
kubectl get pods -n securo
```
The `celery-beat` pod should now be in the `Running` state because it will no longer attempt to mount the missing agent PVCs.

### 4. Test with agents enabled
To ensure it still works when agents are turned on:
```bash
helm upgrade --install securo ./charts/securo -n securo \
  --set config.agentsEnabled=true \
  --set persistence.agentKnowledge.enabled=true \
  --set persistence.agentEmbeddingModels.enabled=true
```

Wait for the pods to restart and verify the PVCs are created and bound:
```bash
kubectl get pvc -n securo
kubectl get pods -n securo
```

### 5. Cleanup
```bash
kind delete cluster --name securo-test
```

## Related Issue
Closes [#855](https://github.com/securo-finance/securo/issues/855).

## Checklist

The Helm chart now mounts agent persistence volumes only when agents are enabled. Agent knowledge and embedding model persistence are disabled by default, so the default `celery-beat` pod does not reference missing PVCs.

- Default deployments no longer mount agent PVCs.
- Agent persistence remains available when agents are enabled.


tassionoronha:
Thanks @jorishaenseler15 , LGTM :)

PR #953, opened 2026-09-17
fix(mcp): shape get_account_summary payload like the other tools

dinoconstantinou87:
## What

`get_account_summary` in the MCP server returned the service-layer dict as is, so `account_id` went out as a raw `uuid.UUID`. Starlette's `JSONResponse` renders with the stdlib `json.dumps`, which has no UUID support, and every call to the tool came back as:

```
Tool error: Object of type UUID is not JSON serializable
```

The tool now builds its payload explicitly the way every other MCP tool does: the id through `str()`, the six amounts through `num()`. The loop that converted dates was dead code, since the service returns no dates, and is gone.

The existing tool test in `test_agents_mcp_tools.py` only checked that a dict came back without an `error` key, which passes on the raw UUID. It now asserts the exact key set and derives each expected value from the `test_account` and `test_transactions` fixtures, so it fails on the unfixed tool.

## Why

The REST endpoint for the same summary never hit this because FastAPI runs its return value through `jsonable_encoder` before responding. The MCP handler hands a hand-built `JSONResponse` back, which FastAPI passes through untouched, so nothing between the tool and the wire converts types. That makes the MCP tools responsible for shaping their own output, and `_helpers.py` says so: small, stable payloads, not the full row. This was the one tool that skipped that step.

I considered adding `jsonable_encoder` in the handler as a blanket safety net. It would also have fixed this, but it hides the next tool that returns a raw row and would dump every column of it, so I kept the fix inside the tool to match the convention the rest of them follow.

## How to Test

1. `cd backend && pytest tests/test_agents_mcp_tools.py::test_get_account_summary`. Fails on `main` with the raw UUID in the `account_id` assertion, passes here.
2. With the `agents` profile up, mint an external MCP token from the UI and call the tool over JSON-RPC:

   ```
   curl -s http://127.0.0.1:8765/mcp \
     -H "Authorization: Bearer <token>" -H "Content-Type: application/json" \
     -d '{"jsonrpc":"2.0","id":1,"method":"tools/call","params":{"name":"get_account_summary","arguments":{"account_id":"<account uuid>"}}}'
   ```

   On `main` the result carries `"isError": true` with the message above. Here `structuredContent.account_id` is a string and the text block holds the same JSON.

Verified both ways locally against the `mcp-server` container on Postgres: restarted it with `main`'s version of the tool and saw the error, then with this branch and saw the summary. The full backend suite passes with `-W error` and the coverage gate (3865 passed, 7 skipped), and `ruff check .` and `ty check .` are clean.

## Related Issue

None filed. Hit it calling the tool from Claude Code against my own workspace.

## Checklist

- [x] Frontend lints clean (`npm run lint`) - n/a, no frontend changes
- [x] Frontend builds (`npm run build`) - n/a, no frontend changes
- [x] Translations updated (if user-facing strings changed) - n/a, no user-facing strings
- [x] For a large or core change, I discussed it first (issue or Discord) so we could align before the code (see [CONTRIBUTING](../CONTRIBUTING.md#before-large-or-core-changes)) - n/a, small self-contained fix in the MCP layer

`get_account_summary` now returns a fixed response shape with string account IDs and normalized amounts. This prevents failures when the response contains a UUID.

- Account summaries include consistent balance, income, and expense fields.
- Account IDs are returned as strings.
- The response test checks exact keys and expected values.

Risk: money math.


tassionoronha:
LGTM, thanks for that @dinoconstantinou87 and welcome here :)

